PR #26, opened 2018-06-12
update version number and authors 

imyiningliu:


HISTORY.rst
```diff
@@ -3,6 +3,8 @@ History
 =======
 
 0.1.0 (2018-03-28)
+0.2.0 (2018-06-12)
 ------------------
 
 * First release on PyPI.
+* Second release on PyPI.
```
jeff-regier: It's our first release on PyPI.

scvi/__init__.py
```diff
@@ -2,6 +2,10 @@
...
 
-__author__ = """Romain Lopez"""
-__email__ = 'romain_lopez@berkeley.edu'
-__version__ = '0.0.1'
+__author__ = "Romain Lopez, Jeffrey Regier, Maxime Langevin, Edouard Mehlman, Yining Liu"
+__email__ = """Romain Lopez <romain_lopez@berkeley.edu>, 
+            Jeffrey Regier <jregier@eecs.berkeley.edu>, 
+            Maxime Langevin <maxime.langevin@polytechnique.edu>, 
+            Edouard Mehlman <edouard.mehlman@polytechnique.edu>, 
+            Yining Liu <lyining@berkeley.edu>"""
```
jeff-regier: Do other projects list a lot of authors? If it's more normal to list just one author, I'd just list Romain.

HISTORY.rst
imyiningliu: Should I remove the two lines about 0.1.0?

scvi/__init__.py
imyiningliu: It's more common to have one author. I'll change it back.

HISTORY.rst
jeff-regier: Oh, I'm not sure. Maybe we should just call this release 0.1.0 after all.

jeff-regier: All the text that is there now is from when I started the project by using template-generating code called cookiecutter: https://github.com/audreyr/cookiecutter

So no need to take the files generated by cookiecutter too seriously. They're not particularly customized to our project.

setup.py
```diff
@@ -4,37 +4,53 @@
...
 with open('README.rst') as readme_file:
     readme = readme_file.read()
 
 with open('HISTORY.rst') as history_file:
     history = history_file.read()
 
-requirements = []
+
+req_path = Path('requirements.txt')
+if not req_path.is_file():
```
jeff-regier: We don't need this case -- it'll always be a file

jeff-regier:
Looks good, please merge it whenever you're ready.

jeff-regier:
By selecting "Squash and merge" from the green drop-down menu.

PR #34, opened 2018-06-14
Semi supervised

Edouard360:
* Semi supervised:

    * Add `get_indices` and `get_data_loaders` functions, useful for semi_supervised training

    * Correction of the semi-supervised training scheme, where if dataloaders were unbalanced (more labelled samples than unlabelled for instance), the actual loss minimized was wrong. Now the two new training schemes are

        * `train_semi_supervised_jointly` : the `classification_ratio` parameter can trigger a difference between the relative importance of classification vs. reconstruction loss

        * `train_semi_supervised_alternately` :  at the end of each epoch to optimize the ELBO, the classification error is, in turn, optimized.

    * Better naming convention: train/test => labelled/unlabelled

    * optional in SVAEC: `LinearLogRegClassifier` classifier, only updated every epoch (using sklearn). It acts like a NN on the forward pass - ie. for unlabelled samples,  computations are still entirely happening on GPU. This works best with the `train_semi_supervised_alternately` in the caseof the purified dataset populations

* Others:

    * added `get_latent_mean` and `get_latents` functions. (Only `get_latent` before). Still backward compatible, but removes the need to call .cpu().numpy(). Useful for models with more than one hidden stochastic layer.

    * Correct wrong multinomial formula

    * Add `base.py` file to normalize algorithms methods (VAE, VAEC, SVAEC)

### DIFF
--- a/examples/scVI-dev.ipynb
+++ b/examples/scVI-dev.ipynb
@@ -499,8 +499,8 @@
     "                         collate_fn = gene_dataset.collate_fn)\n",
     "\n",
     "latent, batch_indices, labels = get_latent(vae,data_loader)\n",
-    "latent=latent.cpu().numpy()\n",
-    "labels = np.concatenate(labels.cpu().numpy())\n",
+    "latent=latent\n",
+    "labels = np.concatenate(labels)\n",
     "\n",
     "if latent.shape[1] != 2:\n",
     "    latent = TSNE().fit_transform(latent)\n"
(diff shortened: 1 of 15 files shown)


jeff-regier:
Looks good. Will you add a Juptyer notebook to the examples directory (like the one Chenling made) to show how to use the new semi-supervised code?

Edouard360:
Yes, I can do that for the Cortex Dataset

jeff-regier:
```
/home/ubuntu/anaconda3/lib/python3.6/site-packages/matplotlib/__init__.py:1067: UserWarning: Duplicate key in file "/home/ubuntu/.config/matplotlib/matplotlibrc", line #2
  (fname, cnt))
```
You can get rid of this warning by deleting two of the three rows `/home/ubuntu/.config/matplotlib/matplotlibrc`.... not sure why they put them there.

You can get rid of the h5py warning by upgrading it to version 2.8 (through conda or pip).

PR #36, opened 2018-06-15
yl/remote loom

imyiningliu:
I tested the code with FISH dataset and the code runs. 

I didn't test it with `CORTEX.loom`. `http://loom.linnarssonlab.org/dataset/cellmetadata/Previously%20Published/Cortex.loom` leads me to a website with visualizations but not the `.loom` file :( .


jeff-regier:
Ah, it should be this one: http://loom.linnarssonlab.org/clone/Previously%20Published/Cortex.loom

imyiningliu:
generic LoomDataset class might need to take extra parameters because different `.loom` files name the `row` and `col` attributes differently. I set them as parameters so users can run `load_datasets` when they provide the `row_names` and `col_names` in the specific file. 

I think I can add extra parameters for `run_benchmarks` so that users can pass them in. It's kinda clumsy ... right now, `run_benchmarks.py` can run on FISH data using 
`python3 ./run_benchmarks.py --dataset=osmFISH_SScortex_mouse_all_cell.loom --gene_row=Gene --cluster_col=ClusterID --url=http://linnarssonlab.org/osmFISH/osmFISH_SScortex_mouse_all_cells.loom`

conda-recipes/scvi/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - torch ==0.4.0
+    - matplotlib ==2.2.2
+    - scikit-learn ==0.19.1
+    - scipy ==1.0.0
+    - h5py
+    - codecov
+    - pandas ==0.23.0
+    - loompy ==2.0.9
+  run:
+    - python
```
jeff-regier: We probably want these requirements to match these: https://github.com/YosefLab/scVI/blob/master/setup.py#L12-L20

They are the minimal required for our package to run.

conda-recipes/scvi/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+  fn: '{{ name }}-{{ version }}.{{ file_ext }}'
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.{{ file_ext }}
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  number: 0
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+
+requirements:
+  host:
```
jeff-regier: What's the difference between `host` and `run`?

conda-recipes/scvi/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - loompy ==2.0.9
+
+test:
+  imports:
+    - scvi
+    - scvi.dataset
+    - scvi.metrics
+    - scvi.models
+  requires:
+    - pytest
```
jeff-regier: Does this run our tests? How do you decide what to put in `imports`?

conda-recipes/scvi/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+about:
+  home: https://github.com/YosefLab/scVI
+  license: MIT
+  license_family: MIT
+  license_file: 'LICENSE'
+  summary: Single-cell Variational Inference
+  description: Single-cell Variational Inference
+  doc_url: https://scvi.readthedocs.io
+  dev_url: https://github.com/YosefLab/scVI
+
```
jeff-regier: It'd be better to put the conda stuff in a separate pull request than the remote loom stuff.

jeff-regier: (That keeps makes it easier to revert one change without the other, and it's more readable for anyone looking through the git history to see how repository has evolved.)

jeff-regier: FYI, you might use `git cherry-pick` to move the conda commit to another branch. See also `git rebase -i` and `git push origin yl/remote_loom`.

scvi/dataset/retina.py
```diff
@@ -14,4 +14,5 @@ def __init__(self, save_path='data/'):
             self.download()
 
         print("Finished preprocessing Retina dataset")
-        super(RetinaDataset, self).__init__(filename=self.download_name, save_path=self.save_path)
+        super(RetinaDataset, self).__init__(filename=self.download_name, save_path=self.save_path,
+                                            batch_col="Batch_id", cluster_col="Clusters")
```
jeff-regier: Can we change the field name "Batch_id" (in `retina.loom`) to whatever convention Sten uses in `cortex.loom`, rather than allowing it to vary?

Or have you seen loom files (other than our `retina.loom` dataset) that refer to `batch_col` and `cluster_col` is multiple ways? What do they call them besides "Batch_id` and "Clusters"?

tests/test_scvi.py
```diff
@@ -6,44 +6,63 @@
...
 
 def test_loom():
-    run_benchmarks(dataset_name="retina.loom", n_epochs=1, show_batch_mixing=False, save_path='tests/data/')
+    retina_dataset = load_datasets("retina.loom", save_path='tests/data/',
+                                   batch_col='Batch_id', cluster_col='Clusters')
+    run_benchmarks(retina_dataset, n_epochs=1, show_batch_mixing=False)
+
+
+def test_remote_loom():
+    FISH_dataset = load_datasets("osmFISH_SScortex_mouse_all_cell.loom", gene_row='Gene',
```
jeff-regier: More common to use lower case for a variable name, .e.g. `fish_dataset`. Helps distinguish it from classes.

conda-recipes/scvi/meta.yaml
imyiningliu: `host` represents packages that need to be specific to the target platform when the target platform is not necessarily the same as the native build platform. `run` include packages required to run the package. (They are automatically generated when I build conda package from pip)

conda-recipes/scvi/meta.yaml
imyiningliu: I found a documentation on [building conda from pip package](https://conda.io/docs/user-guide/tutorials/build-pkgs-skeleton.html). The recipes are automatically generated from `pip` package. It words when I tried building `pyinstrument` from `pip`, but I got the `pytorch` error when I tried building `scvi`. :(

conda-recipes/scvi/meta.yaml
jeff-regier: Can we specify the "python 3.6" is required here? Users could be frustrated if the think scVI will work with their version, but they get an error.

conda-recipes/scvi/meta.yaml
```diff
@@ -0,0 +1,82 @@
+{% set name = "scvi" %}
+{% set version = "0.1.1" %}
```
jeff-regier: The version number is 0.1.2 now. We should make  a small change to the `setup.py` so that this version number gets incremented automatically when we type `bumpversion patch`.

conda-recipes/scvi/meta.yaml
imyiningliu: Yup will do that ☺️

conda-recipes/scvi/meta.yaml
jeff-regier: I don't see host used in other people's conda recipes, so it might not be something we want. Maybe better to omit it and see if that creates problems?

scvi/dataset/retina.py
imyiningliu: One thing I thought of is, for `FISH` data, two columns can be used for `labels`: `ClusterID` and `ClusterName`. 
Also, when users create the `.loom` files, they can specify column and row names. It might be a problem when people use small variations, like `clusterID`, `cluster_id`, or `Cluster_ID`, which might crush the code. 
But I can change the `retina.loom` and remove the parameters if that's needed ^ ^

conda-recipes/scvi/meta.yaml
imyiningliu: Yup 😊 I make these changes and pushed to `yl/conda`.

scvi/dataset/retina.py
jeff-regier: Do `ClusterID` and `ClusterName` specify different clustering, or is one just an integer and one a string? If the former, let's follow the conventions in the existing loom data for `retina.loom` --- `ClusterID` rather than `Clusters`. And then assume for now that loom data will always use `ClusterID` for the cluster column, so that you don't have to pass `cluster_col` around all over the place.  (Same for `batch_col`.)

imyiningliu: They are the same - one is numbers the other is string. Sounds good! I'll make these changes.

conda-recipes/scvi/meta.yaml
jeff-regier: I see. Actually `pytorch` exists in both the default channel and the pytorch channel, so I'm not surprised this didn't help.

jeff-regier: You might post the question on stackoverflow. Seems like there's a got to be a way....

imyiningliu:
`run_benchmarks` works on FISH by running ` ./run_benchmarks.py --dataset=osmFISH_SScortex_mouse_all_cell.loom --url=http://linnarssonlab.org/osmFISH/osmFISH_SScortex_mouse_all_cells.loom` 

`python3 ./run_benchmarks.py --dataset=Cortex.loom --url=http://loom.linnarssonlab.org/clone/Previously%20Published/Cortex.loom` works too. Loading cortex data looks fine. However, `CortexDataset` only take `558` most expressed genes, but the generic `LoomDataset` takes all genes: do we want to add an extra parameter to select genes?

conda-recipes/scvi/meta.yaml
imyiningliu: Okay. 👌🏻

jeff-regier:
I suppose, can we use the same code that brain_large uses for subsampling?

imyiningliu:
I think so! (Double check with @Edouard360: does `subsample_genes` select the `n` number of genes with highest variance?)

jeff-regier:
OK please "Squash and merge" this PR whenever it's ready.

imyiningliu:
Sorry, does this `Cortex.loom` have the exact same data as the one loaded in `CortexDataset`? 
I only selected the first `558` genes with highest variance. The dimension looks fine and the code runs, but the expression matrix looks different from the one generated by `CortexDataset`. I checked the expression matrix without filtering genes and the matrix are different. I'll look into this if they are supposed to be exactly the same.

jeff-regier:
It's fine I think

PR #37, opened 2018-06-15
Datasets10x

Edouard360:
Noticed some common structure in the organisation/structure of the 10x datasets.

`Dataset10X("pbmc8k"), Dataset10X("t_3k") , Dataset10X("frozen_pbmc_donor_a")` all worked for me.

These will be useful for reproducing some harmonisation results. 

@imyiningliu : that also include `BrainSmallDataset` if I do `Dataset10X("neuron_9k")`, but you might want to include some additional parts in preprocessing (for instance selecting the highest most variable genes).

Also, needs unit tests for coverage

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -15,7 +15,7 @@ Types of Contributions
 Report Bugs
 ~~~~~~~~~~~
 
-Report bugs at https://github.com/romain-lopez/scvi/issues.
+Report bugs at https://github.com/YosefLab/scVI/issues.
 
 If you are reporting a bug, please include:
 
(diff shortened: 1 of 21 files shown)


jeff-regier:
Great, let's merge this as is. @imyiningliu If you'll use this for the `BrainSmallDataset` then `test_brain_small` will start testing the code.

@Edouard360 Do we need to test it with additional datasets that have slightly different formats than `brain_small`? Or do these 10x datasets have about the same format?

jeff-regier:
Also, @Edouard360 what about `brain_large`? Is that format very different than these datasets?

Edouard360:
All of the aforementioned datasets come in the same format with 3 files `matrix.mtx`, `genes.tsv`, `barcodes.tsv`. brain_large's dataset contains a h5 format however, which cannot be currently handled by Dataset10X. But if other datasets are also in .h5 format, it might be worth adding an extra specification at \_\_init\_\_ time (using something similar to the mapping in the object `available_datasets`).

PR #40, opened 2018-06-15
CiteSeqDataset

Edouard360:
I suggest this to complete CbmcDataset information. 

* change CbmcDataset dataset to CiteSeqDataset with name option "cbmc"or "pbmc" and info about **ADT counts (proteins markers)**, added as attributes of the dataset.

This is the data from **epitopes** useful for having further labelling information.


Edouard360:
I haven't had the time to create the preprocessed files for unit tests, but:

`CiteSeqDataset("pbmc")` and `CiteSeqDataset("cbmc")` both work and also have as attributes information about the epitopes

tests/test_scvi.py
```diff
@@ -30,7 +30,7 @@ def test_retina():
 
 
 def test_cbmc():
-    run_benchmarks("cbmc", n_epochs=1, show_batch_mixing=False, save_path='tests/data/')
+    run_benchmarks("cite_seq_cbmc", n_epochs=1, show_batch_mixing=False, save_path='tests/data/')
```
jeff-regier: Is `cite_seq_cbmc` the same data as `cbmc` once it's loaded? Can we just continue to call it `cbmc` then?

scvi/dataset/cite_seq.py
```diff
@@ -0,0 +1,46 @@
...
+
+available_datasets = {
+    "cbmc": "CBMC_8K_13AB",
+    "pbmc": "PBMC_vs_flow"
+}
+
+
+class CiteSeqDataset(GeneExpressionDataset):
+    def __init__(self, name='cbmc', save_path='data/citeSeq/'):
+        self.save_path = save_path + name + '/'
```
jeff-regier: `self.save_path` should equal `save_path`. Maybe call it `save_path` something else if we can't just store it directly under `save_path`.

tests/test_scvi.py
jeff-regier: I thought Romain preprocessed the cite seq data that we're downloading now. @Edouard360 @imyiningliu Is this data just as good? If so, I'd rather use it since it's from an official source.

imyiningliu: I think `cbmc` files are downloaded directly from the source.

Edouard360:
- Datasets might have multiple urls from which to download (ex. cite-Seq data): we might either specify `url`, `download_name` attributes or `urls`, `download_names`. Move check if file exists in `_download`.

- `CbmcDataset()` -> CiteSeqDataset('cbmc'), with information about epitopes.

- `PbmcDataset()` can be obtained with:

```
gene_dataset = concat_datasets(
     Dataset10X("pbmc8k", save_path=save_path),
     Dataset10X("pbmc4k", save_path=save_path)
)
```

So I removed data/PBMC

- From citeSeq methods there are actually 3 available datasets (cmbc, pbmc, and cd8). Since there are also 10X pbmc datasets, the `pbmc` name is misleading in the `load_datasets` function. For now we leave as default romain's initial pbmc dataset, which consists in the concatenation of `pbmc8k` and `pbmc4k`

- `concat_datasets` test

Edouard360:
closes #47

jeff-regier:
Very nice!

PR #41, opened 2018-06-15
set up conda installation 

imyiningliu:
This is for `version0.1.1`. I think currently it always has to synchronize with `pip`. 
I can upload to Anaconda by running `anaconda upload /Users/yiningliu/miniconda3/conda-bld/osx-64/scvi-0.1.1-py36_0.tar.bz2`.


conda-recipes/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - pytest-runner ==2.11.1
+    - numpy ==1.14.0
+    - pytorch ==0.4
+    - matplotlib ==2.2.2
+    - scikit-learn ==0.19.1
+    - scipy ==1.0.0
+    - h5py
+    - codecov
+    - pandas ==0.23.0
+    - loompy ==2.0.9
```
jeff-regier: Just the minimal requirements go here.

imyiningliu: The build errors if I use minimal requirement file :(. I suspect the error comes from the inconsistency between `requirements.txt` and version `0.1.1`'s `setup.py`, since `conda` package is built from `pip`, and `0.1.2` hasn't been uploaded on `pip`. 

I think I can either 
1) Use the current `meta.yml` to upload `0.1.1` to `conda`, or 
2) Upload the current `scvi` as `version 0.1.2` to `pip`, and then try building the new version with minimal requirements for `conda`. (I've put the minimal requirements into a separate folder locally for now).

jeff-regier: Uploading 0.1.2 to pip is fine

imyiningliu:
I updated `meta.yml` with minimal requirements. The build fails if I don't include `pytest-runner ==2.11.1`. The rest matches with `requirements.txt`. I'll run the command to upload the package to Anaconda if it's ready.

setup.py
```diff
@@ -2,13 +2,16 @@
 # -*- coding: utf-8 -*-
 
 from setuptools import setup, find_packages
+import scvi
```
jeff-regier: I'm not sure we can import scvi here. This is the script that runs _to install_ scvi. If it imports scvi, that means scvi needs to already be installed.   Are there other examples of `setup.py` files that import the package they are installing?

conda-recipes/scvi/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+  fn: '{{ name }}-{{ version }}.{{ file_ext }}'
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.{{ file_ext }}
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  number: 0
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+
+requirements:
+  host:
```
jeff-regier: Let's get rid of the host section --- unless you're sure it helps. I don't see other conda packages using it.

conda-recipes/scvi/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - numpy >=1.0
+    - pytorch >=0.4
+    - matplotlib >=2.0
+    - scikit-learn >=0.18
+    - scipy >=1.0
+    - pandas >=0.2
+    - loompy >=2.0
+  run:
+    - python 3.6
+    - pytest-runner ==2.11.1
```
jeff-regier: Maybe you're supposed to use the `test: requires:` section for pytest-runner?  I'm pretty sure pytest-runner isn't supposed to be here.

jeff-regier: Also, good to omit the version number here for pytest-runner because any version should do.

jeff-regier: See https://github.com/rmcgibbo/python-appveyor-conda-example/blob/master/conda-recipe/meta.yaml

nose is an alternative to pytest

setup.py
imyiningliu: Sorry you're right 😊 That might cause a problem. I found another way of reading `__version__` from `scvi.__init__.py` and I just pushed the code with the new solution.

conda-recipes/scvi/meta.yaml
imyiningliu: I got the following errors if I don't include `pytest-runner` in `host` section: 
`RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml  url=https://pypi.org/simple/pytest-runner/` 
and 
`subprocess.CalledProcessError: Command '['/bin/bash', '-e', '/Users/yiningliu/miniconda3/conda-bld/scvi_1529258385675/work/conda_build.sh']' returned non-zero exit status 1.` 

I got rid of the other lines in `host`. With the newly pushed 'meta.yml`, `conda-build scvi` runs without errors.

setup.py
```diff
@@ -2,13 +2,20 @@
 # -*- coding: utf-8 -*-
 
 from setuptools import setup, find_packages
+from distutils.util import convert_path
+
+main_ns = {}
+ver_path = convert_path('scvi/__init__.py')
+with open(ver_path) as ver_file:
+    exec(ver_file.read(), main_ns)
```
jeff-regier: I don't get it...why are we changing setup.py at all in this pull request?

jeff-regier:
It looks like Cortex.loom is included in this pull request -- accidentally I think. That file is already committed to master. You might type `git rebase master` for this branch.

setup.py
imyiningliu: It will be easier to update the version number (when I updated to `pip`). We need to change `version` in two places (`setup.py` and `scvi.__init__.py`) before. Now we only need to change it in `scvi.__init__.py`. 😊 But I can change it back haha. Sorry about this 😔

jeff-regier: Yes, better to change it back. Just also use `bumpversion` to change the version numbers. It will change them everywhere.

imyiningliu: Thanks!

PR #50, opened 2018-06-20
add back Cortex.loom test file

imyiningliu:


jeff-regier:
How did the tests pass without it?

imyiningliu:
I passed in the `url` in `test_scvi.py`. It downloads from the web and puts the raw dataset file in `tests/data/` as `Cortex.loom` :(

jeff-regier:
Got it. You'll change it to use the test data for the tests?

imyiningliu:
Yup! I added back the `Cortex.loom` in `tests/data/` so we wouldn't download the raw data when we run unit test. Sorry for accidentally removing it from the repo before.

jeff-regier:
Oh I see. It's no problem at all...I was just trying to understand the change.

PR #54, opened 2018-06-24
Add CsvDataset and make CbmcDataset a subclass of CsvDataset 

imyiningliu:
I also fixed styles so that the parameter names for the generic Datasets (`CsvDataset`, `LoomDataset` and `Dataset10X`) are more consistent.


scvi/dataset/cbmc.py
```diff
@@ -1,32 +1,12 @@
...
-        expression = pd.read_csv(self.save_path + self.download_name, index_col=0, compression='gzip').T
-        gene_names = np.array(expression.columns, dtype=str)
-
-        selected = np.std(expression.values, axis=0).argsort()[-600:][::-1]
-        expression_data = expression.values[:, selected]
-
-        return expression_data, gene_names
+        super(CbmcDataset, self).__init__(filename='GSE100866_CBMC_8K_13AB_10X-RNA_umi.csv.gz',
+                                          save_path=save_path,
+                                          p_genes=600,
```
jeff-regier: Why 600? Could we make 600 the default value for a keyword argument to the constructor?

scvi/dataset/dataset.py
```diff
@@ -77,6 +77,8 @@ def collate_fn(self, batch):
             torch.LongTensor(self.labels[indexes])
 
     def subsample_genes(self, p_genes=1., subset_genes=None):
+        if p_genes > self.X.shape[1] or p_genes == 1.:
```
jeff-regier: Should it be `>=` instead of `>`? Why is `p_genes == 1` a special case?

jeff-regier: Is p_genes a proportion or a count?

tests/test_scvi.py
```diff
@@ -61,17 +61,24 @@ def test_loom():
...
 
 
 def test_anndata():
     ann_dataset = AnnDataset("tests.h5ad", save_path='tests/data/')
     run_benchmarks(ann_dataset, n_epochs=1, show_batch_mixing=False)
+
+
+def test_csv():
+    csv_dataet = CsvDataset("GSE100866_CBMC_8K_13AB_10X-RNA_umi.csv.gz", save_path='tests/data/', compression='gzip')
+    run_benchmarks(csv_dataet, n_epochs=1, show_batch_mixing=False)
```
jeff-regier: Looks good

scvi/dataset/dataset.py
imyiningliu: Since I called `subsample_genes` in all `__init__`, I don't want subsampling happens when `p_genes` is 1. 
`p_genes` can be either. If it's smaller than 1, then it's a proportion. If it's an integer greater than 1, then it's a count.

scvi/dataset/cbmc.py
imyiningliu: Romain chose the 600 genes with highest variance in the paper. Do we want to make 600 the default value? Or will users want to include all genes for most of the cases? 

`brain_small` is similar - I passed in `p_genes=3000` in the constructor 😊

jeff-regier: Yes, please make 600 the default value to a keyword argument taken by the constructor.

scvi/dataset/dataset.py
jeff-regier: `p_genes` should always be either a proportion or a count. It's confusing to use it as both--easy to misunderstand for anyone besides us. Will you find another way to do it?

scvi/dataset/cbmc.py
imyiningliu: Do we also make 3000 the default value for `loom.py` (the same issue for `brain_small`)?

scvi/dataset/dataset.py
imyiningliu: Yes. Will fix it now. 😊

Edouard360: In sklearn `train_test_split` function, there is a `train_size` argument that can either be float or int. That's what motivated my choice. Maybe the terminology `gene_size` would have been better suited. In any case, it's rather true that we only think in terms of the number of genes (rather than the proportions compared to initial number).

PR #57, opened 2018-06-26
add data loading examples 

### LINKED ISSUE #42: notebook of data loading examples
Create a notebook named `examples/data_loading.ipynb` that shows users all the ways they can load data into scVI. They can load
- a loom file
- a csv file
- a 10x file
- any of our "built in" datasets  (list them, and give a little information about each of them)

imyiningliu:
Closes #42 

Also make the print statements more consistent among datasets.


jeff-regier:
Would rebase on master? You can do
```
git pull --rebase origin master
```
followed by a `git push origin yl/loading -f`

imyiningliu:
Should I change `cbmc` and `pbmc` so that they're subclasses of `CiteSeqDataset`?

jeff-regier:
That sounds good but please check with Eddie

Edouard360:
@imyiningliu:  Maybe subclass both of them in the same file as `CiteSeqDataset`.

scvi/dataset/brain_large.py
```diff
@@ -1,5 +1,5 @@
 import collections
-import time
+# import time
```
jeff-regier: Please delete unused code rather than commenting it out

jeff-regier:
Why the drop in unit test coverage?

jeff-regier:
You can get rid of the warning below by upgrading h5py to version 2.8. The notebook will look better without the warning.

```
/anaconda3/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
```

jeff-regier:
Better to align arguments for multi-line function calls, e.g.,

```
loom_dataset = LoomDataset("Cortex.loom", 
                            save_path='data/',
                            url='http://loom.linnarssonlab.org/clone/Previously%20Published/Cortex.loom')
```

jeff-regier:
What about `AnnDataset`? Can you add that to the generic datasets section of the notebook?

jeff-regier:
In the notebook, would you also show how to load a `CsvDataset` and a `LoomDataset` where the data is stored locally (i.e., without using the `url` keyword argument)? That's probably the more common use case.

jeff-regier:
Would you add more information to the notebook about origins of each of the built in datasets? You can pretty much copy from section 2.2 of
https://www.biorxiv.org/content/biorxiv/early/2018/03/30/292037.full.pdf

Also we should probably include citations for each of these datasets in the notebook. Those are in section 2.2 too.

jeff-regier:
The notebook looks good @imyiningliu . Thanks for working on it. Let's ask @chenlingantelope and Nir to look over it too once you've made the revisions above -- they may have a better idea of what our future users will like.

imyiningliu:
I've copied section 2.2 (but i broke up the description for each dataset into two parts. One part is in 'description' of the 7 datasets and the other is mentioned when the dataset is loaded). Should I combine them into one section? Or do I only need to add the citations?

jeff-regier:
It'd be nice to have one section (or subsection) for each of the 7 datasets.

imyiningliu:
I'm having some trouble with Jupyter notebook - I'm not sure how to change the Python interpreter. Jupyter notebook uses a different interpreter so I wasn't able to import `anndata` and the new version of `h5py`. Can I ask you tomorrow about this?

jeff-regier:
I'd rather not add the old cbmc test files back if we can avoid using them. Why does code coverage decrease without them?

imyiningliu:
I'm not too familiar with reading Codedev report  > < Does it say the current code coverage increases by +0.27%? 

I added unit tests for `CbmcDataset` and `PbmcDataset`, and `CbmcDataset` class needs test files, which are under `tests/data/citeSeq/cbmc/`. The old `cbmc` test file is `GSE100866_CBMC_8K_13AB_10X-RNA_umi.csv.gz`, which is currently used to test `CsvDataset`.

jeff-regier:
Yes, it looks like coverage is up 0.27% now.

jeff-regier:
Let's not worry about those test files -- it was a small detail anyway. The notebook and the whole PR looks really good now!

PR #72, opened 2018-06-29
resolves #64

### LINKED ISSUE #64: run_benchmarks uses the same VAE architecture for all datasets
It looks like `run_benchmarks` uses the same VAE architecture (e.g. default n_layers) for every datasets:
https://github.com/YosefLab/scVI/blob/cbd6f41bc6e40bbb0c1082ec9bf1d5b32bcd3f7d/scvi/benchmark.py#L38-L39

Instead, let's instantiate the VAE class with the number of layers etc from Table 2:
https://www.biorxiv.org/content/biorxiv/early/2018/03/30/292037.full.pdf

Edouard360:
- might be extended to give different hyperparameters to the `model` and the `train` function

### DIFF
--- a/scvi/dataset/benchmark_hyperparameters.py
+++ b/scvi/dataset/benchmark_hyperparameters.py
@@ -1,0 +1,19 @@
+from . import BrainLargeDataset
+
+
+def benchmark_hyperparameters(gene_dataset):
+    """
+    Possibly specify hyper parameters according to Table2 in "Bayesian Inference for a Generative Model of
+    transcriptome Profiles from Single-cell RNA Sequencing"
+    :param gene_dataset:
+    :return:
+    """
+    hyperparameters = dict()
+    if isinstance(gene_dataset, BrainLargeDataset):
+        if gene_dataset.nb_genes > 10000 and gene_dataset.nb_genes < 15000:
+            hyperparameters["n_layers"] = 2
+        elif gene_dataset.nb_genes > 15000:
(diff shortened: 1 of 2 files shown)


scvi/dataset/benchmark_hyperparameters.py
```diff
@@ -0,0 +1,19 @@
...
+    :return:
+    """
+    hyperparameters = dict()
+    if isinstance(gene_dataset, BrainLargeDataset):
+        if gene_dataset.nb_genes > 10000 and gene_dataset.nb_genes < 15000:
+            hyperparameters["n_layers"] = 2
+        elif gene_dataset.nb_genes > 15000:
+            hyperparameters["n_layers"] = 3
+            hyperparameters["n_hidden"] = 256
+
```
jeff-regier: @Edouard360 Better to put this code in `benchmark.py` rather than the `dataset` directory.

Edouard360: yes indeed, just changed it. Otherwise should I keep it as a function ?

jeff-regier: Yes, I think it's fine. I appreciate you fixing issue #64 . 

Eventually we may want to define a `Benchmark` class -- each instance has a vae, a dataset, a name string (e.g., `brain_large`), and some hyperparameters. To make this approach work, we'd need to stop downloading and preprocessing the dataset whenever it's `__init__` method is called.

PR #85, opened 2018-07-05
gene names should be in var index

### LINKED ISSUE #84: anndata "obs" are cells, not genes
Had some issues loading our data and realized that there was a transpose issue. Looking at [the code](https://github.com/YosefLab/scVI/blob/master/scvi/dataset/anndata.py#L24), I believe there's a bug. The `obs` of an anndata object should be the cells, not the genes. The genes are `var`. So on the plus side for you, no transpose is necessary for loading.

Correct me if I'm wrong, as I may have just confused myself...

jamestwebber:
Fixes #84


jeff-regier:
Change made in #88

PR #86, opened 2018-07-05
Updated the early stopping criterion.

maxime1310:
The early stopping criterion was not very useful because it was too restrictive. Now it makes more sense, and is based on the same criterion used in other ML libraries (Keras,...)


jeff-regier:
Great! I was just noticing that before it seemed to always stop early after 500 iterations, but I wasn't sure why.

PR #89, opened 2018-07-07
harmonization for n_layer==1

Edouard360:
- sequential layers are [dropout(0.1), linear, batchnorm, relu]. Before, we concatenated [latent_features, one_hot_batch_index] as input to the sequential layer decoder, therefore submitting one_hot_batch_index to the dropout as well, and giving poor batch mixing entropy results. We customize the forward function to circumvent that problem.

- also simplifying the modules surrounding FCLayers. The additional feature is the ability to easily add edges from categorical variables in the graphical model, by specifying n_cat in the construct for each variable as a list (n_cat_list), and providing them as a sequence in the forward function.

Latent space vis. as supporting evidence ! (BE ~=0.6)
new-3

### DIFF
--- a/scvi/models/modules.py
+++ b/scvi/models/modules.py
@@ -8,88 +8,51 @@ from scvi.models.utils import one_hot
 
 
 class FCLayers(nn.Module):
-    def __init__(self, n_in, n_out, n_hidden=128, n_layers=1, dropout_rate=0.1):
+    def __init__(self, n_in, n_out, n_cat_list=[], n_layers=1, n_hidden=128, dropout_rate=0.1):
         super(FCLayers, self).__init__()
         layers_dim = [n_in] + (n_layers - 1) * [n_hidden] + [n_out]
-        self.fc_layers = FCLayers._sequential(layers_dim, dropout_rate=dropout_rate)
-
-    def forward(self, x, *os):
-        return self.fc_layers(x)
-
-    @staticmethod
-    def create(n_in, n_out, n_cat=0, n_hidden=128, n_layers=1, dropout_rate=0.1):
-        if type(n_cat) is int:
-            if n_cat == 0:
-                return FCLayers(n_in, n_out, n_hidden=n_hidden, n_layers=n_layers, dropout_rate=dropout_rate)
-            else:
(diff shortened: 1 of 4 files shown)


Edouard360:
Looks like I merge+delete the branch a little fast so that the second check didn't pass (cannot find the branch related to pull request), but https://travis-ci.org/YosefLab/scVI shows the merge tests on master passed successfully.

All is OK in theory

scvi/models/modules.py
```diff
@@ -8,88 +8,51 @@
...
-    @staticmethod
-    def _sequential(layers_dim, n_cat=0, dropout_rate=0.1):
-        return nn.Sequential(collections.OrderedDict(
+        self.n_cat_list = [n_cat if n_cat > 1 else 0 for n_cat in n_cat_list]  # n_cat = 1 will be ignored
+        self.fc_layers = nn.Sequential(collections.OrderedDict(
             [('Layer {}'.format(i), nn.Sequential(
                 nn.Dropout(p=dropout_rate),
-                nn.Linear(n_in + n_cat, n_out),
+                nn.Linear(n_in + sum(n_cat_list), n_out),
                 nn.BatchNorm1d(n_out, eps=1e-3, momentum=0.99),
```
jeff-regier: Dropout usually happens after fully connected layers--not before. Should these be reordered?

Edouard360: This was the `dense` layer implementation in Romain's code. 

> Typically, dropout is applied after the non-linear activation function (a). However, when using rectified linear units (ReLUs), it might make sense to apply dropout before the non-linear activation (b) for reasons of computational efficiency depending on the particular code implementation.

in https://sebastianraschka.com/faq/docs/dropout-activation.html.

Edouard360: We could move it btw Linear and BatchNorm, or after BatchNorm ? What would you suggest, and what results should we look at for making such change ? I could always run the mixing experiment which results seem to be dropout-dependent. (It always messes up for dropout=0)

jeff-regier:
This order is common: Linear, BatchNorm, ReLU, Dropout.

Also, I usually wouldn't use any dropout or BatchNorm on the final layers of the encoder and the decoder.

Issue #79, opened 2018-07-01
run groups of benchmarks with `run_benchmarks.py`


jeff-regier:
Currently `run_benchmarks.py` just runs one benchmark per call. It'd be nice if it could, optionally, run groups of benchmarks.

For example,
```
./run_benchmarks.py --annotation
```
might run all the annotation (semi-supervised) benchmarks, and then print out a nice table afterwards with all the annotation results for all the dataset.

And
```
./run_benchmarks.py --harmonization
```
might run all the unsupervised harmonization benchmarks. 

And 
```
./run_benchmarks.py --basic
```
might run all of the original seven benchmarks.

And 
```
./run_benchmarks.py --all --epoch 1
```
would be useful for testing that everything works.

One way to implement this without a mess of `if` statements (c.f., `load_dataset`) might be to create a new `Benchmark` class. It would have as fields all the arguments to the `train()`, including one dataset instance (e.g. a `CbmcDataset`) and one model instance (e.g. a `VAE` instance). These `Benchmark` objects could then be organized into groups.

jeff-regier:
I removed mention of `run_benchmarks.py` script from the readme. The API is the primary way of interacting with scVI now -- not that executable `run_benchmark.py` script.

PR #91, opened 2018-07-11
Refactoring

### LINKED ISSUE #75: refactoring: remove use_cuda from models
It's better if the models (i.e. VAE, VAEC, SVEAC) don't "know" whether they are using cuda or not. In the `__init__` method for these models, we can just load everything on the CPU. Then, call the `.cuda()` method (e.g. `vae.cuda()` right after instantiating the model. It's kind of a detail, but the distinction between the model and what chip it runs on is a helpful one to maintain.

See

https://github.com/YosefLab/scVI/blob/704ce19d80e5728542464e488968c136a241eb6f/scvi/models/vae.py#L46-L48

https://github.com/YosefLab/scVI/blob/704ce19d80e5728542464e488968c136a241eb6f/scvi/models/vaec.py#L40-L43

https://github.com/YosefLab/scVI/blob/704ce19d80e5728542464e488968c136a241eb6f/scvi/models/svaec.py#L47-L50

### LINKED ISSUE #76: refactoring: VAEC and SVAEC as subclasses of VAE
There's a lot of duplicated code in the VAE, VAEC, and SVAEC classes. To get rid of it, lets make VAEC and SVAEC subclasses of VAE. In addition to avoiding duplicated code (which is hard to maintain), it's nice conceptually to express our semi-supervised models as specializations or our unsupervised scVI model.

Also, I wonder about whether SVAEC could inherit from VAEC, or whether even we need both SVAEC and VAEC.

### LINKED ISSUE #77: refactoring: combine training method
There's a lot of code duplicated across our four methods for training models: `train`, `train_semi_supervised_jointly`, `train_semi_supervised_alternately`, and `train_classifier`. How about combining all these methods into one method, `train`, and passing it additional arguments (e.g., boolean flags) to control how it behaves?

### LINKED ISSUE #87: refactoring: no import torch in the example notebooks
As discussed, let's refactor the scvi package so we don't have to `import torch` or anything from torch in the notebooks. That keeps users from having to interact directly with data loaders, and it should make #79 straightforward. One way to achieve this is to add a new class to scVI that contains as members both one dataset and one model. These two thing together define a posterior distribution (implicitly). `Inference` might be a reasonable name for the class. We could put nearly all the scvi code except what's under `models` and `datasets` into this class. `Inference` could either have a `Trainer` instance as a class member, or else `Inference` could subsume `Trainer`. In the latter case, w'd have multiple types of `Inference` classes, like `SemiSupervisedInference`, each with a `train` method. I'm not really sure which way is better.

Edouard360:
Refer to #87 for the discussion about the design changes. There is the new convenient `Inference` class (see updated notebooks for usage).

- closes #87, closes #75, closes #76, closes #77 

#79 is halfway through: ideally we would want to write explicitely all the detailed benchmarks for all datasets, and determine what's under --annotation and --harmonization. I started in `scvi/benchmark.py` file, and we might further define all the benchmarks related functions here (?)


Edouard360:
Ah, need to increase coverage before merging !

Edouard360:
Coverage is improved. @jeff-regier : should I squash and merge already ? The other option is to wait a few days, while I am continuing development from this branch (running PBMCs benchmarks + invertible function + other things we mentionned), which might help anticipate a few bugs, perhaps.
(Also #79 is not completely done yet, but I want might focus on this later on (full reproducibility of the results))

jeff-regier:
This all looks really good -- I'm very glad not to have duplicated code any more.  

The "plugin" classes are a bit confusing, e.g. `SemiSupervisedVariationalInferencePlugin`. Is there no way to support all the experiments we have too support without them? (without duplicating code, that is.) For example, could we have just one hierarchy of `Inference` classes and no plugins?

jeff-regier:
So `VariationalInference` inherits from (or "mixes in") `VariationalInferencePlugin`, and
`SemiSupervisedVariationalInference` inherits from `SemiSupervisedVariationalInferencePlugin` which in turn inherits from `VariationalInferencePlugin`. 

Could we just put all the methods from `VariationalInferencePlugin` into `VariationalInference` and get rid of both `VariationalInferencePlugin` and `SemiSupervisedVariationalInferencePlugin`?

scvi/metrics/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+            if verbose:
+                print("Entropy batch mixing :", be_score)
+            return be_score
+
+    entropy_batch_mixing.mode = 'max'
+
+    def show_t_sne(self, name, **kwargs):
+        n_samples = kwargs.pop('n_samples')
+        latent, batch_indices, labels = get_latent(self.model, self.data_loaders[name], use_cuda=self.use_cuda)
+        show_t_sne(latent, np.array([batch[0] for batch in batch_indices]), n_samples=n_samples)
```
jeff-regier: Is this the only place `show_t_sne` (the function, not the method with the same name) is called? If so, can we delete the function and insert its code here? Just nice to have less indirection.

scvi/metrics/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class AccuracyPlugin:
+    def accuracy(self, name, verbose=False):
```
jeff-regier: Probably better to just add methods named `accuracy` to the three classes that (directly and indirectly) inherit from `AccuracyPlugin`. If all accuracy does is call `compute_accuracy`, there's almost no code duplicated across those three classes.  The code will be easier to understand and maintain if we can delete `AccuracyPlugin` and avoid multiple inheritance.

scvi/metrics/__init__.py
```diff
@@ -0,0 +1,93 @@
...
+            model = self.model
+            cls = None
+        acc = compute_accuracy(model, self.data_loaders[name], classifier=cls, use_cuda=self.use_cuda)
+        if verbose:
+            print("Acc for %s is : %.4f" % (name, acc))
+        return acc
+
+    accuracy.mode = 'max'
+
+    def svc_rf(self, **kwargs):
```
jeff-regier: Oh, I see now `svc_rf` is part of `AccuracyPlugin` too. This is tricky.

scvi/metrics/classification.py
```diff
@@ -32,38 +32,38 @@ def compute_accuracy_tuple(y, labels):
...
         sample_batch = sample_batch.type(torch.float32)
         all_labels += [labels.view(-1)]
 
-        if classifier is not None:
-            # Then we use the specified classifier
-            mu_z, _, _ = vae.z_encoder(sample_batch)
-            y_pred = classifier(mu_z).argmax(dim=-1)
-        else:
-            # Then the vae must implement a classify function
+        if hasattr(vae, 'classify'):
```
jeff-regier: Any way to avoid branching on `hasattr`? It's kind of complicated. Or may be there's no other way....

jeff-regier:
@Edouard360  Please squash merge whenever you're ready. It's a big improvement over what we had -- very nice work. Some of the class hierarchy feels a bit complicated to me, so that may be something to think about, though I'm not sure precisely how to do it differently. It may just be a limitation of OOP.

scvi/metrics/__init__.py
Edouard360: Hum I thought Chenling was using it in her notebook, but actually double checked and it is not used (she used the code inside). So it is fine to move it indeed.

scvi/metrics/classification.py
Edouard360: agreed ! this is certainly bad design. If `.accuracy` was a function of the inference object, there wouldn't be any branching. Probably better to do it that way !

scvi/metrics/__init__.py
Edouard360: I could probably ignore the fact the ClassifierInference and SemiSupervisedVariationalInference might inherit common methods like `.accuracy` or `.svc_rf`.
Not even sure if this `.svc_rf`, (or other benchmarking functions we might use), should be set as a method for the infer object.

Edouard360:
I understand and agree with your remarks about the `SemiSupervisedVariationalInferencePlugin` and `VariationalInferencePlugin`. I wanted to separate the training logic (fit, loss, on_epoch_end) from the functional logic (`.ll`, `.differential_expression`, `.batch_entropy_mixing`) of the `Inference` class, but it might actually be more readable to include them directly in the class themselves.

This would actually simplify it I think. Only the "accuracy" plugin was duplicated code, but again since there is branching on this method I might define them separately in each of the Inference class.

scvi/metrics/__init__.py
Edouard360: Exactly, all the more since there is branching afterwards

scvi/dataset/utils.py
```diff
@@ -47,40 +50,137 @@ def get_indices(labels, n_labelled_samples_per_class_array):
...
-    """
-    From a list of data_loaders, return the numpy array suitable for classification with sklearn
-    :param data_loaders: a sequence of data_loaders arguments
-    :return: a sequence of (data, labels) for each data_loader
-    """
+def zip_first_cycle(*args):
+    return zip(args[0], *[cycle(arg) for arg in args[1:]])
+
+
+class DataLoaders:
```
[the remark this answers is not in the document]
jeff-regier: Oh, I guess `zip_first_cycle` could be moved to the `DataLoaders` class. Actually, if you just call `zip_first_cycle` once, since it's a one line function, you might just get rid of it...call `zip(...)` directly in `__iter__`.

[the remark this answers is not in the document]
Edouard360: #95 moves `zip_first_cycle` and renames `utils.py` to `data_loaders.py`.

`get_indices` is used in :
```python
class SemiSupervisedDataLoaders(DataLoaders):
    to_monitor = ['labelled', 'unlabelled']

    def __init__(self, gene_dataset, n_labelled_samples_per_class=50, seed=0, **data_loaders_kwargs):
        """
        :param n_labelled_samples_per_class: number of labelled samples per class
        """
        super(SemiSupervisedDataLoaders, self).__init__(gene_dataset, **data_loaders_kwargs)
        indices_labelled, indices_unlabelled = (
            get_indices(gene_dataset.labels, [n_labelled_samples_per_class] * gene_dataset.n_labels, seed=seed)
        )
```
To get labelled/unlabelled split. Should I move the function inside the \_\_init\_\_, since it is only used here ?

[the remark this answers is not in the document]
jeff-regier: Sure, that'd be great!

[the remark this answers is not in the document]
Edouard360: ok that's done ! I squash&merge

PR #93, opened 2018-07-13
Finish documentation for dataset classes 

imyiningliu:


jeff-regier:
Looks good. Is there a way to make it so that the documentation automatically gets updated at readthedocs when we merge a pull request?

imyiningliu:
It should be updated. I'm not sure why the documentation only renders the title but not the content - still looking into it.

jeff-regier:
The "badge" from readthedocs (appearing on our github project page) shows the documentation build is failing. I'm not sure how to view the error messages though. Can you see them?

imyiningliu:
Here's the [error message] (https://readthedocs.org/projects/scvi/builds/7484048/).

jeff-regier:
It looks like the problem is that the documentation is being compiled with Python 3.5. But loompy, unfortunately, requires Python 3.6. (The string interpolation syntax below was introduced with Python 3.6.)

```
    download_url=f"https://github.com/linnarsson-lab/loompy/archive/{__version__}.tar.gz",
                                                                                        ^
SyntaxError: invalid syntax
```

PR #97, opened 2018-07-17
Load additional dataset 

imyiningliu:
closes #97 

Seqfish dataset, Breast Cancer dataset and MouseOB dataset mentioned in [this paper](https://www.nature.com/articles/nmeth.4636) can be loaded using `SeqfishDataset()`, `BreastCancerDataset()`, and `MouseOBDataset()`. 

@maxime1310 smFISH dataset can be loaded using `SmfishDataset()`. It loads gene expression data, labels, and x,y coordinates


imyiningliu:
The test is failing because of `xldr` (used to load SeqFISH data which is a excel file). Do we want to add it to requirement or remove the SeqFISH dataset loader?

jeff-regier:
It's fine to add a dependency on xlrd. Glad to see this is coming along.

PR #108, opened 2018-07-18
Show t_sne + some docstrings for the inference and model classes.

Edouard360:
Added docstrings for the inference and model classes and use show_t_sne  in notebook


Edouard360:
show_t_sne is now used in notebook but:

A lot of code was added in show_t_sne, to satisfy the notebooks' need, ie. color by batch, color by label or both at the same time.

https://github.com/YosefLab/scVI/blob/295f4b19d365da5579e21b70e48cd7551df0bc1b/scvi/inference/variational_inference.py#L100-L144

It can pbbly be factorized a bit but I wonder if it should remain in that file

PR #129, opened 2018-07-25
Add back the API section on documentation 

imyiningliu:


jeff-regier:
I think we don't want to add those files back -- building without them works perfectly for me locally. Just typing `make docs` gets everything right without them.

jeff-regier:
When you run `make docs`, it looks like it runs
```
sphinx-apidoc -o docs/ scvi
```
and then
```
python -msphinx docs docs/_build 

```

If you run the second without the first, the API section gets omitted from the documentation.

Issue #123, opened 2018-07-22
notebook of figures from the original scVI paper


jeff-regier:
Let's add a Jupyter notebook to the `doc/examples` directory that shows how to generate the figures (and one table) from the original scVI paper. It isn't necessary to reconstruct the parts of the figures and the table about competing methods that we compare to---we just need the parts about scVI.

- [x] Figure 1c. runtime; BRAIN_LARGE
- [x] Table 1. log likelihood; BRAIN_LARGE
- [x] Figure 2. imputation; CORTEX
- [ ] Figure 3. interpreting latent space; tsne plot and heatmap; CORTEX, HEMATO, random zinb
- [x] Figure 4. batch effect removal; tsne plot; RETINA
- [ ] Figure 5. differential expression; plots; PBMC
- [ ] Figure 6. plots; PBMC 
- [ ] Figure 7. posterior plots; CORTEX
- [ ] Figure 8. silhouette and batch mixing entropy; PBMC, CORTEX, CBMC, RETINA
- [x] Figure 9. imputation; CORTEX
- [x] Figure 10. imputation; CORTEX
- [x] Figure 11. imputation; CORTEX and BRAIN_LARGE
- [ ] Figure 12. DE; plots; PBMC
- [ ] Figure 13. DE; plots; BRAIN_SMALL
- [x] Figure 14. imputation; BRAIN_LARGE, CORTEX
- [ ] Figure 15. clustering; random and HEMATO
- [x] Figure 16. tsne; HEMATO

imyiningliu:
Do you remember how many epochs you used for the runtime figure (figure 1)? I only found the runtime code for other softwares on `reproducibility` notebook > <

jeff-regier:
^ @romain-lopez

romain-lopez:
Hi Yining, 

For this updated figure, let us use max_epochs = 125 but with early stopping (it should stop way before 125 epochs).

Romain

PR #132, opened 2018-07-27
Imputation

### LINKED ISSUE #80: imputation based on multiple samples
Our imputation results are based on just one sample from the variational distribution currently:

https://github.com/YosefLab/scVI/blob/704ce19d80e5728542464e488968c136a241eb6f/scvi/metrics/imputation.py#L24

That makes imputation not much of a metric for assessing changes to our model --- I'm surprised it works as well as it does. We should be sure to fix it before relying on imputation scores to guide any modeling decisions.

Edouard360:
resolves #80 and starts #123

From the new implementation, with Cortex [250 epochs], imputation median ~2.3.

imputation-1
 

- imputation:

  - corruption argument: binomial/bernouilli
  - training: before, only a forward pass on corrupted data was done, whereas the model needs to be trained on the corrupted data
  - n_samples: Multiple samples for get_sample_rate (which by the way before was not sampled, but only taking the mean qz_m, ql_m)
  - returning per cell original(but corrupted)/imputed value as output of the imputation
  - In output, print the global median or mean of median (across cells)

- other changes:
  - in modules, the forward pass for sequential layers now handles dim==3, ie. (n_samples, n_batch, n_features).
  - simplification in naming attributes:
    - data_loaders.data_loaders_loop -> data_loaders.loop
    - data_loaders.data_loaders_dict -> data_loaders.dict
    - data_loaders.data_loaders_kwargs -> data_loaders.kwargs
  - had to add a .sequential function to DataLoaderWrapper and a SequentialSubsetSampler, since we'd like to train on shuffled data_loaders, but when we compare corrupted/original values, we iterate jointly on the same indices


scvi/inference/variational_inference.py
```diff
@@ -67,14 +67,27 @@ def ll(self, name, verbose=False):
...
-    def imputation_errors(self, name, **kwargs):
-        return imputation(self.model, self.data_loaders[name], **kwargs)
+    def imputation(self, name, verbose=False, rate=0.1, n_samples=1, n_epochs=1, corruption="uniform",
+                   title="Imputation"):
+        original_list, imputed_list = imputation(self, name, rate=rate, n_epochs=n_epochs, n_samples=n_samples,
+                                                 corruption=corruption)
+        # Median of medians for all distances
+        median_imputation_score = np.median(np.abs(np.concatenate(original_list) - np.concatenate(imputed_list)))
+
+        # Mean of medians for each cell
```
Edouard360: Is this what you had in mind @jeff-regier ?

Edouard360:
@jeff-regier: you mentioned iirc to use the "median per cell" for the genes an then take the mean of that. Whereas the global median gives ~2.3, the mean of medians gives ~4.3. I still output it in `infer.imputation` but I can correct if you meant something else..

scvi/metrics/imputation.py
```diff
@@ -1,18 +1,118 @@
...
+        infer.data_loaders['corrupted_%s' % key] = DataLoaderWrapper(gene_dataset, use_cuda=infer.use_cuda, **kwargs)
+
+    infer.train(n_epochs=n_epochs)
+    infer.data_loaders.loop = original_data_loaders_loop
+
+    original_list = []
+    imputed_list = []
+    batch_size = infer.data_loaders.kwargs["batch_size"] // n_samples
+    for tensors, corrupted_tensors in \
+        zip(infer.data_loaders[name].sequential(batch_size=batch_size),
```
Edouard360: This is where I used the `.sequential` method and the SequentialSubsetSampler needed to iterate simultaneously on the same indices for both the uncorrupted and corrupted data.

scvi/metrics/imputation.py
```diff
@@ -1,18 +1,118 @@
...
+    original_data_loaders_loop = infer.data_loaders.loop
+    infer.data_loaders.loop = ['corrupted_%s' % s for s in infer.data_loaders.loop]
+    original_keys = list(infer.data_loaders.dict.keys())
+    for key in original_keys:
+        kwargs = copy.copy(infer.data_loaders.kwargs)
+        kwargs['collate_fn'] = gene_dataset.collate_fn
+        kwargs['sampler'] = copy.copy(infer.data_loaders[key].sampler)
+        infer.data_loaders['corrupted_%s' % key] = DataLoaderWrapper(gene_dataset, use_cuda=infer.use_cuda, **kwargs)
+
+    infer.train(n_epochs=n_epochs)
```
Edouard360: Before there was no training

scvi/metrics/imputation.py
```diff
@@ -1,18 +1,118 @@
...
 import numpy as np
 import torch
+from scipy.stats import kde
+
+from scvi.dataset import GeneExpressionDataset
+from scvi.dataset.data_loaders import DataLoaderWrapper
+
+
+def imputation(infer, name, rate=0.1, n_samples=1, n_epochs=1, corruption="uniform"):
+    corrupted_data = copy.deepcopy(infer.gene_dataset.X)
```
Edouard360: Not sure if that is memory efficient. (Before we did drop "on the fly", but there was no training of the model on the corrupted data, just forward pass on corrupted batches)

Edouard360:
Yes exactly, the median over all cells and all genes. I'm not sure more helps, there is some variance across runs it seems, like I have had around 2.1 ~ 2.5 sometimes. In 250 epochs, the model might not yet be fully trained (ll decreases again I think), so maybe hte results stabilize more at the end of training

Edouard360:
I will take a look at that + adapt notebooks according to the new slight changes in code

Edouard360:
10 runs on different corrupted instance on Cortex, where at the end of each experiment, we use different values of n_samples for the median imputation score.

| 1  | 8 | 32 | 128 |
| - | - |- |- |
| 2.20968175 | 2.20870447 |2.20864177 | 2.20848131|

These are the values averaged over the 10 runs. The differences are of the order 0.001.

Edouard360:
So we are actually better then ZINB-WAVE with enough epochs for training !

scvi/inference/variational_inference.py
jeff-regier: Yes, I think that is good since that's what in the paper, even though it's kind of an arbitrary metric.

scvi/metrics/imputation.py
jeff-regier: Won't that double the memory use of the program? Does it work to "corrupt" the training data on the fly, in the training loop?

scvi/metrics/imputation.py
jeff-regier: Are you trying to implement wake-sleep here? Or is this data augmentation?

jeff-regier:
If we're only using hidden one layer for Cortex, it's pretty hard to improve much over zinb-wave. (Both models are pretty inflexible then.). A larger dataset might be better for comparison.

scvi/metrics/imputation.py
Edouard360: Yes I agree, it is suboptimal in terms of space but not in term of clarity I think.

I don't see a very clean way to do it on the fly. We have to generate anyway 10% of new data. We could maybe somehow get the indices of the shuffled dataloaders in the training loop, but it would be computationally inefficient to overwrite the input for the corresponding indices at each loop iteration.

For a large dataset like brain large, romain only took a subsample of the dataset to run the imputation benchmark.

scvi/metrics/imputation.py
Edouard360: I want all the dataloaders to have their corrupted equivalent.
For VAE, one data_loader is enough for training, but we still have labelled and unlabelled dataloaders for VAEC, and SVAEC

PR #133, opened 2018-07-27
Pbmc

Edouard360:
Just adding the Purified PBMCs dataset


jeff-regier:
Looks good, but will rebase on master rather than merge master?

jeff-regier:
i.e.,

First, to drop the merge commit,
```
git reset HEAD~
git checkout -- .
```

Then, to rebase on master,
```
git fetch origin master
git rebase origin/master
```

Then to force push the changes,
```
git push -f origin pbcm
```

jeff-regier:
(It doesn't matter that much, but it's a good practice to rebase rather than merge for this case. See how this PR is listed as adding +4,551 new lines of code?)

Edouard360:
Thanks I wasn't sure if I had to force push

Edouard360:
Thank you it seems it has worked but actually the n lines of code, come from the other subsampled  purified datasets from 10X that I had to add: ie:
   "cd14_monocytes",
                       "b_cells",
                       "cd34",
                       "cd56_nk",
                       "cd4_t_helper",
                       "regulatory_t",
                       "naive_t",
                       "memory_t",
                       "cytotoxic_t",
                       "naive_cytotoxic"

Edouard360:
Or maybe we'd rather skip unit test for that one, in which case I can remove them all

jeff-regier:
I see. Can we just include ~ten lines in the test files?

Edouard360:
Ah maybe with symbolic links ?

jeff-regier:
Symbolic links are all platform specific. Better not to use them in a git repo.

jeff-regier:
Is it just test data you're putting in the test directory? If it's an actual dataset, not test dataset, you could add it to scvi-data: https://github.com/YosefLab/scVI-data/

Edouard360:
Okok, it just that all the samples in "test/data/10X/" have the same structure and instead of duplicating the folders for each dataset we could just create the references no ? Creating the link before the travis build wouldn't make it work ?

Edouard360:
It is just for testing - the dataset is a concat of multiple 10X datasets just like in pbmc.py (pbmc4k, pbmc8k), but here there are many more (around 10).

jeff-regier:
unit tests are supposed to be quick running --- better to just test it as well as possible with just a tiny file of testing data (e.g., < 100 entries).

PR #135, opened 2018-07-29
mmd and adversarial wrapper for any VariationalInference instance

Edouard360:
From two scmap datasets (segerstlope/muraro)

Without any of these wrapper, the batch entropy mixing gets up to 0.3.

Increasing the decoder number of layers to 3, it can go up to 0.45 (but slower to train then).

Still with only one layer, any of these penalites substantially increase the batch entropy mixing to around ~0.58 without imparing the likelihood nor the latent space as shown below.

mmd-adversarial

The usage of these wrapper is quite flexible and only augment the train/loss method of the infer object, such that all its attributes and methods are preserved.

Ex usage in tests:

https://github.com/YosefLab/scVI/blob/bbc8668221a51393ded6690edf2f6f84b8de765d/tests/test_scvi.py#L72-L80

In this example we show that we can even combine both of them as simply as this. The warm-up parameter is the number of epochs at which to start adding the penalty (since it makes no sense and would probably impair training to add the penalty while the latent space hasn't started to shape).

### DIFF
--- a/scvi/inference/experimental_inference.py
+++ b/scvi/inference/experimental_inference.py
@@ -48,14 +48,15 @@ def mmd_loss(self, tensors, *next_tensors):
         sample_batch, _, _, batch_index, label = tensors
         qm_z, _, _ = self.model.z_encoder(torch.log(1 + sample_batch), label)  # label only used in VAEC
         loss = mmd_objective(qm_z, batch_index, self.gene_dataset.n_batches)
+    else:
+        loss = 0
     return type(self).loss(self, tensors, *next_tensors) + loss
 
 
 def mmd_wrapper(infer, warm_up=100, scale=50):
     infer.warm_up = warm_up
     infer.scale = scale
-    infer.loss = types.MethodType(adversarial_loss, infer)
-    infer.train = types.MethodType(adversarial_train, infer)
+    infer.loss = types.MethodType(mmd_loss, infer)
     return infer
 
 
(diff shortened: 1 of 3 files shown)


scvi/inference/experimental_inference.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def adversarial_loss(self, tensors, *next_tensors):
+    if self.epoch > self.warm_up:
+        sample_batch, _, _, batch_index, label = tensors
+        qm_z, _, _ = self.model.z_encoder(torch.log(1 + sample_batch), label)  # label only used in VAEC
+        cls_loss = (self.scale * F.cross_entropy(self.GAN1(qm_z), batch_index.view(-1)))
+        self.optimizer_GAN.zero_grad()
+        cls_loss.backward(retain_graph=True)
+        self.optimizer_GAN.step()
```
jeff-regier: Is this a GAN? Or just an adversarial mechanism? Might be confusing for people reading the code to refer to GANs if it's the latter.

jeff-regier: (For a GAN, there's a generator network and a discriminator network that "compete". Not sure if we have that here.)

scvi/inference/experimental_inference.py
```diff
@@ -0,0 +1,89 @@
...
+        qm_z, _, _ = self.model.z_encoder(torch.log(1 + sample_batch), label)  # label only used in VAEC
+        loss = mmd_objective(qm_z, batch_index, self.gene_dataset.n_batches)
+    return type(self).loss(self, tensors, *next_tensors) + loss
+
+
+def mmd_wrapper(infer, warm_up=100, scale=50):
+    infer.warm_up = warm_up
+    infer.scale = scale
+    infer.loss = types.MethodType(adversarial_loss, infer)
+    infer.train = types.MethodType(adversarial_train, infer)
```
jeff-regier: This isn't a particularly idiomatic way to override the loss and train methods. It makes sense as far as getting something working, and it's pretty concise, but I worry it'll be hard to maintain.

scvi/inference/experimental_inference.py
```diff
@@ -0,0 +1,89 @@
...
+        sample_batch, _, _, batch_index, label = tensors
+        qm_z, _, _ = self.model.z_encoder(torch.log(1 + sample_batch), label)  # label only used in VAEC
+        loss = mmd_objective(qm_z, batch_index, self.gene_dataset.n_batches)
+    return type(self).loss(self, tensors, *next_tensors) + loss
+
+
+def mmd_wrapper(infer, warm_up=100, scale=50):
+    infer.warm_up = warm_up
+    infer.scale = scale
+    infer.loss = types.MethodType(adversarial_loss, infer)
```
jeff-regier: Did you mean to put `mmd_loss` here than than `adversarial_loss`?

scvi/inference/experimental_inference.py
jeff-regier: Would it work to make a `VariationalInferenceMMD` class that inherits from `VariationalInference` instead, which overrides `train` and `loss`?

jeff-regier:
Nicely done. These seem like very promising results!

scvi/inference/experimental_inference.py
Edouard360: you're right it is just an adversarial mechanism ! I will rename that

scvi/inference/experimental_inference.py
Edouard360: ah, yes this is a copy paste mistake

Edouard360: I am pretty sure my mmd result was wrong then.. will check it out

scvi/inference/experimental_inference.py
jeff-regier: I wonder if it was a mistake to make a hierarchy of `Inference` classes, rather than using doing something more like functional programming; i.e., calling functions with other functions are arguments. So, for example, a `train` function might be called with `loss` function as an argument. The functional programming approach might have made it easier to implement everything we want, without much boilerplate and without ever duplicating code.

Just something to think about.  Probably better not to go down that rabbit hole yet. Please merge this PR whenever you're ready.

Edouard360: I see, ok I will keep that in mind.

PR #138, opened 2018-07-31
Tighter likelihood

### LINKED ISSUE #73: better variational distribution for library size

maxime1310:
It gives slightly higher results than the reconstruction loss because the z drawn from q(z/x) has a higher likelihood under the measure q(z/x) than under p(z), which makes sense because of the samll discrepancy between the two distributions. The library size doesn't impact differently this computation of the LL. Closes #73

### DIFF
--- a/scvi/metrics/log_likelihood.py
+++ b/scvi/metrics/log_likelihood.py
@@ -2,6 +2,8 @@
 
 import torch
 import torch.nn.functional as F
+from torch.distributions import Normal
+import numpy as np
 
 
 def compute_log_likelihood(vae, data_loader):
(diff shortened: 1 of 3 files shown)


jeff-regier:
Is `compute_log_likelihood` the same as `compute_tighter_log_likelihood` with `n_samples_mc=1`? Could you rename `compute_tighter_log_likelihood` to  `compute_log_likelihood`, and delete what's currently called `compute_log_likelihood`?

jeff-regier:
I think it'd be better not to have two different functions for computing log likelihood, if at all possible, and the new code looks pretty general.

maxime1310:
@jeff-regier actually they're different, compute_log_likelihood only computes the reconstruction loss, so it's still different than when n_samples_mc =1. We could replace the compute_log_likelihood function so that per default it uses the new formula, but what I fear is that if we set n_samples_mc=1 it might have high variance and not be as a good indicator as just the reconstruction loss, and if we set n_samples_mc to be higher it will take too long to compute it every time we compute the general metrics for the model (every 5 epochs for instance). What do you think about it?

jeff-regier:
Ah, I got it. How about renaming `compute_tighter_log_likelihood` then to `compute_marginal_log_likelihood`? That should distinguishes it from `compute_log_likelihood`, which estimates `log p(x | z)` rather than `log p(x)`. Also maybe add some docstrings to clarify it?

maxime1310:
I'm on it

jeff-regier:
Would you call the new function too from the benchmarks, instead of calling the old `.ll` method here?

https://github.com/YosefLab/scVI/blob/d2c6c221fe64b8b2c1b2eeef22d2718afa43d51c/scvi/benchmark.py#L24

It's been `log p(x)` we wanted there all along, not `log p(x | z)`.

scvi/metrics/log_likelihood.py
```diff
@@ -18,6 +24,71 @@ def compute_log_likelihood(vae, data_loader):
...
+        dim = 0
+    s, _ = torch.max(inputs, dim=dim, keepdim=True)
+    outputs = s + (inputs - s).exp().sum(dim=dim, keepdim=True).log()
+    if not keepdim:
+        outputs = outputs.squeeze(dim)
+    return outputs
+
+
+def compute_tighter_log_likelihood_sample(vae, sample_batch, local_l_mean, local_l_var, batch_index,
+                                          labels, n_samples_mc=100):
```
jeff-regier: It'd be better to merge put this code into `compute_marginal_log_likelihood` rather than making it it's own function, since that's the only place it's called

PR #140, opened 2018-08-01
scRNAseq_and_smFISH

maxime1310:


scvi/dataset/data_loaders.py
```diff
@@ -190,3 +190,20 @@ def classifier_data_loaders(self):
...
+                 **data_loaders_kwargs):
+        """
+        :param train_size: float, int, or None (default is 0.1)
+        :param test_size: float, int, or None (default is None)
+        """
+        seq = TrainTestDataLoaders(gene_dataset_seq, train_size, test_size, seed, **data_loaders_kwargs)
+        fish = TrainTestDataLoaders(gene_dataset_fish, train_size, test_size, seed, **data_loaders_kwargs)
+        self.dict = {}
+        self.dict.update({'train_seq': seq['train'], 'test_seq': seq['test'],
+                         'train_fish': fish['train'], 'test_fish': fish['test']})
```
jeff-regier: Looks like we should be able to just use the normal `TrainTtsetDataLoaders` here.

scvi/dataset/dataset.py
```diff
@@ -53,10 +57,18 @@ def download_and_preprocess(self):
     def collate_fn(self, batch):
         indexes = np.array(batch)
         X = torch.FloatTensor(self.X[indexes]) if self.dense else torch.FloatTensor(self.X[indexes].toarray())
-        return X, torch.FloatTensor(self.local_means[indexes]), \
-            torch.FloatTensor(self.local_vars[indexes]), \
-            torch.LongTensor(self.batch_indices[indexes]), \
-            torch.LongTensor(self.labels[indexes])
+        if not self.got_spatial:
```
jeff-regier: Let's just have `if x_coord is not None:` here, and get rid of the `got_spatial` variable.

scvi/dataset/dataset.py
```diff
@@ -32,6 +32,10 @@ def __init__(self, X, local_means, local_vars, batch_indices, labels,
         self.batch_indices, self.n_batches = arrange_categories(batch_indices)
         self.labels, self.n_labels = arrange_categories(labels)
         self.x_coord, self.y_coord = x_coord, y_coord
+        if x_coord is not None:
+            self.got_spatial = True
+        else:
+            self.got_spatial = False
```
jeff-regier: Let's get rid of the `got_spatial` variable, and just use `x_coord not is None` if we need it. Better to have fewer variables.

scvi/dataset/dataset.py
```diff
@@ -53,10 +57,18 @@ def download_and_preprocess(self):
...
         X = torch.FloatTensor(self.X[indexes]) if self.dense else torch.FloatTensor(self.X[indexes].toarray())
-        return X, torch.FloatTensor(self.local_means[indexes]), \
-            torch.FloatTensor(self.local_vars[indexes]), \
-            torch.LongTensor(self.batch_indices[indexes]), \
-            torch.LongTensor(self.labels[indexes])
+        if not self.got_spatial:
+            return X, torch.FloatTensor(self.local_means[indexes]), \
+                   torch.FloatTensor(self.local_vars[indexes]), \
+                   torch.LongTensor(self.batch_indices[indexes]), \
+                   torch.LongTensor(self.labels[indexes])
```
jeff-regier: Better to return dummy values for x_coord and y_coord, like we do for batch_indices when there aren't batches, rather than branching

scvi/dataset/smfish.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, save_path='data/'):
             x_coord=x_coord, y_coord=y_coord)
 
     def preprocess(self):
-        print("Preprocessing smFISH dataset")
+        print("Preprocessing smFISH_utils dataset")
```
jeff-regier: What utils?

scvi/inference/variational_inference.py
```diff
@@ -306,3 +310,103 @@ def loss(self, tensors_all, tensors_labelled):
...
+        self.ponderation = 0
+        self.kl_weight = 0
+        self.classification_ponderation = 0
+        self.data_loaders = TrainTestDataLoadersFish(gene_dataset_seq, gene_dataset_fish,
+                                                     train_size=train_size, use_cuda=use_cuda)
+
+    def loss(self, tensors_seq, tensors_fish):
+        sample_batch, local_l_mean, local_l_var, batch_index, labels = tensors_seq
+        reconst_loss, kl_divergence = self.model(sample_batch, local_l_mean, local_l_var, batch_index, mode="scRNA",
+                                                 ponderation=self.ponderation)
```
jeff-regier: ponderation -> weighting ?

scvi/inference/variational_inference.py
```diff
@@ -306,3 +310,103 @@ def loss(self, tensors_all, tensors_labelled):
...
+        loss = torch.mean(reconst_loss + self.kl_weight * kl_divergence)
+        sample_batch_fish, local_l_mean, local_l_var, batch_index_fish, _, _, _ = tensors_fish
+        reconst_loss_fish, kl_divergence_fish = self.model(sample_batch_fish, local_l_mean, local_l_var,
+                                                           batch_index_fish, mode="smFISH_utils")
+        loss_fish = torch.mean(reconst_loss_fish + self.kl_weight * kl_divergence_fish)
+        loss = loss * sample_batch.size(0) + loss_fish * sample_batch_fish.size(0)
+        loss /= (sample_batch.size(0) + sample_batch_fish.size(0))
+        return loss + loss_fish
+
+    def train(self, n_epochs=20, lr=1e-3, params=None):
```
jeff-regier: Let's find a way to use the `train` method from `VariationalInference` rather than overriding it.

scvi/metrics/clustering.py
```diff
@@ -74,3 +75,52 @@ def entropy(hist_data):
             batches[kmatrix[indices].nonzero()[1][kmatrix[indices].nonzero()[0] == i]]
         ) for i in range(100)])
     return score / 50
+
+
+def get_data(vae, data_loader, mode):
```
jeff-regier: Every variable is data. Let's give this function a name that will help someone who is reading the code for the first time to understand what it does.

scvi/metrics/log_likelihood.py
```diff
@@ -4,13 +4,18 @@
 import torch.nn.functional as F
 
 
-def compute_log_likelihood(vae, data_loader):
+def compute_log_likelihood(vae, data_loader, name):
```
jeff-regier: Better not to take `name` as an argument. Instead, let's make `compute_log_likelihood`  a method in the `VariationalInference` class. Then override `compute_log_likelihood` in the `VariationalInferenceFish` class.

smFISH_utils/metrics/visualisation.py
```diff
@@ -0,0 +1,228 @@
+import matplotlib.pyplot as plt
+import matplotlib.gridspec as gridspec
+
+
+def show_cell_types(latent_seq, labels_seq, latent_fish, labels_fish, title='latent.svg', x_lim=[-85, 85],
+                    y_lim=[-85, 85]):
```
jeff-regier: Would it work to make all these smfish-specific function methods in the `VariationalInferenceFish` class?

jeff-regier: That would keep all the smFISH-specific code together in one place.

maxime1310:
@jeff-regier we just went through all the new code with Ed, he has a lot of ideas on how to factorize it (I don't know how I would survive without him haha), we'll probably be able to get to a much cleaner version.

scvi/dataset/dataset.py
jeff-regier: @Edouard360  Could we move `collate_fn` to the `Inference` classes? (and override it in `VariationalInferenceFish` with this one?)

For that matter, we move all the `DataLoaders` logic (e.g. `TrainTestDataLoader`) to the `Inference` class? It seems like almost every Inference class e.g. (`VariationalInference`) needs a different data loader any way -- it'd be nice not to have two separate hierarchies of classes.

smFISH/metrics/visualisation.py
```diff
@@ -0,0 +1,228 @@
+import matplotlib.pyplot as plt
```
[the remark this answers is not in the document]
jeff-regier: I'm not sure I understand -- maybe just make whatever changes you think are good and let me look over the code before committing it?

I'm interested in getting rid of everything in the `metrics` directory eventually, and moving it all to `Inference` or `VariationalInference`. Currently, it seems like every metric is defined two places --- in an `Inference` class (or subclass of it) and then again in the `metrics` directory. Seems unnecessarily complicated.

But I could easily be missing something about the importance of the `metrics` directory -- glad to see what you can do to simplify it.

scvi/dataset/dataset.py
Edouard360: `TrainTestDataLoader` is also used in `ClassifierInference`, which serves to: 1) the "alternate" training scheme `SemiSupervisedModel` in which at the end of each epoch, we train only the classifier weights on top of the latent space 2) if we wanted to do a NN classifier on top of a VAE or on raw data.

I see what you mean. It is possible to move the functionalities of `DataLoaders`  into `Inference`, only keeping the `DataLoadersWrapper` class.

`collate_fn` is a  method of pytorch's `Dataset` class used by pytorch's dataloaders however... I am not sure to understand how to make it part of Inference.

smFISH/metrics/visualisation.py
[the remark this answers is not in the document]
Edouard360: Shouldn't all this metrics be separated in different files ? If so would you prefer to move them all in Inference directory ? Or if not, this would mean the `variational_inference.py` file would get pretty big, with all the methods following one another. 

It is true that very few functions would be used elsewhere - there are the clustering scores that might serve for other algorithms.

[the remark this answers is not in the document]
jeff-regier: I'm not sure having long files is a problem.

scvi/metrics/classification.py
```diff
@@ -84,29 +86,35 @@ def compute_accuracy_svc(data_train, labels_train, data_test, labels_test, unit_
...
     y_pred_train = clf.predict(data_train)
 
-    return (compute_accuracy_tuple(y_pred_train, labels_train),
-            compute_accuracy_tuple(y_pred_test, labels_test))
+    if not return_labels:
+        return (compute_accuracy_tuple(labels_train, y_pred_train),
+                compute_accuracy_tuple(labels_test, y_pred_test))
+    elif return_labels:
+        return (compute_accuracy_tuple(labels_train, y_pred_train),
+                compute_accuracy_tuple(labels_test, y_pred_test), y_pred_test)
```
jeff-regier: Would you just always return the labels, rather than having this if statement? If you call the function and you don't need them, you can just use and underscore as a variable to assign to; e.g.,
```
train_acc, test_acc, _ = infer.compute_accuracy_classifier(...)
```

scvi/metrics/log_likelihood.py
```diff
@@ -63,4 +49,5 @@ def log_nb_positive(x, mu, theta, eps=1e-8):
     res = theta * torch.log(theta + eps) - theta * torch.log(theta + mu + eps) + x * torch.log(
         mu + eps) - x * torch.log(theta + mu + eps) + torch.lgamma(x + theta) - torch.lgamma(
         theta) - torch.lgamma(x + 1)
+    # Gives the ability to do a warmup on the unobserved smFISH genes when merging the two technologies
```
jeff-regier: How so? Is this comment outdated?

scvi/models/vae_fish.py
```diff
@@ -0,0 +1,202 @@
...
+                 reconstruction_loss_fish="poisson", model_library=False):
+        super(VAEF, self).__init__(n_input, dispersion=dispersion, n_latent=n_latent, log_variational=log_variational,
+                                   reconstruction_loss=reconstruction_loss, n_batch=n_batch, n_labels=n_labels)
+        self.n_input = n_input
+        self.n_input_fish = len(indexes_fish_train)
+        self.indexes_to_keep = indexes_fish_train
+        self.reconstruction_loss_fish = reconstruction_loss_fish
+        self.model_library = model_library
+        # First layer of the encoder isn't shared
+        self.z_encoder = Encoder(n_input, n_hidden, n_hidden=n_hidden, n_layers=1,
```
jeff-regier: Is there a `z_encoder` already in the parent class? If so, you wouldn't want to instantiate a new one here, I suspect.

scvi/dataset/dataset.py
```diff
@@ -255,6 +265,24 @@ def _filter_genes(gene_dataset, gene_names_ref, on='gene_names'):
         subset_genes = np.array([gene_names.index(gene_name) for gene_name in gene_names_ref], dtype=np.int64)
         return gene_dataset.X[:, subset_genes], subset_genes
 
+    @staticmethod
+    def reorganize(x, genes, ordered_genes):
```
jeff-regier: It looks like this is only used in `CortexDataset`. Please move it there to keep it near where it's used. 

Also, please give a name that's a bit more descriptive than `reorganize`. Maybe `reorder_genes` or something?

scvi/inference/experimental_inference.py
```diff
@@ -73,6 +73,22 @@ def adversarial_loss(self, tensors, *next_tensors):
     return type(self).loss(self, tensors, *next_tensors) - cls_loss
 
 
+def adversarial_loss_fish(self, tensors_seq, tensors_fish):
```
jeff-regier: Is it possible to put these smFISH-specific methods all the `VariationalInferenceFish` class? It'd help a lot to keep them all together.

scvi/models/vae_fish.py
```diff
@@ -0,0 +1,221 @@
...
+        if self.log_variational:
+            x_ = torch.log(1 + x_)
+        # Sampling
+        if mode == "scRNA":
+            qz_m, qz_v, z = self.z_encoder(x_)
+            library = torch.log(torch.sum(x, dim=1)).view(-1, 1)
+            batch_index = torch.zeros_like(library)
+        if mode == "smFISH":
+            qz_m, qz_v, z = self.z_encoder_fish(x_[:, self.indexes_to_keep])
+            library = torch.log(torch.sum(x[:, self.indexes_to_keep], dim=1)).view(-1, 1)
```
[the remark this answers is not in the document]
jeff-regier: Less branching sounds good

scvi/metrics/classification.py
maxime1310: I'll do that and modify accordingly the part of the code where we call that function.

scvi/metrics/log_likelihood.py
maxime1310: Yes it's outdated! I'm currently moving the previous logic of weighting some genes more than other within the vae_fish model

scvi/dataset/dataset.py
maxime1310: I'll move that and rename it. I really need to start naming my methods in a more informative way, I always come up with lousy names 😞

scvi/models/vae_fish.py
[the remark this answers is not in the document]
maxime1310: That seems feasible and should spare some lines of code, I'll try to do that

scvi/models/vae_fish.py
maxime1310: There is one but it doesn't take the same parameters as input, so if we use the one from the parent class we'd have to specify different parameters in the init (and then we'd lose the interest of inheriting from VAE as we'd have to respecify most features for the initialization)

scvi/inference/experimental_inference.py
maxime1310: I'm not sure how to do this because this is specific to the adversarial inference class, I'll see if I can do something about it.

maxime1310:
@jeff-regier just realized I forgot to bring coverage to its original level. I'm on it.

scvi/models/vae_fish.py
jeff-regier: You mean in the call to `super(VAEF, self).__init__(...)` you'd have to specify some different parameters? That's OK --it's an improvement over how it is now. Currently we're getting instantiating `z_encoder` twice, but the parent instance is masked so it's never seen. That's inefficient and confusing for someone new to the code. e.g., if some moves the call to `super(VAEF, self).__init__(...)`  to the end of the `__init__` function, it would totally change the behavior.

jeff-regier:
@maxime1310 To get rid of these merge conflicts, you could do
```
git fetch origin master
git rebase origin/master
```

Also, how about 
```
mv scRNA_and_smFISH_new_version.ipynb docs/examples/scRNA_and_smFISH.ipynb
```
?
You've got two smFISH notebooks in the pull request currently.

maxime1310:
I had trouble using rebase (I don't know why it took me back to an old version of my branch, several commits ago, and started the conflict resolving from this point), so I just incorporated the two commits on master that have been done since I branched (namely the addition of compute_marginal_log_likelihood method and you're changes for the dense/sparse issue). I also deleted the second notebook, that was a mistake from my part.

jeff-regier:
That's how rebase works: it applies your old commits one at at time and each time you fix the conflicts before it proceeds to the next commit.

Maybe `git merge origin/master` would work better in this case because this pull request is so big.

maxime1310:
Yes I think rebase was complicated because there was so many commits, `git merge origin/master` was much faster. All the conflicts must be resolved by now.

jeff-regier:
OK. Please make sure the new notebook runs all the way through without errors. You might try running it from the command line like I do below:

https://github.com/YosefLab/scVI/blob/296fdb654c765648747a423ccb614a1531d38dc5/Makefile#L93-L94

jeff-regier:
If you put `cd ../..` in a cell by itself at the top of the notebook, it'll get the path right to import scvi.

jeff-regier:
Please go ahead and click "Squash and merge" once the command below runs wo error:
```
jupyter nbconvert --to notebook --execute docs/examples/scRNA_and_smFISH.ipynb --ExecutePreprocessor.timeout=1800 
```

(I still get errors even with the latest change.)

maxime1310:
Yes I get the same error for all the notebooks , it must be the path that is not well specified

jeff-regier:
The other notebooks should work fine now, after #146

jeff-regier:
Maybe merge once more?

maxime1310:
Oh I hadn't seen that change! I was desperatly trying to make the notebook run from the command line  with `os.chdir('../.') ` instead of `cd ../..`. Seems to be working now, I'll just wait for the whole notebook to run

maxime1310:
Testing the notebooks is pretty cool I think!

maxime1310:
The notebook passes the test now! 😄

maxime1310:
Thanks for putting up with this Jeff! Glad that it's finally done!

PR #160, opened 2018-08-13
[WIP] Adding some opinionated docstrings

jamestwebber:
I've been writing docstrings as a way to wrap my head around the codebase, but as I did I realized I have some fairly strong opinions about how to write them.

 * I took advantage of the fact that scVI is a Python 3.6+ project and added [type hints](https://www.python.org/dev/peps/pep-0484/) to the methods I was documenting. This may be a controversial proposal but I think it will pay dividends for developers as well as users.
 * I added [autodoc-typehints](https://github.com/agronholm/sphinx-autodoc-typehints) to the list of sphinx extensions to parse said hints when generating the documentation.
 * I also added [mathjax](http://www.sphinx-doc.org/en/master/usage/extensions/math.html#module-sphinx.ext.mathjax) configuration to allow the documentation to include LaTeX. This is currently only used in `models.modules.Encoder` but I can imagine it coming in handy in many cases given the nature of the project.
   * For some reason I could not get the `$` inline format to work, but using `\(` and `\)` worked for me.

If those additions to the doc-building process are acceptable to you all, I will continue with that style. Otherwise I can retain the docstrings but remove those dependencies. I do think the type hints are worth keeping, though, if you have no plans to support earlier versions of Python.

Also, this PR is a work in progress and many of my docs are based on a potentially-faulty understanding of the code (that was the motivation for writing them). So I'm happy to get some review from those who know better what is going on.


scvi/models/vae.py
```diff
@@ -36,17 +44,19 @@ class VAE(nn.Module):
...
         self.dispersion = dispersion
         self.n_latent = n_latent
         self.log_variational = log_variational
         self.reconstruction_loss = reconstruction_loss
-        # Automatically desactivate if useless
+        # Automatically deactivate if useless
         self.n_batch = n_batch
         self.n_labels = n_labels
-        self.n_latent_layers = 1
+        self.n_latent_layers = 1 # not sure what this is for, no usages?
```
jamestwebber: This variable (and the equivalent one in `SVAEC`) are never used, and could perhaps be removed?

scvi/models/modules.py
```diff
@@ -94,15 +184,43 @@ def forward(self, dispersion, z, library, *cat_list):
...
 
         self.mean_decoder = nn.Linear(n_hidden, n_output)
         self.var_decoder = nn.Linear(n_hidden, n_output)
 
-    def forward(self, x, *cat_list):
+    def forward(self, x: torch.Tensor, *cat_list: int):
+        r"""The forward computation for a single sample.
+
+         #. Decodes the data from the latent space using the decoder network
+         #. Returns mean and variance for a multivariate Gaussian
```
jeff-regier: I think it should be parameters for a ZINB (zero inflated negative binomial) output distribution

jeff-regier: Well, it can be used several ways -- maybe better just to say it outputs parameters for a distribution.

jeff-regier:
This looks good -- thanks for the contribution, James!

@Edouard360 Will you look over it to and merge it once the comments all look right?

docs/conf.py
```diff
@@ -32,7 +32,8 @@
 
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
-extensions = ['sphinx.ext.autodoc', 'sphinx.ext.viewcode', 'nbsphinx']
+extensions = ['sphinx.ext.autodoc', 'sphinx.ext.viewcode', 'nbsphinx',
+              'sphinx_autodoc_typehints', 'sphinx.ext.mathjax']
```
jeff-regier: Nice, that'll be handy to have mathjax loaded.

scvi/inference/variational_inference.py
```diff
@@ -34,7 +34,7 @@ class VariationalInference(Inference):
         :gene_dataset: A gene_dataset instance like ``CortexDataset()``
         :train_size: The train size, either a float between 0 and 1 or and integer for the number of training samples
          to use Default: ``0.8``.
-        :\**kwargs: Other keywords arguments from the general Inference class.
+        :\*\*kwargs: Other keywords arguments from the general Inference class.
```
jeff-regier: Ah, glad to have that fixed!

scvi/models/modules.py
jamestwebber: Yeah this was one place I was looking for clarification. This model is used in [SVAEC](https://github.com/YosefLab/scVI/blob/master/scvi/models/svaec.py#L56) as `decoder_z1_z2` which I interpreted as translating from one latent space to another (I think), and the results are passed into a normal distribution but as you say it could have other uses.

scvi/models/vae.py
```diff
@@ -36,17 +44,19 @@ class VAE(nn.Module):
...
         self.dispersion = dispersion
         self.n_latent = n_latent
         self.log_variational = log_variational
         self.reconstruction_loss = reconstruction_loss
-        # Automatically desactivate if useless
+        # Automatically deactivate if useless
         self.n_batch = n_batch
         self.n_labels = n_labels
-        self.n_latent_layers = 1
+        self.n_latent_layers = 1  # not sure what this is for, no usages?
```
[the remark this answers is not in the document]
jamestwebber: Should it be set to the value of `n_layers` then? As I commented, it doesn't appear to be used in the code anywhere, and it's always set to 1 (or 2 in `SVAEC`), so I wasn't sure what it was for.

jamestwebber:
I was planning to keep going in this branch, but I guess it's easy to create a new PR if you want to merge this one. I'll just rebase to master (or do some other git thing)

scvi/models/vae.py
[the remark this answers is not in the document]
jeff-regier: Uh, it may be different than `n_layers` by 1. I'm not sure. @Edouard360 do you know what it's called `n_latent_layers` and not `n_layers`?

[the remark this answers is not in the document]
Edouard360: `n_latent_layers = 1`  for `VAE` and `n_latent_layers = 2` for `SVAEC`.
It's the number of levels for the latent stochastic variables.
At one point we had ladder autoencoder models as well, which had multiple levels of stochastic variables. We no longer have the corresponding visualization tools on `master` so perhaps it could be removed altogether..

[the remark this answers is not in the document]
jeff-regier: Ok, good to get rid of it then

PR #163, opened 2018-08-16
Delete metrics

### LINKED ISSUE #153: remove unit_test keyword arguments
We should remove the `unit_test` flags from any functions, e.g.,

https://github.com/YosefLab/scVI/blob/c08d115435499749d599f66793d4278be232ca26/tests/test_scvi.py#L50

We don't want the code to be run a special way by the unit tests. If we have to run code differently for the unit tests, just add new keyword argument to the functions, e.g., `num_samples=10`.

### LINKED ISSUE #155: refactoring: metric -> inference
We have a bunch of methods in our inference classes that do nothing but call a function in the `metric` directory. e.g.,

https://github.com/YosefLab/scVI/blob/e1c210135351e434aa619e72cdcf3dab04d35c3d/scvi/inference/variational_inference.py#L71-L75

It'd be clearer to just have the code the implements the method in the method -- better for reading the code not to have to jump to a different file in a different directory.

I'd be glad to get rid of the `metric` directory entirely, if possible.

### LINKED ISSUE #156: refactoring: DataLoader returns a DataLoaderWrapper?
That seems backwards: the wrapper usually knows about class it wraps, not the other way around. Is there a way to just combine DataLoader and DataLoaderWrapper into one class? I don't think we ever want to use a dataloader without calling the ability to call the `tocuda` method.

https://github.com/YosefLab/scVI/blob/c08d115435499749d599f66793d4278be232ca26/scvi/dataset/data_loaders.py#L77-L88

### LINKED ISSUE #157: refactoring: SemiSupervisedVariationalInference -> classifier_inference.py
It might be nice to move `SemiSupervisedVariationalInference` and all it's subclasses to `classifier_inference.py`, so that the annotation inference code is all together. Then we could also put the code from `metrics/classification.py` into `inference/classifier_inference.py` too, above and outside of all the classes in that file for all to use.

### LINKED ISSUE #162: imputation without copying the data
For imputation, I think we want to corrupt the data on the fly, rather than making a copy of the data matrix:

https://github.com/YosefLab/scVI/blob/9e1eacd7fad861bd47968a1850d5ab5bf22a6ed7/scvi/metrics/imputation.py#L14

If for some reason we need to do multiple passes over the same corrupted data, then we can store and reset a seed so that every time the noise gets introduced in the same way.  But I'm not sure that's necessary.

### LINKED ISSUE #164: impute method + notebook example
From an email:

```
I've looked at tutorial in: scVI/docs/notebooks/basic_tutorial.ipynb

I'm trying to figure out imputed data matrix from this tutorial. For example, the Cortex Dataset is 3005
 X 558, I'm expecting the imputed data matrix should be
3005 X 558.

How can I exact the imputed data matrix from this tutorial? or How can I exact the imputed data matrix from scVI?
```

I think in the notebook currently we show how to score the imputation results, but we don't show how to get access to the imputed values.

Edouard360:
Closes #153
Closes #155
Closes #156
Closes #157
Closes #162
Closes #164

### Improvements:

- removing `metrics` folder.

- A `Posterior` object containing a data_loader makes it more convenient for the user to interact with the data unit rather than with a keyword. Separating the model's training logic for the data_loaders posterior logic seems also like a good idea. Adapting all `tests` accordingly:

https://github.com/YosefLab/scVI/blob/0b24af17291b0fcf593c7259cd40feafd59c7f44/tests/test_scvi.py#L22-L28

### Some issues :

- Should there be inheritance in the `Posterior` classes ? `PosteriorSemiSupervised`, `PosteriorFish` have methods that the general `Posterior` class might not have, like `.accuracy`, `.hierarchical_accuracy`, ect for `PosteriorSemiSupervised` or `show_spatial_expression` `ll_fish` for `PosteriorFish`. Separating the model's training logic from the posterior's methods logic through the `Posterior` object seem to introduce as well this class hierarchy, though I know we want to avoid this hierarchy appearing both in `Trainer` and `Posterior`... One way out might be to put all possible posterior's methods under the same general `Posterior` class, regardless of their specificity.

- The imputation task can easily be included just as another `Posterior`'s method. However for the imputation **benchmark**, there should be an `ImputationTrainer`. If we want to make this benchmark general and be able to work for all models (`VAE`, `VAEC`, `SVAEC`), I think:
   - `ImputationTrainer` needs to be instanciated with another trainer. 
   - All the data_loaders used for the training the model are replaced by their corrupted version (eventually corrupted on the fly, but for now still copied). 
   - The model is trained its corresponding trainer but the updated data_loaders (or `Posterior` instances). 
   - `ImputationTrainer` has an `.imputation` method returning the benchmark result, because it cannot be held by a `Posterior` object - indeed this result is obtained through iteration over two data_loaders, and a `Posterior` has only one. 

This imputation benchmark is probably not needed for SVAEC, and VAEC, but to generalize and corrupt systematically all the data_loaders/ posterior's instance involved in the training of these models, I sticked with `data_loaders_dict` / `data_loaders_loop` rename accordingly to `posteriors_dict` / `posteriors_loop`. This way I can translate all the posteriors into their corrupted version like this:

https://github.com/YosefLab/scVI/blob/0b24af17291b0fcf593c7259cd40feafd59c7f44/scvi/inference/trainers.py#L108-L118   

- `SemiSupervisedTrainer` still has a `ClassifierTrainer` instance.

- Some of the  `Posterior` method some methods like `.nn_overlap_score` only work with "sequential" data since we need the protein expression levels (an attribute of the dataset instance but not iterated over in the data_loaders). Two options here; (For now I chose 1))
   - add a `.sequential` method on the `ToCudaDataLoader` wrapper so as to iterate on the corresponding data_loader but sequentially.
   - recover the exact indices that the data_loader used to iterate over the dataset. 

- No notebook translated yet. Would you have time to approve or comment these changes @jeff-regier ? 

### Other Remarks

- A `Posterior` instance holds a data_loader but also the model, and the `gene_dataset` as attributes.


jeff-regier:
> One way out might be to put all possible posterior's methods under the same general Posterior class, regardless of their specificity.

This sounds good to me as long as all the methods run without errors. (e.g., as long as some `Posterior` objects aren't missing fields that are needed for any method to work). It's good to put off creating an extra hierarchy of classes.

jeff-regier:
Oh, though `show_spatial_expression` really does seem very specific to osmFISH. Maybe it's time for a `SpatialPosterior` class.

jeff-regier:
> The imputation task can easily be included just as another Posterior's method. 

Yes, perfect.

> However for the imputation benchmark, there should be an ImputationTrainer. 

Yes, that seems right to me.

> If we want to make this benchmark general and be able to work for all models (VAE, VAEC, SVAEC)

Seems like it might as well work for all the models

> `ImputationTrainer` needs to be instanciated with another trainer.

You mean take another trainer instance as an argument? That doesn't sound so good.

I had in mind that `ImputationTrainer` would be a subclass of `Trainer`. It would use the data loaders from `Trainer` and the `train` method. But if would override a method (a hook) that lets it corrupt each minibatch of training data (on the fly).

It's fine to defer `ImputationTrainer` until a later pull request if it seems like a lot work. (Did we ever get improved results from that kind of training? Seems like more work was needed, right?) Might be better to handle that as a separate project.

Edouard360:
I have never `ImputationTrainer` as another training scheme - it could perhaps give better regularization to train the model on slightly corrupted data but I never looked into that. It was really just for the original benchmark of the scVI paper. 

If we only wanted to do this benchmark for the VAE's model, it would be fine to have `ImputationTrainer` a subclass of `UnsupervisedTrainer`, which would indeed take the `train_set`, `test_set` posteriors, corrupt them, and there would be no need to rewrite the `.train` function.

But the semi-supervised models work with other posteriors, by their name and by their number:  `train_set`, `test_set` becomes `labelled_set`, `unlabelled_set`, `full_dataset`. So the `ImputationTrainer` as described just above wouldn't work.

I agree that's probably not the main focus of that pull request...

jeff-regier:
Is `nn_overlap_score` a core function of scVI, or is it just for computing a benchmark score for a paper? Maybe it could be in a notebook instead of part of the main scVI library?

Are there other functions that need to access the data sequentially?

Edouard360:
Ok I will add docstring. I think `.nn_overlap` is the only case where we want to do that, though using sequential might be useful if you want to combine output from different posterior's methods like `.get_latent`, `.imputation`;  or `.compute_predictions` if you want to have the labels. Actually the ideal posterior instance would iterate only once over the dataloader to give all the outputs we are interested in, and this would remove the need for a sequential data_loader altogether. But that is a big change...

jeff-regier:
You could always create a new `torch.uitls.DataLoader`  within `nn_overlap` every time it is a called, with a sequential sampler, as a sort of work around. It's is that method is a bit of hack, if it keeps us from having to change the overall code structure.

Edouard360:
Yes ! That's what I had come up with !

https://github.com/YosefLab/scVI/blob/0b24af17291b0fcf593c7259cd40feafd59c7f44/scvi/inference/trainer.py#L200-L228

Also the user might want to sequentially extract data from `.get_latent` or `.imputation`, for instance if they have any meta-information they would like to analyse with scVI's output.

Ok so, for now I am collapsing the Posterior's hierarchy into the unique posterior class that doesn't do anything if there is missing data, ie.

`train_set.accuracy()` would do nothing (and throw no error), but

`labelled_set.accuracy()` would work properly.

`train_fish.ll()` would do nothing (and throw no error), but

`train_fish.ll_fish()` would properly

ect...

Also two minor question with file organization:

-  Do we keep inference as folder name ? We no longer have Inference class or inference.py file.

- For now I have
   - `trainer.py`:
      - Trainer (the general trainer class)
      - `EarlyStopping`
   - `trainers.py`
      - UnsupervisedTrainer
      - AdapterTrainer (that's for when we only train the encoder on the test set)
      - ImputationTrainer
      - TrainerFish
   - `classifier_trainer.py`.
     - ClassifierTrainer
     - SemiSupervisedTrainer
 
But I think splitting all trainers into different files is best... Do you prefer to keep it together ?

jeff-regier:
I'd probably keep all the trainers together in one file, but either way is fine. Better not to have a two files whose names only difference by a 's'. I.e. trainers.py / trainer.py.

"inference" still seems to apply to the directory... Doesn't need a be a file named inference.py in it.

Edouard360:
I rebased on master and completed:

### Imputation Method

This is now included in the basic tutorial notebook as discussed in #164.

`trainer.train_set.imputation()`

### Dealing with "Sequential" data_loader

Any posterior's sequential version can be access through the `.sequential()` method that returns the posterior but with an updated version of its data_loader.

```
indices = trainer.train_set.indices 
imputed_values_indices = trainer.train_set.sequential().imputation()
```

For the nn_overlap method, it ends up being quite elegant, I think

https://github.com/YosefLab/scVI/blob/eee3cc15b895cf1286323105d9090e7440335fbe/scvi/inference/posterior.py#L269-L277

### Imputation Benchmark

I adressed the on-the-fly corruption mentionned in #162.

We first corrupt the dataset, and keep in `self.corrupted_dict` the info for the data corruption (a dictionary with as key the line indices and values the columns indices + corresponding corrupted values).

https://github.com/YosefLab/scVI/blob/eee3cc15b895cf1286323105d9090e7440335fbe/scvi/dataset/dataset.py#L80-L81

The `collate_fn_corrupted` then corrupts data on the fly.

https://github.com/YosefLab/scVI/blob/eee3cc15b895cf1286323105d9090e7440335fbe/scvi/dataset/dataset.py#L68-L78

If the posterior has a gene_dataset that has been corrupted, we can either iterate over its regular dataloader, or its corrupted version: `trainer.train_set.corrupted()`, (which returns a shallow copy of the posterior instance attributes but an updated data_loader, with the `collate_fn_corrupted` instead of `collate_fn`).

Also, the `Trainer` parent class has the `.corrupt_posteriors` and `.uncorrupt_posteriors` methods to update them accordingly:

https://github.com/YosefLab/scVI/blob/eee3cc15b895cf1286323105d9090e7440335fbe/scvi/inference/trainer.py#L164-L171

Running the imputation benchmark then looks like:

https://github.com/YosefLab/scVI/blob/eee3cc15b895cf1286323105d9090e7440335fbe/tests/test_scvi.py#L33-L37

Which gets rid of the `ImputationTrainer` class, and better adapts to other `Trainer` instances.

Edouard360:
Also `ToCudaDataLoader` has been merged in Posterior which makes it clearer I think

jeff-regier:
Outstanding! Thanks Eddie.

PR #194, opened 2018-09-19
Better solved the dropout problem on the latent space to still apply …

romain-lopez:
…dropout to further layers


romain-lopez:
Closed until we figure out the source of the numerical unstability

PR #195, opened 2018-09-20
WIP: Output/save posterior summary statistics

### LINKED ISSUE #188: Extracting data from the model
We have some requests on how to extract the following information as np.array type from the model such as:
+ A latent space for a given set of cells
+ Imputed values on a given set of cells

It is important to update our introductory notebook to reflect those new features. The latent space is what our users would like to see at first sight.

jules-samaran:
I made this changes so that a way to retrieve and save interesting information like the latent space and the imputed value for the whole dataset might be more clear and apparent in the basic tutorial notebook.

closes #188


scvi/inference/trainer.py
```diff
@@ -65,6 +66,7 @@ def __init__(self, model, gene_dataset, use_cuda=True, metrics_to_monitor=None,
         self.use_cuda = use_cuda and torch.cuda.is_available()
         if self.use_cuda:
             self.model.cuda()
+        self.all_dataset = self.create_posterior(self)
```
jeff-regier: I think this should just be `self.create_posterior()`. The first argument is automatically self, right?

jeff-regier: It'd be better to call `create_posterior` in `get_dataset_information`, and just make `all_dataset` a local variable, not a class variable. `create_posterior` is fast so better to call it near where `all_dataset` is used.

scvi/inference/trainer.py
```diff
@@ -220,6 +222,34 @@ def create_posterior(self, model=None, gene_dataset=None, shuffle=False, indices
         return type_class(model, gene_dataset, shuffle=shuffle, indices=indices, use_cuda=self.use_cuda,
                           data_loader_kwargs=self.data_loader_kwargs)
 
+    def get_dataset_information(self, save_imputed=False, file_name_imputation='imputed_values',
+                                save_shape_genes_by_cells=False, save_latent=False, file_name_latent='latent_space'):
```
jeff-regier: The name `get_dataset_information` is really vague. Better to give it a name that describes what this function actual does. Also, good to give a docstring.

scvi/inference/trainer.py
```diff
@@ -220,6 +222,34 @@ def create_posterior(self, model=None, gene_dataset=None, shuffle=False, indices
...
+        self.model.eval()
+        to_get = {"latent": [], "imputed_values": [], "batch_indices": [], "labels": []}
```
jeff-regier: More standard to call this variable `ret` rather than `to_get`.

jeff-regier:
This looks good to me Jules. Is the code exercised by any of unit/integration tests? I believe Eddie set up the pytest to run the notebooks; if you're calling `get_dataset_information` from a notebook it should be getting run. But good to verify that.

jeff-regier:
@romain-lopez Does this PR do what you had in mind for #188 ?

scvi/inference/trainer.py
jules-samaran: Absolutely the additional self is a typing error

jules-samaran: This would work too but I thought it would be useful to have this as an attribute of trainer at the same level than the train_set and test_set attributes. Furthermore there might be situations where we want to draw information on the whole dataset which aren't the latent space or the imputed values so I thought having this attribute might be useful

romain-lopez:
That looks good to me. (Further work is the imputation across different batch. Once Chenling tested her own code from the harmonization we can simply merge her part of the code. )

jules-samaran:
I tested it manually by calling the get_data_information on a dataset in a local file I created and by checking that the order of the cells wasn't changed but running the code of the notebooks would be a good test

scvi/inference/trainer.py
jules-samaran: Wouldn't get_all_latent_and_imputed_values be a little bit too long?

scvi/inference/trainer.py
jeff-regier: Either way is fine. I can sort of see the first point---that it should be on the same level as the posterior objects. Though it's not like all three posterior objects get created on three consecutive lines.

I'm not so receptive to the second, because I think adding functionality that *might* be useful in future tends to make the code less readable. It's harder to keep track of the state of a class variable than local variable.

scvi/inference/trainer.py
jeff-regier: It's an improvement. Short and clear is the best, but too long is better than too vague.

scvi/inference/trainer.py
jules-samaran: I'll make it a local one so we'll just have to change it if it revealed to be useful later

jules-samaran:
@jeff-regier the make test command which is executed by the travis build executes the test_notebook.py code so the notebooks should have been checked

jeff-regier:
I don't think travis runs the makefile. Here it's just running pytest through setup.py:

https://github.com/YosefLab/scVI/blob/43b3787e0d66c9c7905b099a4587ea21aeb4a669/.travis.yml#L19

jeff-regier:
Check out the report from codecov above. It doesn't look like the new code is getting exercised by the tests.

https://codecov.io/gh/YosefLab/scVI/pull/195/diff?src=pr&el=tree#diff-c2N2aS9pbmZlcmVuY2UvZmlzaC5weQ==

jules-samaran:
I executed locally on my computer the basic_tutorial.ipynb and no errors so can merge the branch @jeff-regier if you think this is enough

jeff-regier:
No, the problem is this PR lowers unit test by 1.06%. See the post from Codecov above. We want to increase or at least keep unit test coverage the same. It's about keeping the code tested in the future as people make changes -- not that verifying that it works now.

jeff-regier:
On travis, I see this error:
```
  Could not find a version that satisfies the requirement jupyter==4.1.1 (from -r requirements.txt (line 22)) (from versions: 1.0.0)
No matching distribution found for jupyter==4.1.1 (from -r requirements.txt (line 22))
```

But the newest version of jupyter is 1.0.0. (You might be looking at the version of jupyter-client, a dependency that probably gets installed automatically when you install jupyter.)

jules-samaran:
That's strange because in the ubuntu command line, I get 4.4.0 when I type jupyter --version but I'm going to try withjupyter = 1.0.0

jeff-regier:
You can use `pip list` to see what versions of pip packages you have installed. On my laptop:
```
(tt) Jeffreys-MacBook-Pro-3:scVI jeff$ pip list | grep jupyter
jupyter            1.0.0    
jupyter-client     5.2.3    
jupyter-console    5.2.0    
jupyter-core       4.4.0    
```

jeff-regier:
@jules-samaran I see what's happening. The notebook is indeed being run in a separate process by the unit tests, so codecov isn't aware of that run. Here's the problematic code:
https://github.com/YosefLab/scVI/blob/43b3787e0d66c9c7905b099a4587ea21aeb4a669/tests/test_notebook.py#L23-L25

This looks like a solution:
https://jupyter-notebook.readthedocs.io/en/stable/examples/Notebook/Importing%20Notebooks.html

jules-samaran:
@jeff-regier I'm not sure I read the Codecov report correctly but it doesn't seem that implementing this solution of importing the notebooks as modules as I did is making codecov aware that the code of the notebooks is executed. I will check if some small adjustments might solve that

jeff-regier:
It looks great to me. Codecov says
```
Merging #195 into master will increase coverage by 4.48%.
```
Doesn't the mean it's working?

jeff-regier:
Before your last commit, it was say that this PR would decrease coverage by ~1%.

jeff-regier:
Great job @jules-samaran !

PR #212, opened 2018-11-09
Updating the data_loading notebook and modifying some Dataset subclasses

jules-samaran:
- made it possible to create a Dataset10X instance from 10X data that is not necessarily on the 10X website and that you want to load locally

- Added the possibility to create a CsvDataset instance while giving a labels_file argument which is the name of a csv file containing labels for the cells. However this labels csv file must follow a certain format (labels are in the second column starting on the second row) but this shouldn't be too restrictive

- Updated the data_loading notebook

- minor modifications to the scvi.inference.posterior show_tsne() method, to the scvi.dataset.dataset arrange_categories function and modified some docstrings


jeff-regier:
Looks good!

PR #219, opened 2018-11-16
CITE-seq integration

adamgayoso:
- Implements a new model to handle CITE-seq data 
- Adds a test notebook CITE_seq_scVI that displays basic functionality on the cbmc dataset
- Updates the cbmc dataset class to work with this model while maintaining old functionality by default
- In the general case, assumes proteins are the last columns of the dataset; however a more general extension probably needs to be made to the dataset class to handle multi-modal data, so for now, only works with cbmc data.
- There's an issue that library statistics need to be calculated separately for RNA and Protein data, so current implementation is a bit hacky, where protein lib statistics are calculated outside the model and entered on model initialization. This means that for now, other CITE-seq dataset classes will have to mirror the cbmc example.

Not yet implemented or not tested:

- Differential expression
- Imputation


jeff-regier:
@jules-samaran Would you review the changes to `test_notebooks.py` ?

scvi/inference/cite.py
```diff
@@ -0,0 +1,215 @@
...
+class CitePosterior(Posterior):
+
+    def ll_umi(self, verbose=False):
+        self.mode = 'umi'
+        ll = self.compute_log_likelihood(self.model)
+        if verbose:
+            print("LL UMI: %.4f" % ll)
+        return ll
+
+    def ll_adt(self, verbose=False):
```
jeff-regier: Better to get rid of these "wrapper" methods that just call other another method, and to remove "verbose" logging functionality. The code will be simpler, for exampke, if users just call `compute_log_likelihood` directly rather than calling `ll` which in turn calls `compute_log_likelihood`.

scvi/inference/cite.py
```diff
@@ -0,0 +1,215 @@
...
+        r"""
+        Output average over statistics in a symmetric way (a against b)
+        forget the sets if permutation is True
+        :param vae: The generative vae and encoder network
+        :param data_loader: a data loader for a particular dataset
+        :param M_sampling: number of samples
+        :param mode: UMI or ADT data
+        :return: A 1-d vector of statistics of size n_genes
+        """
+        px_scales = []
```
jeff-regier: This appears to be duplicated code, found in another file. It's best to keep blocks of duplicated code (say, more than 1 or 2 lines) out of the master branch. Otherwise the code base gets unwieldy pretty fast.

jeff-regier:
Oh, I see, the unit test coverage drops by 4% with this PR. We should modify the PR so that it doesn't decrease unit test coverage _before_ merging it.

PR #220, opened 2018-11-20
Cleaning the test dataset file directories after running the tests

### LINKED ISSUE #210: Running the unit tests leaves files all over the place
Currently all the files below get left behind whenever the unit tests run. But the unit test code should delete any files it creates. To use pytest properly, I think we'd want to do something like this: https://docs.pytest.org/en/latest/fixture.html#fixture-finalization-executing-teardown-code

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	Cortex_dataset_imputed_values
	Cortex_dataset_latent_space
	Imputation.png
	spatial-comparison.png
	tests/data/10X/b_cells/filtered_gene_bc_matrices/
	tests/data/10X/cd14_monocytes/filtered_gene_bc_matrices/
	tests/data/10X/cd34/filtered_gene_bc_matrices/
	tests/data/10X/cd4_t_helper/filtered_gene_bc_matrices/
	tests/data/10X/cd56_nk/filtered_gene_bc_matrices/
	tests/data/10X/cytotoxic_t/filtered_gene_bc_matrices/
	tests/data/10X/memory_t/filtered_gene_bc_matrices/
	tests/data/10X/naive_cytotoxic/filtered_gene_bc_matrices/
	tests/data/10X/naive_t/filtered_gene_bc_matrices/
	tests/data/10X/neuron_9k/filtered_gene_bc_matrices/
	tests/data/10X/pbmc4k/filtered_gene_bc_matrices/
	tests/data/10X/pbmc8k/filtered_gene_bc_matrices/
	tests/data/10X/regulatory_t/filtered_gene_bc_matrices/
	tests/data/Rep11_MOB_count_matrix-1.tsv
	tests/data/SeqFISH.xlsx
	tests/data/gene_info_pbmc.csv
	tests/data/neuron_9k/
	tests/data/pbmc_metadata.pickle
	zeiler_clustering_dimension_stability.pdf
```

jules-samaran:
- tests are run inside a temporary directory handled by pytest and which contains a copy of tests/data, this directory is deleted at the end of the test session so there are no residual files

- added missing datasets to test/data so that tests can now be run without having to download anything, without an internet connection

closes #210


jeff-regier:
Also `tests/data/gene_info_pbmc.csv` is thousands of lines currently. Please shorten it to ~10 lines.

tests/test_scvi.py
```diff
@@ -117,79 +122,79 @@ def test_synthetic_3():
...
-    retina_dataset = RetinaDataset(save_path='tests/data/')
+def test_retina(save_path):
+    retina_dataset = RetinaDataset(save_path=save_path)
     base_benchmark(retina_dataset)
 
 
-def test_cite_seq():
-    pbmc_cite_seq_dataset = CiteSeqDataset(name='pbmc', save_path='tests/data/citeSeq/')
+def test_cite_seq(save_path):
+    pbmc_cite_seq_dataset = CiteSeqDataset(name='pbmc', save_path=save_path+'citeSeq/')
```
jeff-regier: Better to use `os.path.join` here (and below). For example,
```
>>> import os
>>> os.path.join("test/data", "specialdir")
'test/data/specialdir'
>>> os.path.join("test/data/", "specialdir")
'test/data/specialdir'
```

romain-lopez:
@jules-samaran, would you correct these changes and push ?

tests/config_notebooks_tests/annotation.config.json
```diff
@@ -1,7 +1 @@
-{
-    "save_path": "tests/data/",
-    "n_epochs": 1,
-    "n_labelled_samples_per_class": 3,
-    "M_permutation": 10,
-    "M_sampling": 1
-}
+{"save_path": "/tmp/pytest-of-jules/pytest-0/temp_data/", "n_epochs": 1, "n_labelled_samples_per_class": 3, "M_permutation": 10, "M_sampling": 1}
```
jeff-regier: Can we avoid hard coding this path somehow? We'd use the temporary directory pytest creates ideally.

jeff-regier: I'm not sure we can always write a directory called `/tmp/pytest-of-jules`. I don't think windows systems even have a `/tmp` directory.

jeff-regier: On the other hand, I'm not quite sure what the right way to do it is, and I think this PR overall is a nice improvement, so let's get it merged.

PR #236, opened 2018-12-03
Using jupyter notebooks without json configuration files

### LINKED ISSUE #234: remove config.json files
Our approach of storing settings for notebooks in `config.json` files has a couple of problems. 
1. It's confusing for users reading the notebooks because the notebooks load settings from these mysterious config files. 
2. It's not great for unit testing either, because the unit tests overwrite some of the config files each time they run, changing the contents of version controlled files.

It's also just kind of complicated.

I think a better approach would be to get rid of these `config.json` files, and write the notebooks with the real settings of variables, e.g. `n_epochs = 500`. Then, in `tests/test_notebook.py`, use regular expressions to change some cells' code to run faster. For example, here
https://github.com/YosefLab/scVI/blob/e979e4a3639e9410226e70b608d6493e2af842f6/tests/test_notebook.py#L69-L71
we might insert the regular expression substitution
```
code = re.sub("n_epoch = \d+$", "n_epoch = 1", code)
```
before the call to `exec`.

jules-samaran:
Instead of having for each notebook two json configuration files containing the values of the parameters we use to run the notebooks (eg the path of the directory containing the datasets, the number of epochs, etc.) either for tests or outside of test sessions, we now modify the code of the notebooks when running them for tests. Inside `test_notebook.py` each notebook is tested by extracting the code from each of its cells and running this collected code. During that step we now replace, in the code extracted, the fragments of code where the parameters are defined  so that the code executed  by `test_notebook.py` has the 'test' parameters (few epochs, small subsampled datasets so that tests run fast). Like this we can just directly put the normal parameters inside the notebooks, and only modify the extracted code when we run the tests, without affecting the .ipynb files. 

In order to run these tests inside the temporary directory created by pytest, we need to retrieve its path (which can only be done a certain way), this is why during the test session we save the path in a file locally in order to be able to use it during tests, and we delete that residual file at the end of the tests session. I can develop the explanation about this if needed.

Closes #234

### DIFF
--- a/tests/notebooks/annotation.ipynb
+++ b/tests/notebooks/annotation.ipynb
@@ -44,14 +44,8 @@
     }
    ],
    "source": [
-    "import os\n",
-    "\n",
     "save_path = 'data/'\n",
-    "n_epochs_all = None\n",
-    "n_labelled_samples_per_class = None\n",
-    "M_sampling_all = None\n",
-    "M_permutation_all = None\n",
-    "n_samples_tsne = None"
+    "n_epochs_all = None"
    ]
   },
   {
@@ -60,13 +54,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "assert n_samples_tsne == 10\n",
-    "assert n_epochs_all == 1\n",
-    "assert n_labelled_samples_per_class == 3\n",
-    "assert M_sampling_all == 1\n",
-    "assert M_permutation_all == 10\n",
-    "\n",
+    "import os\n",
     "import numpy as np\n",
+    "import torch\n",
     "import matplotlib.pyplot as plt\n",
     "from scvi.dataset import CortexDataset, LoomDataset\n",
     "from scvi.models import SCANVI, VAE\n",
@@ -112,7 +102,7 @@
    ],
    "source": [
     "simulation_1 = LoomDataset(filename='simulation_1.loom',\n",
-    "                           save_path=save_path+'simulation/',\n",
+    "                           save_path=os.path.join(save_path, 'simulation/'),\n",
     "                           url='https://github.com/YosefLab/scVI-data/raw/master/simulation/simulation_1.loom')\n",
     "\n",
     "scanvi = SCANVI(simulation_1.nb_genes, simulation_1.n_batches, simulation_1.n_labels)\n",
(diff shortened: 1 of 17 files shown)


tests/notebooks/annotation.ipynb
```diff
@@ -66,6 +60,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "assert n_samples_tsne == 10\n",
+    "assert n_epochs_all == 1\n",
+    "assert n_labelled_samples_per_class == 3\n",
+    "assert M_sampling_all == 1\n",
```
jeff-regier: Does this notebook run when it's not run by unit tests? Wouldn't these assert statements be false?

tests/notebooks/annotation.ipynb
```diff
@@ -45,19 +45,13 @@
...
-    "n_samples_posterior_density = config['n_samples_posterior_density'] if 'n_samples_posterior_density' in config else None\n",
-    "train_size = config['train_size'] if 'train_size' in config else None\n",
-    "M_sampling_all = config['M_sampling'] if 'M_sampling' in config else None\n",
-    "M_permutation_all = config['M_permutation'] if 'M_permutation' in config else None\n",
-    "n_labelled_samples_per_class = config['n_labelled_samples_per_class'] if 'n_labelled_samples_per_class' in config else None"
+    "save_path = 'data/'\n",
+    "n_epochs_all = None\n",
+    "n_labelled_samples_per_class = None\n",
+    "M_sampling_all = None\n",
+    "M_permutation_all = None\n",
```
jeff-regier: I'm not sure I understand these. Why are they set to None? Don't we want to set them to the recommended values?

tests/notebooks/annotation.ipynb
jules-samaran: Thanks for pointing this out, I forgot to delete it after testing that the parameters's values were replaced, I'll commit this fix

tests/notebooks/scRNA_and_smFISH.ipynb
```diff
@@ -496,7 +487,8 @@
    "outputs": [],
    "source": [
     "from sklearn.manifold import TSNE\n",
-    "def get_common_t_sne(latent_seq, latent_fish, n_samples=1000):\n",
+    "n_samples_tsne = 1000 if n_samples_tsne is None else n_samples_tsne\n",
```
jeff-regier: I think it would be better just to have `n_samples_tsne = 1000` here (without checking for None values). We can still find and replace the code in the unit tests. (Use `\d+` in a regex to match any digits)

tests/notebooks/annotation.ipynb
jules-samaran: I think the method in my last commit is better and clearer

jules-samaran:
Rather than setting the parameters to None and the, when we are testing the notebooks, replacing this assignment with the assignment `parameter = test_value` and then using the boolean `parameter is None` to identify if the notebook is currently being run as a test or not, I now target the assignments `parameter = true_value` except for n_epochs_all (since in one notebooks there are different variables for n_epochs eg n_epochs_vae, n_epochs_scanvi it is simpler to use n_epochs_all)

PR #245, opened 2018-12-18
switch to os.path.join

cgreene:
minor improvement to improve cross-platform when labels are passed


jeff-regier:
Thanks Casey! We appreciate the contribution.

PR #250, opened 2018-12-19
Minor clarifications on how to load data with scvi

jules-samaran:
Several users encounter errors when they try to load data that is stored locally on their machine because they make mistakes when entering the `filename` and `save_path` arguments. 

- Replacing occurences of path1 + path2 with os.path.join(path1, path2) is an improvement.
- Added some recommandations in the notebook
- Added a message to the user when the module can't find the data because the path is broken

### DIFF
--- a/scvi/dataset/dataset.py
+++ b/scvi/dataset/dataset.py
@@ -240,6 +240,9 @@ class GeneExpressionDataset(Dataset):
         if os.path.exists(os.path.join(save_path, download_name)):
             print("File %s already downloaded" % (os.path.join(save_path, download_name)))
             return
+        if url is None:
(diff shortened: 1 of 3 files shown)


jules-samaran:
@jeff-regier I think that having two arguments, `filename` and `save_path` can be useful when downloading data for example, this is the kind of situation where we need them separately. However we could manually from a single path given by the user decompose it into a filename and a save_path, this might make it simpler for users who wish to load local data and it wouldn't change much when we download data on a remote server

jeff-regier:
I see. Maybe better not to worry about it for now then. Let's just merge this since it's clearly an improvement.

PR #257, opened 2019-01-03
Speed up performance on brain_large

triyangle:
via storing intermediate values and reordering calculations

### DIFF
--- a/scvi/models/log_likelihood.py
+++ b/scvi/models/log_likelihood.py
@@ -86,8 +86,9 @@ def log_zinb_positive(x, mu, theta, pi, eps=1e-8):
         torch.lgamma(x + theta) - \
         torch.lgamma(theta) - \
         torch.lgamma(x + 1)
+    mul_case_non_zero = torch.mul((x > eps).type(torch.float32), case_non_zero)
 
-    res = mul_case_zero + torch.mul((x > eps).type(torch.float32), case_non_zero)
+    res = mul_case_zero + mul_case_non_zero
 
     return torch.sum(res, dim=-1)
 
@@ -105,8 +106,10 @@ def log_nb_positive(x, mu, theta, eps=1e-8):
     if theta.ndimension() == 1:
         theta = theta.view(1, theta.size(0))  # In this case, we reshape theta for broadcasting
 
-    res = theta * (torch.log(theta + eps) - torch.log(theta + mu + eps)) + \
-        x * (torch.log(mu + eps) - torch.log(theta + mu + eps)) + \
+    log_theta_mu_eps = torch.log(theta + mu + eps)
+
+    res = theta * (torch.log(theta + eps) - log_theta_mu_eps) + \
+        x * (torch.log(mu + eps) - log_theta_mu_eps) + \
         torch.lgamma(x + theta) - \
         torch.lgamma(theta) - \
         torch.lgamma(x + 1)
--- a/scvi/models/log_likelihood.py
+++ b/scvi/models/log_likelihood.py
@@ -72,14 +72,23 @@ def log_zinb_positive(x, mu, theta, pi, eps=1e-8):
     if theta.ndimension() == 1:
         theta = theta.view(1, theta.size(0))  # In this case, we reshape theta for broadcasting
 
-    case_zero = (F.softplus((- pi + theta * torch.log(theta + eps) - theta * torch.log(theta + mu + eps))) -
-                 F.softplus(-pi))
+    softplus_pi = F.softplus(-pi)
+    log_theta_eps = torch.log(theta + eps)
+    log_theta_mu_eps = torch.log(theta + mu + eps)
+    pi_theta_log = - pi + theta * (log_theta_eps - log_theta_mu_eps)
 
-    case_non_zero = - pi - F.softplus(-pi) + theta * torch.log(theta + eps) - theta * torch.log(
-        theta + mu + eps) + x * torch.log(mu + eps) - x * torch.log(theta + mu + eps) + torch.lgamma(
-        x + theta) - torch.lgamma(theta) - torch.lgamma(x + 1)
+    case_zero = F.softplus(pi_theta_log) - softplus_pi
+    mul_case_zero = torch.mul((x < eps).type(torch.float32), case_zero)
 
-    res = torch.mul((x < eps).type(torch.float32), case_zero) + torch.mul((x > eps).type(torch.float32), case_non_zero)
+    case_non_zero = - softplus_pi + \
+        pi_theta_log + \
+        x * (torch.log(mu + eps) - log_theta_mu_eps) + \
+        torch.lgamma(x + theta) - \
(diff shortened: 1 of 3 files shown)


scvi/models/log_likelihood.py
```diff
@@ -72,14 +72,23 @@ def log_zinb_positive(x, mu, theta, pi, eps=1e-8):
...
+    mul_case_zero = torch.mul((x < eps).type(torch.float32), case_zero)
+
+    case_non_zero = - softplus_pi + \
+        pi_theta_log + \
+        x * (torch.log(mu + eps) - log_theta_mu_eps) + \
+        torch.lgamma(x + theta) - \
+        torch.lgamma(theta) - \
+        torch.lgamma(x + 1)
+
+    res = mul_case_zero + torch.mul((x > eps).type(torch.float32), case_non_zero)
```
romain-lopez: Should we have a mul_case_non_zero variable ?

scvi/models/log_likelihood.py
```diff
@@ -96,7 +105,10 @@ def log_nb_positive(x, mu, theta, eps=1e-8):
...
         theta = theta.view(1, theta.size(0))  # In this case, we reshape theta for broadcasting
 
-    res = theta * torch.log(theta + eps) - theta * torch.log(theta + mu + eps) + x * torch.log(
-        mu + eps) - x * torch.log(theta + mu + eps) + torch.lgamma(x + theta) - torch.lgamma(
-        theta) - torch.lgamma(x + 1)
+    res = theta * (torch.log(theta + eps) - torch.log(theta + mu + eps)) + \
+        x * (torch.log(mu + eps) - torch.log(theta + mu + eps)) + \
+        torch.lgamma(x + theta) - \
+        torch.lgamma(theta) - \
+        torch.lgamma(x + 1)
```
romain-lopez: Should we add some named variable here too ?

tests/test_brain_large_speedup.py
```diff
@@ -0,0 +1,94 @@
...
+"""Test improved calculation of log_zinb_forward for better performance on brain_large."""
+
+from scipy.sparse import random
+import torch
+import torch.nn.functional as F
+
+batch_size = 128
+num_genes = 720
+
+x = torch.from_numpy(random(batch_size, num_genes,
```
romain-lopez: This is a hacky way of simulating data but is probably fine. If you're curious, we have a simulation API I believe here, which would be probably more appropriate
https://github.com/YosefLab/scVI/blob/master/scvi/dataset/synthetic.py

scvi/models/log_likelihood.py
triyangle: Yeah I've added it now

scvi/models/log_likelihood.py
triyangle: Yeah good catch, I've added `log_theta_mu_eps = torch.log(theta + mu + eps)` now

tests/test_brain_large_speedup.py
triyangle: Thanks for the pointer, I've updated to use `SyntheticDataset` now

scvi/models/log_likelihood.py
```diff
@@ -72,14 +72,24 @@ def log_zinb_positive(x, mu, theta, pi, eps=1e-8):
...
+    case_zero = F.softplus(pi_theta_log) - softplus_pi
+    mul_case_zero = torch.mul((x < eps).type(torch.float32), case_zero)
+
+    case_non_zero = - softplus_pi + \
+        pi_theta_log + \
+        x * (torch.log(mu + eps) - log_theta_mu_eps) + \
+        torch.lgamma(x + theta) - \
+        torch.lgamma(theta) - \
+        torch.lgamma(x + 1)
+    mul_case_non_zero = torch.mul((x > eps).type(torch.float32), case_non_zero)
```
romain-lopez: Very good

tests/test_brain_large_speedup.py
```diff
@@ -0,0 +1,100 @@
...
+from scvi.dataset import SyntheticDataset
+import torch
+import torch.nn.functional as F
+
+batch_size = 128
+nb_genes = 720
+n_batches = 1
+
+
+synthetic_dataset = SyntheticDataset(batch_size=batch_size, nb_genes=nb_genes,
```
romain-lopez: Very good

tests/test_brain_large_speedup.py
```diff
@@ -0,0 +1,100 @@
...
+            torch.lgamma(x + theta) - \
+            torch.lgamma(theta) - \
+            torch.lgamma(x + 1)
+
+        res = torch.mul((x < eps).type(torch.float32), case_zero) + \
+            torch.mul((x > eps).type(torch.float32), case_non_zero)
+
+        return torch.sum(res, dim=-1)
+
+    def new_method(x, mu, theta, pi, eps=1e-8):
```
jeff-regier: The unit test is suppose to test the actual code (the code in under the `scvi` directory). Here it looks like the test is just testing the test. @triyangle @romain-lopez  Am I misunderstanding what's going on here?

jeff-regier: So to write a unit test, you don't general duplicate code. You call the code you want to test that is in the `scvi` directory directly, rather than copying that code to the test file. Does that make sense?

triyangle: Based on my understanding, I believe you're saying we should be calling the functions I've modified directly from `scvi` instead of copying the new implementation over? That makes sense to me, but would the test still have the old implementation to compare the results between the new and old implementations then? For reference I based this test on [test_remove_empty](https://github.com/YosefLab/scVI/blob/master/tests/test_remove_empty.py).

jeff-regier:
That's it. No, typically unit tests wouldn't have an old implementation either. They would just contain some results from the old implementation hard coded.

jeff-regier:
Ah I see. `test_remove_empty` may need to be revised.

triyangle:
Thanks for the clarifications! So, in this case, I would be running the previous implementation on some input to get a result which will then be hardcoded in the test to be compared against the result from calling the new implementation of the function on the same input? Let me know if this sounds right and I can work on updating the tests appropriately and maybe take a look at `test_remove_empty` as well.

jeff-regier:
Yes, that's it. You might test it with several different input/output pairs. Thanks for making the change!

jeff-regier:
I think it'd be fine to just delete the file containing `test_remove_empty` ... It's not actually testing anything in the codebase. William, would you delete that file along with a new pull request?

triyangle:
Alright thanks, I'll work on updating these tests and submit a new pull request for the changes. As for `test_remove_empty`, I do think it tests the modified portion of code in `get_attributes_from_matrix` from this [commit](https://github.com/YosefLab/scVI/commit/d4be31361110e75274f610194a4376bdbb3e9ca9), but may also need to be updated to use the testing pattern we're adopting for these tests.

PR #271, opened 2019-02-08
added functions for sampling gene expression scale from vae

chenlingantelope:
I added scale_from_z to vae scvi/models/vae.py and get_harmonized_scale to scvi/inference/posterior.py to compute differential expression

### DIFF
--- a/scvi/inference/posterior.py
+++ b/scvi/inference/posterior.py
@@ -310,6 +310,15 @@ class Posterior:
         return libraries.ravel()
 
     @torch.no_grad()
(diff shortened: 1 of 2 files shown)


romain-lopez:
@chenlingantelope you need your code to pass the flake8 as well as the unit tests

romain-lopez:
Please resubmit

chenlingantelope:
@romain-lopez looks like it passed the test so I merged it. If you can double check it that would be good.

PR #272, opened 2019-02-09
get_all_latent_and_imputed_values(): skip impute

sjfleming:
Add the option to skip imputation when calling trainer.get_all_latent_and_imputed_values().
When working on datasets with hundreds of thousands of cells and thousands of variable genes, imputation leads to unacceptably-high memory use.  I would like a way to just return the latents without doing imputation.

### DIFF
--- a/scvi/inference/trainer.py
+++ b/scvi/inference/trainer.py
@@ -222,7 +222,7 @@ class Trainer:
     @torch.no_grad()
     def get_all_latent_and_imputed_values(self, save_imputed=False, filename_imputation='imputed_values',
                                           save_shape_genes_by_cells=False, save_latent=False,
-                                          filename_latent='latent_space'):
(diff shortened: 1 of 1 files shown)


romain-lopez:
Hi, 

Thanks for your proposition, it looks like code get_all_latent_and_imputed_values() should be present in the posterior class and not the trainer class. I will refactor this code in the future and in the meantime, I suggest to use the code Posterior().get_latent(). 

Also, your pull request has some flake8 errors. 

Best, 
Romain

romain-lopez:
Hi,

Thanks a lot for your contribution to scVI, unfortunately, we have opted for a more principled change of API by deleting this function get_all_latent_XXX and having the end user playing with the posterior class.
We are updating the notebooks that should be pushed today to reflect these changes, 

Best, 
Romain

PR #274, opened 2019-02-18
Linearly decoded VAE

vals:
In this pull request I have added an LDVAE subclass of VAE, as described in #222 

I think a separate model inheriting most of the functionality from the VAE model is probably less confusing than having a flag in VAE.

I also added a notebook demonstrating the use of the model on 48,000 cells from mouse visual cortex.


jeff-regier:
Looks good! Thanks for the contribution.

Would mind fixing the two flake8 errors?
```
./scvi/models/vae.py:256:1: W293 blank line contains whitespace
./scvi/models/vae.py:259:1: W391 blank line at end of file
```

vals:
Should be fixed now!

I also added a little helper method to extract the 'loadings' from the decoder.

jeff-regier:
Thanks!

It looks like the unit test coverage is down with this pull request. Does the notebook you added get run by pytest?

jeff-regier:
(We mainly do "smoke testing" for scVI, where we run the code without checking output, usually by having our unit tests run our Jupyter notebooks.)

vals:
Ah, hm, I didn't make it run my notebook. I used a kind of large dataset that's not in the repo, and it takes a while. (It also uses some non-scvi packages).

I could make a little script that just runs it on a smaller dataset as a way to check that it doesn't throw errors?

jeff-regier:
We have the same issue with some of our notebooks that typically train for many epochs. In the file `tests/run_notebooks.py`, we'll use regex substitution to dynamically change the code in the notebooks to run more quickly. E.g.,
https://github.com/YosefLab/scVI/blob/07f77a2480397fa83a9dcf01bbf4ffd03fa315da/tests/test_notebook.py#L78

Might a similar approach work for your PR? The nice thing about this approach is that it smoke tests the notebook too.

vals:
That's clever! I'll do that

But I will be traveling for a couple of weeks, so I guess I'll let this PR stew a bit if that's fine?

romain-lopez:
Yes np Valentine, thanks for your contribution

vals:
I ended up writing tests in the `test_scvi.py` file rather than testing the notebook. Codecov now shows 100%: https://codecov.io/gh/YosefLab/scVI/pull/

Is this all right?

jeff-regier:
That could be fine. The only down side is that someone might make change that break the code in the notebook, but the unit tests would still pass. On the other hand, it's good to get things merged. @romain-lopez , what do you think? Are we planning to link to this new notebook in the readme?

jeff-regier:
And thanks for all your work, Valentine!

romain-lopez:
Hi Valentine, 

Thanks for pushing this, I appreciate you testing the code. I am a bit worried about the readability of this notebook. I would be very happy to merge this but the notebook you added does not have much comments or text and uses additional library as well as a local dataset of your own. It would be very hard for a user to reproduce / run your notebook. 

I have several questions to try to improve this work.

(1) Is there a way that you could copy / paste some of the text you added on the very first github issue that you posted on this repo ?
(2) Is there a way to use on of our own datasets, especially if the conclusion for the "loadings" are the same than for example the PBMC seurat notebook ? https://satijalab.org/seurat/pbmc3k_tutorial.html

If you have no time, I will delay this merge until we find someone here that can make this notebook more readable.

vals:
Hi Romain,

Ok, I think what I will do is remake the notebook with the included PBMC data instead of the Hrvatin et al data.

Personally I generally try to look at newer data whenever I try methods. It's easy to get stuck looking at old data that isn't so relevant anymore (c.f. all the new methods reporting results based on the Zeisel 3,005 cells data from 2015 that used Fluidigm...)

It will take me a bit, I have a couple of upcoming deadlines in the next week. But I will get to it eventually.

Best,
/Valentine

romain-lopez:
Hi @vals, it look like there is a conflict and the unit test cannot be run ?

vals:
Hey @romain-lopez it's not finished yet, not sure how to push to my Github branch in a way that doesn't launch the CI.

vals:
Hi @romain-lopez , let me know what you think of the updated notebook.

romain-lopez:
Hi Valentine, 

Thanks for putting more work on this notebook, it looks really good now!

Best, 
Romain

PR #276, opened 2019-03-04
Add scale/rate/dropout decoders to the main decoder as modules

gokceneraslan:
This fixes things like `model.decoder.cpu()` which converts CUDA model to a CPU model.


gokceneraslan:
That's really strange. I don't get an error even without the PR now, so closing this.

PR #277, opened 2019-03-05
Classifier accuracy early stopping

davek44:
Previously, the accuracy metric couldn't be used for early stopping of training because the parent class Posterior of AnnotationPosterior didn't recognize the attribute accuracy:

`AttributeError: type object 'Posterior' has no attribute 'accuracy'
scvi/inference/trainer.py:279: AttributeError`

These modifications enable early stopping with the accuracy metric and demonstrate that it works with a test.


jeff-regier:
Looks good to me. Thanks David!

@romain-lopez Should we merge this PR?

romain-lopez:
Looks good to me too, thanks David!

PR #280, opened 2019-03-11
Zl encoder

davek44:
These commits add functionality to classifiers hooked on top of a VAE. First, I added access to soft classifications, before argmax. Second, I added an option to concatenate the VAE l_encoder output to the z_encoder output so that the classifier has access to the library depth estimate.


romain-lopez:
Looks good to me, thanks for your contribution @davek44

PR #289, opened 2019-04-09
Scanpy compatibility

### LINKED ISSUE #206: scanpy integration

ANazaret:
closes #206


tests/test_notebook.py
```diff
@@ -76,7 +76,9 @@ def load_module(self, fullname):
                         code = re.sub(r'n_samples_tsne = \d+', "n_samples_tsne = 10", code)
                         code = re.sub(r'n_samples_posterior_density = \d+', "n_samples_posterior_density = 2", code)
                         code = re.sub("save_path = 'data/'", "save_path = '"+os.getcwd()+"'", code)
+                        code = re.sub('save_path = "data/"', "save_path = '"+os.getcwd()+"'", code)
```
romain-lopez: is that a duplicate line ?

ANazaret: This line was added on purpose. 
The notebook tests failed locally because the regex was looking for ... = 'data/' and it was ... = "data/" in the code.
There are no best practices in Python regarding simple quotes or double quotes, the tests will now account for both practices.

romain-lopez: That looks dangerous. Ok let's stick to that, thanks for pointing it out

romain-lopez:
Is there a way to increase this code coverage ? That would be helpful

romain-lopez:
supposedly we could just test more the anndata class with a toy test in test_scvi ?

ANazaret:
Good point, the AnnDataset is now tested for AnnData inputs containing dense data.
The remaining uncovered code (2 lines) are just helpful exceptions to notify the user about wrong input type or not implemented abstract functions.

scvi/inference/posterior.py
```diff
@@ -208,15 +208,14 @@ def sample_scale_from_batch(self, n_samples, batchid=None, selection=None):
         else:
             if selection.dtype is np.dtype('bool'):
                 selection = np.asarray(np.where(selection)[0].ravel())
+        old_loader = self.data_loader
```
jeff-regier: Why are we creating a new variable called `old_loader`? Is there a way to make this whole block of code simpler? I'm not sure a few lines below why we'd overwrite `self.data_loader` once for each batch id in the first place.

.travis.yml
```diff
@@ -13,6 +13,7 @@ notifications:
 
 install:
   - pip install --quiet -r requirements.txt
+  - pip install --quiet -r requirements.test.txt
```
jeff-regier: Can we put these new requirements in `requirements.txt` rather than creating a new file? I think `requirements.txt` is only used ever for travis testing.

setup.py
```diff
@@ -35,6 +35,8 @@
 
 test_requirements = ['pytest', ]
 
+extras_requirements = {'test': ['scanpy', 'louvain', ]}
```
jeff-regier: What happens if we list these as test_requirements rather than extras_requirements?

ANazaret: It doesn't work. Tox doesn't handle well test_requirements ...
Here is a description of the problem, with a solution using extras (deps= .[test])
https://stackoverflow.com/questions/29870629/pip-install-test-dependencies-for-tox-from-setup-py

I rather use the suggested syntax from tox documentation to install extras: extras=test

.travis.yml
ANazaret: Alright! It makes sense (not only travis, also developer that forks the repo can use requirements.txt. But it makes sense to install the required test packages as a developer)

scvi/inference/posterior.py
jeff-regier: (I may just be missing something, but if the whole reason for introducing `old_loader` is because `self.data_loader` is getting overwritten below unnecessarily, it'd be better to avoid all the unnecessary assignment.)

ANazaret: The purpose of the function is to impute the scales for each batch individually. To do this, you must sample the data in only one batch at a time. The `DataLoader' class is exactly designed for this task as soon as you change its `Sampler' to sample only from a subset of inputs. 

There was a huge problem with the way it was done before. It was overwriting `self.data_loader' without restoring it as it was before. Indeed, he restored it with a RandomSampler, but in many cases - as after calling .sequential on the Posterior - the sampler was not Random but Sequential. When I was asking for the first input, I was receiving a random one by the Sampler.

Using `old_loader` fixes this and restores the DataLoader: no more side effect!

romain-lopez:
@ANazaret let me know when you push the requirements.txt

PR #299, opened 2019-04-20
add automatic hyperparameter tuning with hyperopt

### LINKED ISSUE #258: automatic hyperparameter tuning
It'd be nice if scVI would find good settings for it's hyperparameters automatically.

https://github.com/hyperopt/hyperopt

gabmis:
closes #258

### DIFF
--- a/scvi/inference/inference.py
+++ b/scvi/inference/inference.py
@@ -1,9 +1,18 @@
+from collections import defaultdict
 import copy
+from functools import partial
+import logging
+import os
+import sys
+from typing import Any, Dict, Type, Union
 
+from hyperopt import fmin, tpe, Trials, hp, STATUS_OK
 import matplotlib.pyplot as plt
 import torch
 
 from . import Trainer
+from ..dataset import GeneExpressionDataset
+from ..models import VAE
 
 plt.switch_backend('agg')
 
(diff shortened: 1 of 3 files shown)


gabmis:
Still need to integrate in basic_tutorial notebook, so that it is visible and tested.

scvi/inference/inference.py
```diff
@@ -71,3 +80,221 @@ def train(self, n_path=10, n_epochs=50, **kwargs):
...
+
+    Examples:
+        >>> gene_dataset = CortexDataset()
+        >>> best_trainer, trials = auto_tuned_scvi_parameters(gene_dataset)
+        >>> posterior = best_trainer.create_posterior()
+    """
+    logging.warning("Because of incompatibility between scVI and Hyperopt progress bar,"
+                    "during training, sys.stdout is redirected to os.devnull"
+                    "For example, you will not see the result of any 'print(..)' "
+                    " during the training process.")
```
jeff-regier: Is this issue only when it's run from a Jupyter notebook?

scvi/inference/inference.py
```diff
@@ -71,3 +80,221 @@ def train(self, n_path=10, n_epochs=50, **kwargs):
...
+    space: dict = None,
+    use_batches: bool = False,
+    max_evals: int = 10,
+    verbose: bool = True,
+) -> (Type[Trainer], Trials):
+    """Perform automatic hyperparameter optimization of an scVI model
+    and return best model and hyperopt Trials object.
+    Trial object contains hyperparameter space and loss history for each trial.
+    We provide a default hyperparameter search space (see source code),
+    but we recommend the user to build a custom one for each application.
```
jeff-regier: If at all possible, we should find a single set of bounds for the hyperparameter search space that work for all applications. It's important for three reasons. 1) Most users will have no idea how to customize the search space, 2) It makes it hard to compare results if we are using multiple procedures for selecting hyperparameters, and 3) reviewers for our papers have been uncomfortable with procedures that have free parameters to set.

How about by default we use a search space that is so large that it will work for all applications, but we give the user the option to override those defaults if he/she happens to know tighter bounds?

scvi/inference/inference.py
```diff
@@ -71,3 +80,221 @@ def train(self, n_path=10, n_epochs=50, **kwargs):
...
+    trainer = trainer_class(
+        model,
+        gene_dataset,
+        **trainer_specific_kwargs,
+    )
+
+    # redirect scVI progress bar to null because of incompatibility with hyperopt
+    f = open(os.devnull, 'w')
+    orig_stdout = sys.stdout
+    sys.stdout = f
```
jeff-regier: It's clever you thought of this, but if it's just to avoid errors from Jupyter notebook, should we make this redirect something we do in the Jupyter notebook rather than in the function?

gabmis: The bad behaviour also happens when running in script
image

jeff-regier: Looks like an issue with tqdm, which both scVI and hyperopt have as dependencies. It isn't supposed to work that way. It may also have to do with how the terminal executible was compiled. Let me search a bit for bug reports.

jeff-regier: Do you have the latest version of the tqdm installed? If not, would you try manually installing the latest one to see if the undesirable behavior persists? It might be that some dependency is keeping the latest version of tqdm from getting installed. I see some bug reports online that suggest this issue with tqdm may have been fixed.

gabmis: It's still there

image

The issue is documented in hyperopt and might cause them to revert their progbar apparently: https://github.com/hyperopt/hyperopt/issues/476

gabmis: A possible workaround would be to make scVI's progress bar optional
It seems like the best option right now

jeff-regier: That seems like a good solution

romain-lopez:
I am not familiar with hyperopt but the code looks very readable. I can help out defining the default parameters early next week!

gabmis:
@jeff-regier I've looked into the mongodb paralel opt. The way I see it we have a parameter `parallel` in the `auto_tune_scvi_model` function which if set to True will run a bash script which sets up the mongo daemon and one worker per available gpu.
This requires that we are able to get the number of available gpus and their ids. This is non-standard info which will probably mean that we will have to restrict ourself to a somewhat specific range of platforms/gpus. I believe we should build this as if the program was running on linux with nvidia gpus (as it is in aws, right?).
Another issue we need to tackle is the question of testing (at least once), do we have a machine with multiple gpus available somewhere, I am not sure I can instantiate one in our aws cluster.
If we end up being successful, I suggest we write a small tutorial on how to install mongodb and run the optimization.

jeff-regier:
Might `torch.cuda.device_count()` give you the right number on both Linux and Mac?

Can we use a remote mongodb instance rather than launching one locally? I guess either is fine but a remote one might make it easier to use multiple machines, and it would make configuration easier. Those free instances should suffice for us.

@romain-lopez will you file a ticket with aws to asking to let everyone on the scVI account launch the p2.8xlage and p2.16xlarge servers?

gabmis:
Yes good idea. I can use torch to get the gpu count and pass that to the script which launches the workers.

Regarding the mondo daemon the way I see it is the `auto_tune_scvi_model` function should be called on the remote instance thus launching a local mongod on the remote instance. Isn’t that what is usually done?
Also, when you say multiple machines, do you mean we’d want to run the trainings on different kernels (several linux machines for instance) ? I thought we were talking about a single machine with several gpus..

jeff-regier:
If you register for a free mongodb server here,no don't think you'd need to launch a daemon yourself on ec2, right? 
https://docs.mongodb.com/manual/tutorial/atlas-free-tier-setup/

The main advantage of this is that you don't have to run/configure mongodb yourself. I'm not sure how difficult that is but it could be a pain. However if you've already figured it out then it's fine to do that -- multi-machine support isn't necessary for now

jeff-regier:
Come to think of it, Gabriel, you may also be able to file a ticket with aws to request the ability to launch p2.8xlarge instances. I just gave their help line a call before and they added it to my account.

gabmis:
MongoDb Community Edition is pretty standard stuff in terms of install (e.g. packaged versions for standard (even "Amazon") linux). I don't think running a daemon requires any specific config, I don't believe you need an account for example. I am not 100% on this because I already had it installed, but i'll go through the install from scratch asap.
The process of registering and starting an Atlas cluster seems more cumbersome.

gabmis:
Thanks for pointing it out, I have requested a limit increase for these machines.

romain-lopez:
@gabmis  great to hear, this is exciting ! Let's just make sure to not run these machines for too long, it's really expensive to work with them :) you can check the cost on the aws website

gabmis:
as a conveniently placed reminder
Screenshot 2019-04-22 at 10 40 51

jeff-regier:
The new code looks good. Ready to merge? I see this PR lowers test coverage but I think that's probably OK since it doesn't seem straightforward to write tests for it, and most bugs aren't going to be so subtle.

gabmis:
I want to switch to calls to the python hyperopt code instead of calling the hyperopt-mongo-worker script, for several reasons. I also wanted to include the code in the basic_tutorial notebook, thus making it more visible, the non-parallel method could be used at test time. I was also planning on thinking about test for the parallel version, but it doesn't seem straightforward

jeff-regier:
Actually it might not be so hard to test the parallel code. It looks like we'd just need to add these lines to the .travis-ci.yml file:

```
services:
  - mongodb


before_script:
  - sleep 15  # mongo takes time to start
  - mongo mydb_test --eval 'db.createUser({user:"travis",pwd:"test",roles:["readWrite"]});'
```

See
https://github.com/hyperopt/hyperopt/blob/0465230ad8f5a16a97d757d512f632dcf1a616c9/.travis.yml#L12-L19

gabmis:
looks nice, I was looking into things that were more about mocking mondodb collections but this would be much easier!

gabmis:
If tests pass, ready to merge.

jeff-regier:
Outstanding! :confetti_ball: :tada:

PR #304, opened 2019-04-24
feat synthetic correlated class

### LINKED ISSUE #303: confusion matrix for classification of zeros with synthetic data
With simple-but-non-trivial synthetic data, let's test that scVI correctly classifies zeros as either biological (i.e., the expression level really is low) or technical (i.e., due to a "dropout" event). Results could be reported as a confusion matrix.

It'd also be nice to make the code for this experiment into a test within the `tests` directory -- basically we want to test that scVI usually predicts biological vs technical correctly. It would be a long-running test since each time we have to train a model, so it can't run on travis-ci. We can skip it by default this way:
https://docs.pytest.org/en/latest/example/simple.html#control-skipping-of-tests-according-to-command-line-option

PierreBoyeau:
The purpose of the PR is to create new synthetic data class for which imputation makes sense. We consider clusters of cells for which specific genes are highly expressed and others have lower expression.
This piece of code is used in to validate our PPC approach.

closes #303


romain-lopez:
"Merging #304 into master will decrease coverage by 1.3%."

Thanks for this contribution, shall we add some unit tests to increase the coverage ? You can add them in test_scvi.py

jeff-regier:
We're getting there. This is a work in progress. We had in mind using this new class to scVI's ability to classify zeros as either technical or biological. Unfortunately that kind of test requires training a neural network so it will take too long to run on travis-ci (which does not have a GPU). Therefore, that test wouldn't increase coverage on travis-ci. I'm not exactly sure what to do about that.

Perhaps we could have pytest run the new test with just one epoch of training if pytest isn't invoked without a special flag (e.g., `--long-running`) and also only check that conditions help (e.g., most zeros are correctly identified) if the special flag is present. That way we get higher coverage on travis (smoke testing) and, with the `--long-running` flag, we can also test for problems with accuracy.

scvi/dataset/synthetic.py
```diff
@@ -42,3 +42,113 @@ def __init__(self, mu=4.0, theta=2.0, dropout=0.7, save_path='data/'):
...
+            newdata = data * mask
+        labels = np.random.randint(0, n_labels, size=(n_batches, batch_size, 1))
+        super().__init__(
+            *GeneExpressionDataset.get_attributes_from_list(newdata, list_labels=labels),
+            gene_names=np.arange(n_genes_total).astype(np.str))
+
+
+class CorrNBDataset(SyntheticDatasetCorrelated):
+    def __init__(self):
+        super().__init__(mode='nb')
```
jeff-regier: If these classes just call the constructor for `SyntheticDatasetCorrelated`, I'm not sure it's worth having them. It's probably clearer just to call `SyntheticDatasetCorrelated` with the right mode directly.

Alternatively, you might remove the `mode` parameter from `SyntheticDatasetCorrelated`, keep these subclasses, and implement the mode-specific behavior in the subclasses by overriding a method in the parent class. (Rather than branching on mode.) If you're overriding methods that's a good use of subclasses. Either way seems fine and preferable to the current flow.

scvi/dataset/synthetic.py
```diff
@@ -42,3 +42,113 @@ def __init__(self, mu=4.0, theta=2.0, dropout=0.7, save_path='data/'):
...
+
+            submask = np.random.binomial(n=1, p=1 - p_dropout,
+                                         size=(n_batches, batch_size, n_genes_zi))
+            mask = np.ones_like(data)
+            random_permutation = np.random.permutation(np.arange(n_genes_total))
+            self.zi_genes_idx = random_permutation[:n_genes_zi]
+            mask[:, :, random_permutation[:n_genes_zi]] = submask
+
+            newdata = data * mask
+        labels = np.random.randint(0, n_labels, size=(n_batches, batch_size, 1))
```
jeff-regier: How about giving all the cells in each cluster the same label? Then we can plot the tsne of our latent space to confirm that all the cells with the same label are clustered together. (The labels are used to color code the tnse plots in one of our functions)

scvi/dataset/synthetic.py
```diff
@@ -42,3 +42,113 @@ def __init__(self, mu=4.0, theta=2.0, dropout=0.7, save_path='data/'):
...
+        :param weight_low: Level weight for lowly expressed genes
+        :param lam_0: Proportionality in the Poisson distributions parameter
+        :param p_dropout: Probability of dropout (only makes sense if mode='zi'
+        or mode='mixed'
+        :param n_batches: (useless for now) Number of batches in dataset
+        :param n_labels: (useless for now) Number of labels
+        :param mode: Should be either
+            - 'nb': mode without zero inflation
+            - 'zi': mode including zero inflation for ALL genes
+            -'mixed': mode in which SOME genes are Zero-Inflated while some other are not
```
jeff-regier: Do we need "mixed" mode?

scvi/dataset/synthetic.py
oscarclivio: Ok thanks, indeed this would be more readable. We are implementing the second option.

scvi/dataset/synthetic.py
oscarclivio: Great, thanks for this advice!

scvi/dataset/synthetic.py
```diff
@@ -42,3 +42,113 @@ def __init__(self, mu=4.0, theta=2.0, dropout=0.7, save_path='data/'):
...
+
+        if n_genes_total % n_clusters > 0:
+            print("Warning, clusters have inequal sizes")
+
+        if n_genes_high > (n_genes_total // n_clusters):
+            print("Overlap of", n_genes_high - (n_genes_total // n_clusters), "genes")
+
+        # Generate data before dropout
+        batch_size = n_cells_cluster * n_clusters
+        data = np.ones((n_batches, batch_size, n_genes_total))
```
jeff-regier: Better to give this a more descriptive name than "data" (all variables store data). You might call it the `expression_mat` (expression matrix), or `count_mat`, or something like that.

scvi/dataset/synthetic.py
oscarclivio: Well, it may be very specific to our work on comparing ZINB and NB models.
Basically, the mixed classes would allow us to verify that our approach is able to differentiate between ZI and non-ZI genes.
Maybe this mode is not relevant for scVI's general purpose and tasks though

jeff-regier:
This PR looks good so far! I wonder how scVI will do with this dataset.

scvi/dataset/synthetic.py
oscarclivio: OK we change that immediately, thanks!

scvi/dataset/synthetic.py
jeff-regier: It's probably fine to include it if it's relevant for your experiments. I just didn't realize that mixed mode was a realistic scenario.

Rather than having separate "mixed" and "zi" modes, how about using the ZIFA dropout model to create this synthetic data, and letting the user specify the rate parameters lambda? I think that could give use more realistic dropout. Almost none of the highly expressed genes will be dropped out then, and many of the infrequently expressed genes will be dropped out for some settings of lambda.

scvi/dataset/synthetic.py
```diff
@@ -42,3 +42,113 @@ def __init__(self, mu=4.0, theta=2.0, dropout=0.7, save_path='data/'):
...
+            weights[ind_first_gene_cluster:ind_last_high_gene_cluster] = weight_high
+            weights /= weights.sum()
+
+            exprs = np.random.poisson(lam_0 * weights, size=(n_batches, n_cells_cluster,
+                                                             n_genes_total))
+            data[:, cluster * n_cells_cluster:(cluster + 1) * n_cells_cluster, :] = exprs
+        # Apply dropout depending on the mode
+        if mode == "zi":
+            mask = np.random.binomial(n=1, p=1 - p_dropout,
+                                      size=(n_batches, batch_size, n_genes_total))
```
jeff-regier: Probably better to make the dropout probability depend on `exprs` here, using the ZIFA model --- I don't think highly expressed genes will ever get dropped out.

scvi/dataset/synthetic.py
```diff
@@ -42,3 +42,118 @@ def __init__(self, mu=4.0, theta=2.0, dropout=0.7, save_path='data/'):
...
+        return data * mask
+
+
+# class MixedSyntheticDatasetCorr(SyntheticDatasetCorr):
+#     def __init__(self, p_dropout=0.2, ratio_genes_zi=0.5, **kwargs):
+#         self.p_dropout = p_dropout
+#         self.ratio_genes_zi = ratio_genes_zi
+#         self.zi_genes_idx = None
+#         super(MixedSyntheticDatasetCorr, self).__init__(**kwargs)
+#
```
jeff-regier: Better not to commit commented out code: https://kentcdodds.com/blog/please-dont-commit-commented-out-code

PierreBoyeau: Sorry about that, we wanted to keep a trace of that piece of code since it was not committed before ... We should not have done that!
Thanks for the advice!

PierreBoyeau:
Everything seems fine now

tests/test_zeros_synth.py
```diff
@@ -0,0 +1,189 @@
...
+    tech_zeros = nb_data_zeros[is_technical_mask].sum()
+    bio_zeros = nb_data_zeros[~is_technical_mask].sum()
+    print("Prop of technical zeros :", tech_zeros)
+    print("Prop of biological zeros :", bio_zeros)
+    assert .2 <= tech_zeros / float(bio_zeros) <= 5.
+    assert tech_zeros >= 1000
+
+
+@pytest.mark.model_fit
+@pytest.mark.skipif(not is_available(), reason='Test requires GPU')
```
jeff-regier: Will these decorators stop the test from running on travis-ci? Ideally the test would run on travis-ci still, but train just for zero or one epochs, and the assertions wouldn't be checked.

jeff-regier: It's OK if the test runs on a CPU as long as a user is willing to wait awhile. It's enough that they use the --model-fit flag.

PierreBoyeau: Hey Jeff, ok  will change that way!

tests/test_zeros_synth.py
```diff
@@ -0,0 +1,187 @@
...
+    nb_data_zeros = nb_data.X == 0
+    tech_zeros = nb_data_zeros[is_technical_mask].sum()
+    bio_zeros = nb_data_zeros[~is_technical_mask].sum()
+    print("Prop of technical zeros :", tech_zeros)
+    print("Prop of biological zeros :", bio_zeros)
+    assert .2 <= tech_zeros / float(bio_zeros) <= 5.
+    assert tech_zeros >= 1000
+
+
+@pytest.mark.model_fit
```
jeff-regier: Please remove the `@pytest.mark.model_fit` decorator too. Instead, set `n_epoch = 1` if model fit wasn't specified (and `n_epoch = 150` otherwise.

tests/test_zeros_synth.py
```diff
@@ -0,0 +1,187 @@
...
+    axes[1, 0].set_title('Poisson Distribution Parameter GT')
+    plt.savefig(os.path.join(folder, 'params_comparison.png'))
+    plt.close()
+
+    # TODO: Decrease test tolerances
+    l1_poisson = np.abs(poisson_params - poisson_params_gt).mean()
+    print('Average Poisson L1 error: ', l1_poisson)
+    assert l1_poisson <= 0.75, \
+        'High Error on Poisson parameter inference'
+    l1_dropout = np.abs(p_dropout_infered_all - synth_data.p_dropout).mean()
```
jeff-regier: Only check these assertions if the user specified --model-fit

PierreBoyeau: Normally fixed by 76fc30c

tests/test_zeros_synth.py
PierreBoyeau: Normally fixed by 76fc30c

jeff-regier:
Looks great! Nice work Pierre and Oscar!

PR #312, opened 2019-05-03
[WIP] Differential Expression - Gamma

### LINKED ISSUE #309: Differential Expression
We should compare performing a DE test on the `rho` values to a test on the `w` values, which are the samples from the gamma. 

It looks like we can calculate in [closed form](https://math.stackexchange.com/questions/2426140/probability-of-a-gamma-r-v-greater-than-another) the probability that `w_a > w_b` for cells `a` and `b` and for the same gene `g` because they have the same rate (`theta` in the context of scVI)

PierreBoyeau:
Closes #309 

Quick implementation of DE in which the Gamma distribution is involved instead of rho.
Still WIP, as some functions are duplicated for now.
We also need synthetic datasets relative to DE as a whole to evaluate this work.

@adamgayoso


romain-lopez:
I wouldn't push this work before a full evaluation on simulated data. Let's work on that first.

PR #318, opened 2019-05-09
Gabriel/fix anndataset

### LINKED ISSUE #288: Loading a big AnnDataset produces a MemoryError
Attempting to load a large (300,000 cells by 30,000 genes) .h5ad file immediately causes a memory error.  This is induced by the following line in preprocess():

```python
data = ad.X.toarray()  # Sparse
```

Is there a way to keep the matrix sparse?  I can successfully work with this dataset using other tools, but scvi's attempt to make a dense matrix cannot succeed for a large dataset.

Environment:
-OS: Ubuntu 16.04
-Python: 3.6.8
-scvi: 0.3.0

This is the output:

Preprocessing dataset
```
MemoryError                               Traceback (most recent call last)
<ipython-input-3-138de299bfb8> in <module>
      1 # load
      2 dataset = AnnDataset('data/dataset.h5ad', 
----> 3                      save_path=path)

~/anaconda3/envs/pyro/lib/python3.6/site-packages/scvi/dataset/anndata.py in __init__(self, filename, save_path, url, new_n_genes, subset_genes)
     36         self.url = url
     37 
---> 38         data, gene_names = self.download_and_preprocess()
     39 
     40         super().__init__(*GeneExpressionDataset.get_attributes_from_matrix(data),

~/anaconda3/envs/pyro/lib/python3.6/site-packages/scvi/dataset/dataset.py in download_and_preprocess(self)
     61     def download_and_preprocess(self):
     62         self.download()
---> 63         return self.preprocess()
     64 
     65     def collate_fn(self, batch):

~/anaconda3/envs/pyro/lib/python3.6/site-packages/scvi/dataset/anndata.py in preprocess(self)
     52             data = ad.X.copy()  # Dense
     53         else:
---> 54             data = ad.X.toarray()  # Sparse
     55         select = data.sum(axis=1) > 0  # Take out cells that doesn't express any gene
     56         data = data[select, :]

~/anaconda3/envs/pyro/lib/python3.6/site-packages/scipy/sparse/compressed.py in toarray(self, order, out)
    960         if out is None and order is None:
    961             order = self._swap('cf')[0]
--> 962         out = self._process_toarray_args(order, out)
    963         if not (out.flags.c_contiguous or out.flags.f_contiguous):
    964             raise ValueError('Output array must be C or F contiguous')

~/anaconda3/envs/pyro/lib/python3.6/site-packages/scipy/sparse/base.py in _process_toarray_args(self, order, out)
   1185             return out
   1186         else:
-> 1187             return np.zeros(self.shape, dtype=self.dtype, order=order)
   1188 
   1189 

MemoryError: 
```

gabmis:
closes #288

### DIFF
--- a/scvi/dataset/anndata.py
+++ b/scvi/dataset/anndata.py
@@ -1,5 +1,7 @@
+import operator
 import os
 
+from functools import reduce
 from typing import List, Union
 
 import anndata
@@ -87,7 +89,7 @@ class AnnDataset(GeneExpressionDataset):
             data = ad.X.values
         if isinstance(ad.X, csr_matrix):
             # keep sparsity above 1 Gb in dense form
-            if ad.X.toarray().nbytes < 1e9:
+            if reduce(operator.mul, ad.X.shape) * ad.X.dtype.itemsize < 1e9:
                 data = ad.X.toarray()
             else:
                 data = ad.X.copy()
--- a/scvi/dataset/anndata.py
+++ b/scvi/dataset/anndata.py
@@ -86,8 +86,8 @@ class AnnDataset(GeneExpressionDataset):
         if isinstance(ad.X, pd.DataFrame):
             data = ad.X.values
         if isinstance(ad.X, csr_matrix):
-            # keep sparsity above 100 Mb
-            if ad.X.data.nbytes < 1e8:
+            # keep sparsity above 1 Gb in dense form
+            if ad.X.toarray().nbytes < 1e9:
                 data = ad.X.toarray()
             else:
                 data = ad.X.copy()
--- a/scvi/dataset/anndata.py
+++ b/scvi/dataset/anndata.py
@@ -88,16 +88,10 @@ class AnnDataset(GeneExpressionDataset):
             data = ad.X.values
         if isinstance(ad.X, csr_matrix):
             # keep sparsity above 1 Gb
-            if csr_matrix.data.nbytes < 1e9:
+            if ad.X.data.nbytes < 1e6:
                 data = ad.X.toarray()
             else:
                 data = ad.X.copy()
-
-        # warn if some last minute filtering is done
-        select = data.sum(axis=1) > 0  # Take out cells that doesn't express any gene
-        if len(select) != data.shape[0]:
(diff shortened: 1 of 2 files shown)


scvi/dataset/anndata.py
```diff
@@ -79,13 +80,17 @@ def extract_data_from_anndata(self, ad: anndata.AnnData):
...
-
-        select = data.sum(axis=1) > 0  # Take out cells that doesn't express any gene
-        data = data[select, :]
+            data = ad.X.copy()
+        if isinstance(ad.X, pd.DataFrame):
+            data = ad.X.values
+        if isinstance(ad.X, csr_matrix):
+            # keep sparsity above 100 Mb
+            if ad.X.data.nbytes < 1e8:
+                data = ad.X.toarray()
```
jeff-regier: How about deciding whether to keep sparsity based on the size the dense matrix would be? That seems most relevant. I'd say, switch to dense format always if the dense matrix would be less than 1 GB.

gabmis: haha obviously, I was tired..

scvi/dataset/anndata.py
```diff
@@ -79,13 +80,17 @@ def extract_data_from_anndata(self, ad: anndata.AnnData):
...
-            data = ad.X.toarray()  # Sparse
-
-        select = data.sum(axis=1) > 0  # Take out cells that doesn't express any gene
-        data = data[select, :]
+            data = ad.X.copy()
+        if isinstance(ad.X, pd.DataFrame):
+            data = ad.X.values
+        if isinstance(ad.X, csr_matrix):
+            # keep sparsity above 1 Gb in dense form
+            if ad.X.toarray().nbytes < 1e9:
```
jeff-regier: Does this call convert the sparse matrix to dense to find out how many bytes it would use if converted?

jeff-regier: Could you multiply the number of genes by the number of cells by 4 (for single precision floats) to estimate the size the dense matrix would be, without actually forming the dense matrix?

jeff-regier:
Looks good to me. Thanks Gabriel!

Issue #325, opened 2019-05-10
Large local datasets in hyperparameters autotuning


raevskymichail:
HI! 
First, thank you for developing the __hyperopt__ solution for hyperparameters optimization.
Unfortunately, I am faced with the following problem while applying `auto_tune_scvi_model()` on my local dataset.
The snippet I ran:
```python
import logging
logger = logging.getLogger("scvi.inference.autotune")
logger.setLevel(logging.WARNING)

# set high n_epochs since default behaviour has early stopping
n_epochs = 1000 if n_epochs_all is None else n_epochs_all
max_evals = 1000 if n_epochs_all is None else 5

best_trainer, trials = auto_tune_scvi_model(
    gene_dataset=GSE96769_dataset,
    parallel=True,
    exp_key="exp1",
    train_func_specific_kwargs={'n_epochs':n_epochs},
    max_evals=max_evals
)
```
Which returned an error:
```
  0%|          | 0/1000 [00:00<?, ?it/s][2019-05-10 09:27:24,368 - fmin Process - MainThread] ERROR - scvi.inference.autotune
Caught ('cannot serialize a bytes object larger than 4 GiB',) in _fmin_parallel, starting cleanup
Traceback (most recent call last):
  File "/home/mraevsky/miniconda3/lib/python3.7/site-packages/scvi/inference/autotune.py", line 140, in decorated
    return func(*args, **kwargs)
  File "/home/mraevsky/miniconda3/lib/python3.7/site-packages/scvi/inference/autotune.py", line 594, in _fmin_parallel
    show_progressbar=show_progressbar,
  File "/home/mraevsky/miniconda3/lib/python3.7/site-packages/hyperopt/fmin.py", line 388, in fmin
    show_progressbar=show_progressbar,
  File "/home/mraevsky/miniconda3/lib/python3.7/site-packages/hyperopt/base.py", line 639, in fmin
    show_progressbar=show_progressbar)
  File "/home/mraevsky/miniconda3/lib/python3.7/site-packages/hyperopt/fmin.py", line 405, in fmin
    show_progressbar=show_progressbar)
  File "/home/mraevsky/miniconda3/lib/python3.7/site-packages/hyperopt/fmin.py", line 126, in __init__
    msg = pickler.dumps(domain)
OverflowError: cannot serialize a bytes object larger than 4 GiB
Process fmin Process:
Traceback (most recent call last):
[...]

PR #330, opened 2019-05-11
autotune : make autotune objective function customizable

gabmis:
First step in fixing #325


jeff-regier:
Looks good.  I like that the changes so far are all in autotune.py

Issue #331, opened 2019-05-11
inconsistent code in posterior get_latent


adamgayoso:
It seems like this function could be cleaned up. Instead of this

https://github.com/YosefLab/scVI/blob/ebd36c1bd6a5dc5336a661404d7dc90e85cc334a/scvi/inference/posterior.py#L153

One could do something like

`latent += [self.model.sample_from_posterior_z(sample_batch, give_mean=True).cpu()]`

I believe this would make the code more readable.

jeff-regier:
I'm not sure I follow precisely what you're proposing...are you just talking about modifying just line 153? Or the whole function that contains line 153?

But no matter: I'd welcome any PR that makes the code more readable! It'll be easiest for me to verify that the change is beneficial if you propose it in a PR, where I can see precisely what lines we're revising.

PR #332, opened 2019-05-12
modify posterior get_latent to be more consistent

adamgayoso:
`model.sample_from_posterior_z` can already handle returning the posterior mean or a sample, so it doesn't make sense to handle this twice (once in the model, once in posterior)

resolves: #331


jeff-regier:
Nice! Thanks Adam.

PR #335, opened 2019-05-14
remove bioconda snake logo

jeff-regier:
I think it was the bioconda snake logo after all that was preventing the docs from building.


jeff-regier:
@romain-lopez Feel free to add the bioconda snake character back if you can figure out how to escape it. readthedocs was running the latex compiler on the documentation, which included the readme; latex was throwing errors when it got to the snake character.

Issue #342, opened 2019-05-18
One vs all DE bug?


adamgayoso:
https://github.com/YosefLab/scVI/blob/336aeb0b1566717aea436b6837c48bd162b4d307/scvi/inference/posterior.py#L295

It's hard for me to tell if this is a bug, in particular, I believe `sample_pairs` should be True, otherwise it's not necessary to report permuted Bayes factors in this function. As an aside, it would be great if someone added docstrings to all DE functions :)

jeff-regier:
^ @romain-lopez

romain-lopez:
sample_pairs=False means that you do not subsample the pairs. I did not write any of the code in this section and it is extremely hard for me to write a docstring. 

There are several open question to designing the API for the differential expression, notably:
- Do we subsample the cells / from the variational distribution. For efficiency, we may need the first and for accuracy we might need the second. how many times then ?
- Do we use importance sampling ? If yes, the code has to be substantially refactored. I partially did it in the scANVI paper repository, but there would be more work to put it in here and I am not sure yet we have a problem with our vanilla estimator.
- What kind of high-level test do we want to perform ? One-VS-All, One-VS-One, etc...
- How to perform DE with multiple batches, with scVI and scANVI ? A lot of code for this is on this repo and the scANVI repo, but there would be some refactorization to perform in order to make it readable. The multiple batches scenario is when the API really needs to be designed well. 
- Is the permutation test really informative or meaningful ? It was more of a sanity check when we wrote the scVI paper and I am not even sure it can be meaningfully used to assign a frequentist p-value for differential expression. If this is the case, might as well get rid of this code. 
- Do we want to reweight the rho values by the ZI weight for example ?

I was hoping we would get some partial answers from the work of Pierre and that would help us move forward on having a functional code for differential expression.

jeff-regier:
Thanks for the information about `sample_pairs`.

@PierreBoyeau It'd be great if you'd think about the issues Adam and Romain have raised. I'm not sure how many of the old DE functions you'll want to keep, but for those we do keep, it'd be good to docstrings.

PierreBoyeau:
Hey everyone,
For sure I will think about that. I can add docstrings but I think that refactoring right now could be a good idea.
@romain-lopez even if your scANVI implementation do not include all old DE functionalities, I think your new implementation is much clearer than before.
[...]

PR #349, opened 2019-05-21
[WIP] refac DE

PierreBoyeau:
What I add in mind to refactor #342 based on Romain's work on Importance Sampling
There is still a couple things I need to do so that Importance sampling works.

I think adding tests specifically for importance sampling might be interesting.

### DIFF
--- a/scvi/inference/posterior.py
+++ b/scvi/inference/posterior.py
@@ -174,10 +174,10 @@ class Posterior:
         """
         Output average over statistics in a symmetric way (a against b)
         forget the sets if permutation is True
-        :param vae: The generative vae and encoder network
-        :param data_loader: a data loader for a particular dataset
         :param M_sampling: number of samples
-        :return: A 1-d vector of statistics of size n_genes
+        :return: Tuple px_scales, all_labels where:
+            - px_scales: scales of shape (M_sampling, n_genes)
+            - all_labels: labels of shape (M_sampling, )
         """
         px_scales = []
         all_labels = []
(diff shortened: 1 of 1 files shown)


scvi/inference/posterior.py
```diff
@@ -224,62 +224,103 @@ def sample_scale_from_batch(self, n_samples, batchid=None, selection=None):
...
+            Remember that computing the Bayes Factor requires sampling
+            q(z_A | x_A) and q(z_B | x_B)
+
+        :param idx1: MASK of subpopulation cells 1
+        :param idx2: MASK of subpopulation cells 2
+        :param batchid1: batch ids of cells 1
+        :param batchid2: batch ids of cells 2
+        :param genes: Names of genes for which Bayes factors will be computed
+        :param n_samples: Number of times the posterior will be sampled for each pop
+        :param M_permutation: Number of times we will "mix" posterior samples in step 2
```
romain-lopez: can we write something more precise ? I'm open to changing the name M_permutation. Also, M_permutation should follow sample_pairs ? Ideally first sample_pairs and then M_permutation since the latter is useful only provided that the former is useful only if the latter is True. for example, could we call M_permutation n_pairs ?

scvi/inference/posterior.py
```diff
@@ -224,62 +224,103 @@ def sample_scale_from_batch(self, n_samples, batchid=None, selection=None):
...
+        :param all_stats: Returns either Bayes factors of complete info
```
romain-lopez: Typo and what is complete info ?

scvi/inference/posterior.py
```diff
@@ -224,62 +224,103 @@ def sample_scale_from_batch(self, n_samples, batchid=None, selection=None):
...
+        :param sample_pairs: Activates step 2 described above
```
romain-lopez: from my understanding, step 2 is always active.  Just you either subsample a certain number of pairs or you don't.

scvi/inference/posterior.py
```diff
@@ -224,62 +224,103 @@ def sample_scale_from_batch(self, n_samples, batchid=None, selection=None):
...
+        Computes gene specfic Bayes factors using masks idx1 and idx2
+
+        To that purpose we sample the Posterior in the following way:
+            1. The posterior is sampled n_samples times for each subpopulation
+            2. For computation efficiency (posterior sampling is quite expensive), instead of
+            comparing element-wise the obtained samples, we can permute posterior samples.
+            Remember that computing the Bayes Factor requires sampling
+            q(z_A | x_A) and q(z_B | x_B)
+
+        :param idx1: MASK of subpopulation cells 1
```
romain-lopez: mask = binary array of size n_cells in the posterior ? You need to expand a bit on that since it's unusual

scvi/inference/posterior.py
```diff
@@ -224,62 +224,103 @@ def sample_scale_from_batch(self, n_samples, batchid=None, selection=None):
...
+        :param idx2: MASK of subpopulation cells 2
+        :param batchid1: batch ids of cells 1
```
romain-lopez: We might have to expand a little bit on that. That means that each cell from batch 1 is going to be reimputed in all the different batches of the LIST batchid1. Chenling can explain to you why this is useful in multiple batches studies. This is not the batch_id of every cell in the list idx1.

scvi/inference/posterior.py
```diff
@@ -224,62 +224,103 @@ def sample_scale_from_batch(self, n_samples, batchid=None, selection=None):
...
+            return pd.Series(data=bayes1, index=self.gene_dataset.gene_names)
 
     @torch.no_grad()
     def one_vs_all_degenes(self, subset=None, cell_labels=None, min_cells=10,
-                           n_samples=None, M_permutation=None, output_file=False,
+                           n_samples=1000, M_permutation=None, output_file=False,
                            save_dir='./', filename='one2all'):
+        """
+
+        :param subset: MASK: subset of cells you are interested in
```
romain-lopez: Define MASK more precisely

scvi/inference/posterior.py
```diff
@@ -224,62 +224,103 @@ def sample_scale_from_batch(self, n_samples, batchid=None, selection=None):
...
+        :param cell_labels: Labels of cells
```
romain-lopez: idem, is this a list, of what ?

scvi/inference/posterior.py
```diff
@@ -224,62 +224,103 @@ def sample_scale_from_batch(self, n_samples, batchid=None, selection=None):
...
+        :param min_cells: Ceil number of cells used to compute Bayes Factors
+        :param n_samples: see `differential_expression_score`
```
romain-lopez: You might have to copy over ? Is it good practice to see "go to xxx" ?

romain-lopez: You might have to copy over ? Is it good practice to see "go to xxx" ?

scvi/inference/posterior.py
```diff
@@ -643,47 +705,149 @@ def entropy(hist_data):
...
         # then constitutes the pairs
-        first_set = samples[u]
-        second_set = samples[v]
+        first_samples = samples[u]
+        second_samples = samples[v]
+        first_weights = weights[u]
+        second_weights = weights[v]
+        to_sum = first_weights[:, np.newaxis] * second_weights[:, np.newaxis] * (
+                first_samples >= second_samples)
+        incomplete_weights = first_weights * second_weights
```
romain-lopez: I had some problem in practise with the imcomplete weights which would sometimes sum to zero. So it seems like using importance sampling means subselecting pairs of cells and using all the importance samples to compute the Bayes factor. I tried here to subsample from both the cells and the importances samples weight but it is really annoying in practise.

romain-lopez:
- If you want to use importance sampling, we might have to push some changes in the VAE code to support for importance weight calculations (see scANVI repo)
- I also propose in the scANVI repo a much nicer way to reimputing gene expression on another batch than creating another data loader (which sounds quite extreme and dangerous to me). It consists in modifying the inference function in VAEs.
From here 
https://github.com/YosefLab/scVI/blob/master/scvi/inference/posterior.py#L205
to there 
https://github.com/chenlingantelope/HarmonizationSCANVI/blob/master/scvi/models/vae.py#L142

scvi/inference/posterior.py
jeff-regier: Better to copy

PR #352, opened 2019-05-21
trainer : reduce lr on plateau

gabmis:
provides a "reduce learning rate on plateau" behaviour which can be set through the `EarlyStopping` attribute of the `Trainer`.


jeff-regier:
Great!

Issue #317, opened 2019-05-07
Deprecated diff exp function?


adamgayoso:
It's not clear to me what the purpose of this function is. It doesn't seem to be used for anything else and it is a bit confusing because the doc string is incorrect and the output is not sequential.

https://github.com/YosefLab/scVI/blob/1584b9ef3c21b77d3feb88a6a4727a9f389fd963/scvi/inference/posterior.py#L172

jeff-regier:
It doesn't look like it's used in any of our notebooks either -- not even the one that recreates the results of the original scVI paper.

@romain-lopez Does this function look at all familiar to you?

If this function isn't doing anything useful any more, it'd be great to delete it.

romain-lopez:
yes, we use a modified version of this function in the scANVI paper but I am not sure that one is absolutely necessary. In particular, we are probably going to need to design a more flexible API with people are going to want to use scVI and scANVI for DE with respect to multiple batches.

jeff-regier:
Maybe it would be best if @PierreBoyeau removed (or revised) this function as part of his work on DE?

In general I think we want all the code in the `scvi` directory to get executed by running the notebooks. But `differential_expression_stats` isn't called by any notebook.

romain-lopez:
yes, this is what I was thinking about !

PierreBoyeau:
Sure it makes sense!

romain-lopez:
to change the docstring

romain-lopez:
fixed by #354

PR #354, opened 2019-05-22
Pierre/doc posterior

### LINKED ISSUE #342: One vs all DE bug?

PierreBoyeau:
Fixes #342

### DIFF
--- a/scvi/inference/posterior.py
+++ b/scvi/inference/posterior.py
@@ -18,6 +18,8 @@ from sklearn.neighbors import NearestNeighbors, KNeighborsRegressor
 from sklearn.utils.linear_assignment_ import linear_assignment
 from torch.utils.data import DataLoader
 from torch.utils.data.sampler import SequentialSampler, SubsetRandomSampler, RandomSampler
+from typing import List, Optional, Union
+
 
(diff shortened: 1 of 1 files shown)


gabmis:
Thanks Pierre! While you're at it it would be nice to have parameter annotations using python's `typing`module.  Do you have time to add that before we merge?

@jeff-regier @romain-lopez Is it ok if with review and merge simple PRs ourselves?

PierreBoyeau:
@gabmis Sure thingk!

jeff-regier:
@gabmis Yes, it's fine if whomever you assign to review your PR does the merge. Please use "Squash and merge" so that the whole PR is one commit in master.

PierreBoyeau:
Issue with blame history. I will do a new PR and promise never to use Black again

romain-lopez:
also fixes #317

PR #356, opened 2019-05-30
fix nans

jeff-regier:
This PR should fix the nan issue reported by Gabriel on the brain_large dataset. We were letting variance of z get arbitrarily small before; this PR ensures it's always greater than 1e-4.


scvi/dataset/brain_large.py
```diff
@@ -51,7 +51,7 @@ def preprocess(self):
         print("Preprocessing Brain Large data")
 
         filtered_matrix_h5 = os.path.join(self.save_path, self.download_name)
-        with h5py.File(filtered_matrix_h5) as f:
+        with h5py.File(filtered_matrix_h5, 'r') as f:
```
jeff-regier: This change lets multiple processes read the dataset concurrently

PR #358, opened 2019-06-06
posterior : add elbo metric and rename ll to reconstruction_error

### LINKED ISSUE #359: trainer : we should check that early_stopping_metric is in metrics_to_monitor

### LINKED ISSUE #361: trainer : metrics_to_monitor are not monitored, only to_monitor attributes of the posteriors

gabmis:
fixes #361 
fixes #359


scvi/inference/posterior.py
```diff
@@ -31,7 +32,7 @@ def __iter__(self):
 
 class Posterior:
     r"""The functional data unit. A `Posterior` instance is instanciated with a model and a gene_dataset, and
-    as well as additional arguments that for Pytorch's `DataLoader`. A subset of indices can be specified, for
+    as well as additional arguments that for Pytorch"s `DataLoader`. A subset of indices can be specified, for
```
jeff-regier: Did you mean to replace this apostrophe with a quote mark?

scvi/inference/posterior.py
```diff
@@ -46,7 +47,7 @@ class Posterior:
 
     Examples:
 
-    Let's instanciate a `trainer`, with a gene_dataset and a model
+    Let"s instanciate a `trainer`, with a gene_dataset and a model
```
jeff-regier: Here too

scvi/inference/fish.py
```diff
@@ -5,17 +5,24 @@
...
         if verbose:
-            print("LL Fish: %.4f" % ll)
-        return ll
+            print("ELBO Fish: %.4f" % elbo)
+        return elbo
+
+    @torch.no_grad()
+    def reconstruction_error(self, verbose=False):
+        reconstruction_error = compute_reconstruction_error(self.model, self, mode="smFISH")
+        if verbose:
```
jeff-regier: Are these `verbose`  flags useful? There's something to be said for just adding in breakpoints for debugging. But I'm just asking.

jeff-regier: If they aren't useful, it'd be nice to just directly call `compute_reconstruction_error` rather than adding another method that does almost nothing except call `compute_reconstruction_error`.

scvi/inference/fish.py
```diff
@@ -65,8 +72,8 @@ def __init__(self, model, gene_dataset_seq, gene_dataset_fish, train_size=0.8, t
         self.train_seq, self.test_seq = self.train_test(self.model, gene_dataset_seq, train_size, test_size, seed)
         self.train_fish, self.test_fish = self.train_test(self.model, gene_dataset_fish,
                                                           train_size, test_size, seed, FishPosterior)
-        self.test_seq.to_monitor = ['ll']
-        self.test_fish.to_monitor = ['ll']
+        self.test_seq.to_monitor = ['reconstruction_error']
+        self.test_fish.to_monitor = ['reconstruction_error']
```
jeff-regier: Do you want double quotes here? I prefer double quotes I guess but I don't mind a mix either. But other places it seems like you're switching to double.

scvi/models/log_likelihood.py
```diff
@@ -7,12 +7,32 @@
...
-def compute_log_likelihood(vae, posterior, **kwargs):
-    """ Computes log p(x/z), which is the reconstruction error .
-        Differs from the marginal log likelihood, but still gives good
-        insights on the modeling of the data, and is fast to compute
+def compute_elbo(vae, posterior, **kwargs):
+    """ Computes the ELBO.
+
+    The ELBO  is the reconstruction error + the KL divergences
+    between the variational distributions and the priors.
+    Differs from the marginal log likelihood, but still gives good
```
jeff-regier: You might say that the ELBO is a lower bound on the marginal log likelihood plus a term that is constant with respect to the variational distribution.

scvi/inference/posterior.py
gabmis: Misuse of ctrl-replace I guess

scvi/inference/fish.py
gabmis: > Are these `verbose` flags useful? There's something to be said for just adding in breakpoints for debugging. But I'm just asking.

Not really, especially especially when we'll have completely switched to logging



> If they aren't useful, it'd be nice to just directly call `compute_reconstruction_error` rather than adding another method that does almost nothing except call `compute_reconstruction_error`.

I did not want to break the calls that had `verbose` param in them, especially in the notebooks since it would more or less require a full re-run to have to corresponding output.

scvi/inference/fish.py
gabmis: I care for consistency and prefer double quotes. As I was modifying posterior.py I thought i'd harmonize to double quotes. I did not do it here to avoid breaking potential consistency.

scvi/models/log_likelihood.py
gabmis: Thanks, i'll put that in.

scvi/inference/trainer.py
```diff
@@ -42,8 +42,11 @@ class Trainer:
     default_metrics_to_monitor = []
 
     def __init__(self, model, gene_dataset, use_cuda=True, metrics_to_monitor=None, benchmark=False,
-                 verbose=False, frequency=None, weight_decay=1e-6, early_stopping_kwargs=dict(),
-                 data_loader_kwargs=dict(), show_progbar=True):
+                 verbose=False, frequency=None, weight_decay=0, early_stopping_kwargs=None,
```
jeff-regier: It looks like you're changing `weight_decay` here from 1e-6 to 0. It should be the former I think, right?

gabmis:
@jeff-regier Thank you for your review, this PR is done on my part, feel free to merge.

PR #367, opened 2019-06-13
quick fixes

### LINKED ISSUE #300: switch from print to logging
It'd be good to switch the remaining print statements in scVI to calls to `logging`. We have a mix of calls to logging and print currently.

https://docs.python-guide.org/writing/logging/

### LINKED ISSUE #362: weight_decay is disabled
https://github.com/YosefLab/scVI/blob/1cc8698aaa357f4a8dd12d2f49fbd21d8b29ef90/scvi/inference/trainer.py#L112

Why?

### LINKED ISSUE #363: unused variable in vae.py
Is this variable useless ?
https://github.com/YosefLab/scVI/blob/master/scvi/models/vae.py#L59

jeff-regier:
There wasn't much to these three issues, so I quickly addressed them in this PR. Closing them would keep help us keep the number of open issues under control.

fixes #363
fixes #362
fixes #300

### DIFF
--- a/scvi/__init__.py
+++ b/scvi/__init__.py
@@ -11,3 +11,4 @@ import logging
 from logging import NullHandler
 
 logging.getLogger(__name__).addHandler(NullHandler())
+logging.basicConfig(level=logging.INFO)
--- a/scvi/dataset/anndata.py
+++ b/scvi/dataset/anndata.py
@@ -1,5 +1,6 @@
 import operator
 import os
+import logging
 
 from functools import reduce
 from typing import List, Union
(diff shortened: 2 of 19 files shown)


jeff-regier:
1. Deleted now, I think .. didn't see those before. For many of these FIXME comments in the code, it'd be better if we opened github issues for them.
2. I changed the default logging level to INFO, so that `logging.info(...)` should function by default pretty much like `print(...)`.

gabmis:
@jeff-regier 

I would rather use the formulation:
```
logger = logging.get_logger(__name__)

logger.info(...)
```

This allows for a per-module logging behaviour. 

Also, what did you do with the verbose parameter of e.g the `Trainer` class?

jeff-regier:
I opened a new issue (#368) for the `verbose` flags, and for per-module logging.

jeff-regier:
Yeah, those `verbose` flags need to go. I agree, per-module logging can help with that.

jeff-regier:
@romain-lopez OK to merge as is? This PR doesn't adequately deal with logging but I opened a new issue for that.

romain-lopez:
Sounds good to me

PR #374, opened 2019-06-19
[WIP] Refactor GeneExpressionDataset

### LINKED ISSUE #270: library local_var and local_mean not calculated by batch
In the ` __init__` function of LoomDataset, it calls `GeneExpressionDataset.get_attributes_from_matrix` to calculate the local_means and local_vars of library size. The `preprocess` function takes batchID from loom file and stores it in batch_indices. However, it seems like `get_attributes_from_matrix` calculates the local_means and local_vars without grouping by the batch labels. 
``` python
    @staticmethod
    def library_size(X):
        log_counts = np.log(X.sum(axis=1))
        local_mean = (np.mean(log_counts) * np.ones((X.shape[0], 1))).astype(np.float32)
        local_var = (np.var(log_counts) * np.ones((X.shape[0], 1))).astype(np.float32)
        return local_mean, local_var

    @staticmethod
    def get_attributes_from_matrix(X, batch_indices=0, labels=None):
        ne_cells = X.sum(axis=1) > 0
        to_keep = np.where(ne_cells)
        if not ne_cells.all():
            X = X[to_keep]
            removed_idx = np.where(~ne_cells)[0]
            print("Cells with zero expression in all genes considered were removed, the indices of the removed cells "
                  "in the expression matrix were:")
            print(removed_idx)
        local_mean, local_var = GeneExpressionDataset.library_size(X)
        batch_indices = batch_indices * np.ones((X.shape[0], 1)) if type(batch_indices) is int \
            else batch_indices[to_keep]
        labels = labels[to_keep].reshape(-1, 1) if labels is not None else np.zeros_like(batch_indices)
        return X, local_mean, local_var, batch_indices, labels
```

Currently, I just import each dataset separately and use the `concat_datasets` function, which calculates the local_var and local_mean by batch. 

```
local_means = np.concatenate([gene_dataset.local_means for gene_dataset in gene_datasets])
local_vars = np.concatenate([gene_dataset.local_vars for gene_dataset in gene_datasets])
```

### LINKED ISSUE #337: refactor GeneExpressionDataset
`GeneExpressionDataset` is getting pretty messy. Currently, the constructor takes the already loaded data as arguments:
https://github.com/YosefLab/scVI/blob/221cf0fe3b392f0b592aa8040dce4870a0dd908b/scvi/dataset/dataset.py#L23-L24

It'd be preferable for the constructor to take no (or almost no) arguments. Instead, we'd add a new method like this:
```python
    def load(self): ...
```
Notice that the `load(...)` method isn't called with `local_means` and `local_vars`. Instead, it computes `local_means` and `local_vars` by calling `get_attributes_from_matrix(...)` itself. 

The `load(...)` method should also call `download_and_preprocess()` near its beginning.

The method `get_attributes_from_matrix` doesn't need to be called from anywhere besides the `load(...)` method. Currently all the subclasses of `GeneExpressionDataset` call this method before calling `GeneExpressionDataset.__init__()`. But that is duplicated code -- an anti-pattern. And it's confusing because the child class is using a static method of the parent class to generate the data structures it needs in order to instantiate the parent class.

In addition to keeping our codebase readable, adding this `load(...)` method to `GeneExpressionDataset` should help us with issue #325, and perhaps with #270.

### LINKED ISSUE #383: brain large : bug in gene subset selection
https://github.com/YosefLab/scVI/blob/0830cd625e9de613074db14925905a653e778be5/scvi/dataset/brain_large.py#L67-L73

We are missing a call to transpose when selecting the most variable genes. Thus, we are effectively selecting the most variable cells and using their index to subselect genes.

### LINKED ISSUE #389: Numpy version is too old
Hey @jeff-regier , you limited Numpy version in #209 (https://github.com/YosefLab/scVI/blob/66cc63ead0d1a219f8d0ba1018d6f610f2b91679/setup.py#L13) but there are useful new functionalities in 1.15 and 1.16
We should require at least numpy 1.16 in `setup.py`

### LINKED ISSUE #391: can we remove the run_benchmarks.py script? it seems rather useless
@romain-lopez @jeff-regier

### LINKED ISSUE #392: can we remove the scmfish notebook? ultimately replaced by gimVI notebook
@romain-lopez

gabmis:
fixes #337 
fixes #270 
fixes #383
closes #389 
closes #391
closes #392
prepares fix for #325

- [x] Refactor GeneExpressionDataset
- [x] Implement DownloadableDataset abstraction
- [x] Propagate new api to sub-classes
- [x] Add CellMeasurements logic, for e.g protein data.
- [x] Propagate CellMeasurement logic to datasets containing such data, e.g 10X, CiteSeq
- [x] Harmonize documentation


gabmis:
@jeff-regier @romain-lopez This is still very WIP, @ANazaret is going to help me with it and I made the PR for our collaboration to be easily traceable. We'll also be able to raise discussion points with you from here.

scvi/dataset/dataset.py
```diff
@@ -21,52 +21,246 @@
...
+
     @property
     def X(self):
         return self._X
 
     @X.setter
-    def X(self, X):
+    def X(self, X: Union[np.ndarray, sp_sparse.csr_matrix]):
+        """Recomputes the library size"""
         self._X = X
```
ANazaret: Maybe add logging.INFO when performing costly operations in setter

scvi/dataset/dataset.py
```diff
@@ -21,52 +21,246 @@
...
+        """Sets and registers and attribute mapping, e.g labels to named cell_types."""
+        if not len(np.unique(getattr(self, source_attribute_name))) == len(
+            mapping_values
+        ):
+            raise ValueError("Number of labels and cell types mismatch")
+        self.attribute_mappings[source_attribute_name].append(mapping_name)
+        if mapping_values:
+            setattr(self, mapping_name, mapping_values)
+
+    def library_size_batch(self):
```
ANazaret: Change name, use a verb like `compute_library_size_per_batch`

scvi/dataset/dataset.py
```diff
@@ -21,52 +21,246 @@
...
+        """Computes the library size per batch."""
```
ANazaret: With a proper name we don't even need this comment ;)

scvi/dataset/dataset.py
```diff
@@ -421,17 +615,51 @@ def _filter_genes(gene_dataset, gene_names_ref, on='gene_names'):
...
+    if not len(mapping_to) == len(mapping_from):
+        raise ValueError("Length mismatch between mapping_to and mapping_from.")
 
     new_categories = np.copy(original_categories)
     for idx_from, idx_to in zip(mapping_from, mapping_to):
         new_categories[original_categories == idx_from] = idx_to
     return new_categories.astype(int), n_categories
+
+
+def library_size(
```
ANazaret: Same: use a verb

adamgayoso:
Just want to add that there are three datasets that we can currently load 
                       'pbmc_1k_protein_v3',
                       'pbmc_10k_protein_v3',
                       'malt_10k_protein_v3',
That have protein counts appended as columns, which would lead to weird scVI results. Once we get a better sense of the new API, I can add code to handle this case.

ANazaret:
@gabmis , I don't think it is a good idea to hide the normalization of the dataset using a property.
It would be better to have a method `compute_normalized_counts` and eventually an error if the user tries to access `norm_X` raising "norm_X must be computed before, use the `compute_normalized_counts` method".

gabmis:
> @gabmis , I don't think it is a good idea to hide the normalization of the dataset using a property.
> It would be better to have a method `compute_normalized_counts` and eventually an error if the user tries to access `norm_X` raising "norm_X must be computed before, use the `compute_normalized_counts` method".

I agree, we could have `compute_...` methods for both `norm_X` and `corrupted_X`. This would be more transparent.

scvi/dataset/dataset.py
```diff
@@ -3,275 +3,888 @@
...
-        self.update_genes(subset_genes)
 
-    def filter_genes(self, gene_names_ref, on='gene_names'):
+        if subset_genes is None:
+            logger.info(
+                "Not subsampling. No parameter given".format(new_n_genes=new_n_genes)
+            )
+            return
+
+        self.update_genes(np.array(subset_genes))
```
jeff-regier: Would it be clearer to construct and return a new `GeneExpressionDataset` rather than to mutate the current `GeneExpressionDataset`?

jeff-regier: And then delete the `update_cells` method?

jeff-regier: But just an idea...feel free not to incorporate it. This PR is already really impressive.

ANazaret: Actually all the methods in GeneExpressionDataset are inplace:
- It was the former behavior
- With huge dataset we cannot duplicate the data for each update

jeff-regier: Glad you changed that! Formerly numpy was called in a way that made a copy of the count matrix -- not very space efficient.

scvi/dataset/dataset.py
```diff
@@ -1,431 +1,1125 @@
...
+            categorical=True,
+        )
+
+        self.compute_library_size_batch()
 
         if gene_names is not None:
-            assert self.nb_genes == len(gene_names)
-            self.gene_names = np.array(gene_names, dtype=np.str)
+            self.initialize_gene_attribute(
+                "gene_names", np.char.lower(np.asarray(gene_names, dtype=np.str))
```
ANazaret: Uppercase is standard gene notation, rather than lowercase. @romain-lopez

scvi/dataset/dataset.py
```diff
@@ -241,28 +241,29 @@ def populate_from_datasets(
...
-        :param on: attribute to select gene intersection
-        :param sharing_instructions_dict: Instructions on how to share cell-wise attributes between datasets.
-            Keys are the attribute name and values are either:
-                * "offset": to add an offset corresponding to the number of categories already existing.
-                e.g for batch_indices, if the first dataset has batches 0 and 1,
-                in the merged dataset, the second dataset's batch indices will start at 2.
-                * "concatenate": concatenate the attribute, no changes applied.
+        :param gene_datasets_list: ``GeneExpressionDataset`` objects to be merged.
+        :param shared_labels: whether to share labels through ``cell_types`` mapping or not.
+        :param mapping_reference_for_sharing: Instructions on how to share cell-wise attributes between datasets.
```
ANazaret: I don't really understand the docstring for `mapping_reference_for_sharing`.
You don't have `offset` or `concatenate` anymore?

scvi/dataset/dataset.py
```diff
@@ -241,28 +241,29 @@ def populate_from_datasets(
...
             )
 
-        # set default sharing behaviour
-        if sharing_intstructions_dict is None:
-            sharing_intstructions_dict = {"batch_indices": "offset"}
+        # set default sharing behaviour for batch_indices and labels
+        if mapping_reference_for_sharing is None:
+            mapping_reference_for_sharing = {}
+        if shared_labels:
+            mapping_reference_for_sharing = {"labels": "cell_types"}
```
ANazaret: So if `shared_labels` is True it will override `mapping_reference_for_sharing`? 
I guess `mapping_reference_for_sharing` needs more documentation

scvi/dataset/dataset.py
```diff
@@ -241,28 +241,29 @@ def populate_from_datasets(
...
+        mapping_reference_for_sharing.update({"batch_indices": None})
```
ANazaret: You typed the dictionary as `Dict[str, str]`. Here why do you have None?

scvi/dataset/dataset.py
```diff
@@ -334,10 +335,46 @@ def populate_from_datasets(
...
+                        if is_categorical
+                        else local_attribute_values
+                    )
+                    attribute_values.append(new_values)
+                    offset += len(np.unique(local_attribute_values))
+                    for mapping_name in mapping_names_to_keep:
+                        mappings[mapping_name].extend(
+                            getattr(gene_dataset, mapping_name)
+                        )
+            # for categorical only, contenate shared mapping
```
ANazaret: *concatenate?

scvi/dataset/dataset.py
ANazaret: Ok I got it, you should indicate what happens if the mapping is not provided, here we only have "If provided the mapping is merged ..." . Also i don't understand what is the meaning of "merged" here.

scvi/dataset/dataset.py
```diff
@@ -241,28 +241,29 @@ def populate_from_datasets(
...
+        :param gene_datasets_list: ``GeneExpressionDataset`` objects to be merged.
+        :param shared_labels: whether to share labels through ``cell_types`` mapping or not.
+        :param mapping_reference_for_sharing: Instructions on how to share cell-wise attributes between datasets.
+            Keys are the attribute name and values are registered mapped attribute.
+            If provided the mapping is merged across all datasets and then the attribute is
+            remapped using index backtracking between the old and merged mapping.
         """
         # sanity check
-        if not all([hasattr(gene_dataset, on) for gene_dataset in gene_datasets_list]):
+        if not all([hasattr(gene_dataset, "gene_names") for gene_dataset in gene_datasets_list]):
```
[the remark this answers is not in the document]
gabmis: agreed

scvi/dataset/dataset.py
gabmis: typo, it should be ``.update({"labels": "cell_types"})``

scvi/dataset/dataset.py
gabmis: typo, we don't need this line as the call is ``mapping_reference_for_sharing.get(.., default=None)``

scvi/dataset/dataset.py
```diff
@@ -241,28 +241,29 @@ def populate_from_datasets(
...
-            sharing_intstructions_dict = {"batch_indices": "offset"}
+        # set default sharing behaviour for batch_indices and labels
+        if mapping_reference_for_sharing is None:
+            mapping_reference_for_sharing = {}
+        if shared_labels:
+            mapping_reference_for_sharing = {"labels": "cell_types"}
+        mapping_reference_for_sharing.update({"batch_indices": None})
 
-        # get insterection based on gene attribute `on` and get attribute intersection
+        # get insterection based on gene_names and get attribute intersection
```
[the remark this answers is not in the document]
gabmis: we do need a comment to explain what gene attributes we keep, but I'm not sure what the desired behaviour should be, either keep attributes that are in all datasets and use the values from the first one, or take all the gene attributes from all the datasets, and keep the version that comes first (in the order given by the dataset list). latter sounds better, what do you think?

scvi/dataset/dataset.py
gabmis: indeed

scvi/dataset/dataset.py
gabmis: no i changed that. edited the docstring. tell me if its clearer

scvi/dataset/anndataset.py
```diff
@@ -0,0 +1,141 @@
+import logging
```
jeff-regier: Would using `git rename anndata.py anndataset.py` maintain the git history of the file even though name has changed?

scvi/dataset/dataset.py
ANazaret: Good

scvi/dataset/dataset.py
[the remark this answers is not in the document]
ANazaret: Yes we can document a bit but I would prefer to explain the behavior in the function docstring, not in the middle of the code. For the comments I think this is enough:
```
# Keep the common genes between all datasets
        genes_to_keep = set.intersection(
            *[set(gene_dataset.gene_names) for gene_dataset in gene_datasets_list]
        )
# Keep the common cell attributes between all datasets
        cell_attributes_to_keep = set.intersection(
            *[
                set(gene_dataset.cell_attribute_names)
                for gene_dataset in gene_datasets_list
            ]
        )

# Keep all gene attributes. If there are multiple instances of the same attribute between
# datasets, the attribute from the first dataset (in the input list order) is kept.
```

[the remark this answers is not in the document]
ANazaret: I've written a test for the behavior about gene attributes and the code is not working.
The function doesn't keep the attribute of the first dataset but the one from the last dataset.

scvi/dataset/dataset.py
ANazaret: I still don't understand the usage of "merged" here `If provided the mapping is merged across all datasets`. The docstring is supposed to explain how the merging is done and to use "the mapping is merged" confuses me a bit.
For instance if I have: 
#### Dataset 1
`cell_types = ["T cells", "B cells"]` and `labels = [0,1,0,0,1]`
#### Dataset 2
`cell_types = ["B cells - 1", "B cells - 2", "T cells"]` and `labels = [1,2,0,0,1]`
####
Then what merging means? To have `labels = [0,1,0,0,1] + [1,2,0,0,1]`? And what about `cell_types`

Thanks!

scvi/dataset/anndataset.py
ANazaret: Do you think it could be a GitHub glitch? Because the PyCharm git blame still has the git history in the new `anndataset.py`. I am not familiar with that.

scvi/dataset/dataset.py
```diff
@@ -1,431 +1,1169 @@
...
+            np.concatenate(labels_per_batch).astype(np.int64)
+            if labels_per_batch is not None
+            else None
+        )
+        X = X.reshape(-1, X.shape[2])
+        self.populate_from_data(
+            X=X, batch_indices=batch_indices, labels=labels, gene_names=gene_names
+        )
+
+    def populate_from_per_batch_list(
```
ANazaret: I can't find any usage of this function in the codebase (neither for its previous implementation in current master).
In addition, I don't really see a dataset which can have a shape `(n_batches, n_cells, n_genes)` as different batches usually mean different cells. 
We should remove it.

scvi/dataset/dataset.py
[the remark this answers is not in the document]
gabmis: Missing [::-1] when iterating through datasets, my bad

scvi/dataset/dataset.py
gabmis: In essence, you take a mapping e.g ``cell_types`` you merge all of those into a set of aggregated mapping ``["T cells", "B cells", "B cells - 1", "B cells - 2"]`` then remap each datasets attribute, e.g ``labels``, using the previous and aggregated mapping.

scvi/dataset/dataset.py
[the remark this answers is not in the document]
ANazaret: Not much of a difference here but the built-in `reversed` is better than slicing (which returns a copy of the data whereas `reversed` just return an iterator).

scvi/dataset/dataset.py
ANazaret: Done

ANazaret:
Hey @adamgayoso, can you check this commit 93731ee91a761be148030e043597724951e67fdf about Paired Measurements? Is the API fine for you?
It can replace #382 
Thanks!

gabmis:
@ANazaret Still need to handle the unitest incompatiblity with pytest fixtures

scvi/dataset/dataset10X.py
```diff
@@ -1,201 +1,263 @@
...
+    "2.1.0": "http://cf.10xgenomics.com/samples/cell-exp/{}/{}/{}_{}_gene_bc_matrices.tar.gz",
+    "3.0.0": "http://cf.10xgenomics.com/samples/cell-exp/{}/{}/{}_{}_feature_bc_matrix.tar.gz",
 }
-available_specification = ['filtered', 'raw']
+available_specification = ["filtered", "raw"]
 
 
-class Dataset10X(GeneExpressionDataset):
-    r""" Loads a file from `10x`_ website.
+class Dataset10X(DownloadableDataset):
```
adamgayoso: There's a minor issue here where the datasets with "protein" in the name contain columns that are proteins and not genes. We can parse the features because they indicate if a column is a "Gene" or "Antibody Capture". If a gene has the word "total" in it and is "Antibody Capture" we should store those as another cell measurement. Else if it has "Antibody Capture" and something other than "total" it should be stored as a miscellaneous cell measurement.

adamgayoso: We can probably also add a few lines to remove genes that are all 0.

scvi/dataset/loom.py
```diff
@@ -1,86 +1,110 @@
...
-            batch_indices = np.reshape(batch_indices, (batch_indices.shape[0], 1))[select]
-
-        if 'ClusterID' in ds.ca:
-            labels = np.array(ds.ca['ClusterID'])
+        ds = loompy.connect(os.path.join(self.save_path, self.filenames[0]))
+        select = (
+            ds[:, :].sum(axis=0) > 0
+        )  # Take out cells that doesn't express any gene
+
+        if "Gene" in ds.ra:
```
adamgayoso: Are these keys standards? Seems like we can add a dictionary as an argument with the key structure of the loom file.

scvi/dataset/csv.py
```diff
@@ -1,100 +1,137 @@
...
-                 batch_ids_file=None):
-        self.download_name = filename  # The given csv file is
-        self.save_path = save_path
-        self.url = url
+    def __init__(
+        self,
+        filename: str,
+        save_path: str = "data/",
+        url: str = None,
+        new_n_genes: int = 600,
```
adamgayoso: Is there a reason for this default? Seems like we are implying 600 is the correct number of genes to use.

scvi/dataset/dataset10X.py
romain-lopez: "We can probably also add a few lines to remove genes that are all 0." 
Let's not do that, it's not user friendly to get your dataset altered. The behavior with the cell removed is already not perfect

scvi/dataset/loom.py
```diff
@@ -1,86 +1,110 @@
...
-
-        if 'BatchID' in ds.ca:
-            batch_indices = ds.ca['BatchID']
-            batch_indices = np.reshape(batch_indices, (batch_indices.shape[0], 1))[select]
-
-        if 'ClusterID' in ds.ca:
-            labels = np.array(ds.ca['ClusterID'])
+        ds = loompy.connect(os.path.join(self.save_path, self.filenames[0]))
+        select = (
+            ds[:, :].sum(axis=0) > 0
```
adamgayoso: Should this be here if we have a filter cells by count function?

scvi/dataset/dataset10X.py
adamgayoso: I think we need to address this somehow though. A lot of the datasets from 10x have thousands of empty columns.

romain-lopez: these should not be selected when you apply gene filtering right ?

adamgayoso: It looks like if a column is all zeros, the standardscaler gives it a scale of 1. This could be problematic.

scvi/dataset/csv.py
ANazaret: The only reason is that it is the current default in master: https://github.com/YosefLab/scVI/blob/66cc63ead0d1a219f8d0ba1018d6f610f2b91679/scvi/dataset/csv.py#L37. We don't want to change the behavior of the class because users might already use it.

scvi/dataset/loom.py
gabmis: we can use the api

scvi/dataset/csv.py
gabmis: It would definitely mean adding a warning for users already using it.

scvi/dataset/loom.py
ANazaret: I don't like it either but it was the original implementation. @romain-lopez do you think it is worth it to remove it and so changing the returned dataset? and so let the user do the filtering?

scvi/dataset/csv.py
adamgayoso: A logging statement + explanation in the docstring should suffice.

scvi/dataset/anndataset.py
```diff
@@ -0,0 +1,141 @@
...
+    def __init__(self, ad: anndata.AnnData):
+        super().__init__()
+        (
+            X,
+            batch_indices,
+            labels,
+            gene_names,
+            cell_types,
+            self.obs,
+            self.obsm,
```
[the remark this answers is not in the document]
adamgayoso: Not really, no.

scvi/dataset/loom.py
gabmis: @ANazaret we can load and use the API call is what I meant

gabmis: but it's more efficient this way so no reason to change it. @ANazaret we should keep this behaviour, to respect the contract that our database should be trainable directly, and for consistency with the rest of the codebase

scvi/dataset/dataset10X.py
romain-lopez: oh that I did not know, then I think we should modify the gene_filter function but not the data loading API, right ?

scvi/dataset/loom.py
romain-lopez: yes, I think for the cells we need to remove them (plus log a warning somewhere that we did that and which cells, especially if this is a dataset manually imported) otherwise we won't be able to train the model directly. And as @gabmis said, we want the models to be trainable directly.

scvi/dataset/dataset10X.py
gabmis:
```python
scaler.fit(np.zeros((10, 10)))
Out[9]: StandardScaler(copy=True, with_mean=True, with_std=True)
scaler.var_
Out[10]: array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
```

gabmis: what do you mean @adamgayoso ?

gabmis: I agree that we should try to isolate the protein data and store them as ``CellMeasurements`` . For the rest, the user can filter if he wants, apart from empty cells he's solely responsible of what's in the dataset when he trains!

adamgayoso:
```
In [61]: scaler = StandardScaler(with_mean=False)                                                                                                                                                                                   

In [62]: scaler.fit(np.zeros((10, 10))) 
    ...:                                                                                                                                                                                                                            
Out[62]: StandardScaler(copy=True, with_mean=False, with_std=True)

In [63]: scaler.var_                                                                                                                                                                                                                
Out[63]: array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])

In [64]: scaler.scale_                                                                                                                                                                                                              
Out[64]: array([1., 1., 1., 1., 1., 1., 1., 1., 1., 1.])
```

adamgayoso: We use `.var_` so should be fine. Very weird behavior.

scvi/dataset/loom.py
gabmis: okay, added arguments to specify which names to use to retrieve standard scVI attributes (labels, batch_indices, etc..) and loaded the rest using the attribute dicts.

scvi/dataset/anndataset.py
jeff-regier: That could be it. If not, no big deal.

scvi/dataset/csv.py
jeff-regier: We don't need to worry a lot about backward compatibility until we hit version 1.0.

setup.py
```diff
@@ -30,41 +30,49 @@
...
 
-author = 'Romain Lopez, Jeffrey Regier, Maxime Langevin, Edouard Mehlman, Yining Liu'
+author = (
+    "Romain Lopez, "
+    "Jeffrey Regier, "
+    "Maxime Langevin, "
+    "Edouard Mehlman, "
+    "Yining Liu, "
+    "Achille Nazaret, "
+    "Gabriel Misrachi"
```
jeff-regier: Would you add Pierre and Oscar too please?

gabmis: of course

gabmis:
I found some attributes named "X" in "osmFish...loom", probably coordinates.
This messes up our attribute handling.

One workaround I found is to have a field of `protected_attributes` = ["X"] for now and prevent the `"initalize.."` functions to set this attribute. And, if they're asked to, they log a warning and set the attribute as "X_cell" or "X_gene" instead, depending on which initializer is called.

This is not very elegant but seems like a good fix overall and allows a safer general attribute handling.

The ideal way would be to have different objects for each type of attributes, e.g have a structure like `dataset.cell_attributes.X`  and  `dataset.gene_attributes.X` .. (as I expected  in the beginning, we're recoding `anndata.AnnData`)

jeff-regier:
Is this still a WIP? @ANazaret it looks like we need a review from you before merging this PR.

gabmis:
> Is this still a WIP? @ANazaret it looks like we need a review from you before merging this PR.

We want to merge @adamgayoso 's fixes before merging this PR. I'll also have a last look around today

gabmis:
@romain-lopez @jeff-regier Do you know why we inherit from `torch.utils.data.Dataset` ?
It does not seem like we're doing anything with it. We have a dummy `__getitem__` method which clearly doesn't look useful at all.

jeff-regier:
It doesn't seem necessary to inherit from dataset

scvi/dataset/csv.py
gabmis: I removed the weird defaults

romain-lopez:
see #404, closing this one

PR #375, opened 2019-06-19
add kl_epochs parameter to control warmup procedure

### LINKED ISSUE #307: KL Weight (deterministic warmup)
The KL weight warmup of `self.epoch/400` might be too slow to converge to the ELBO.

https://github.com/YosefLab/scVI/blob/e27cb81812e7bff1b8dfa3624292c353dfbcda96/scvi/inference/inference.py#L50

adamgayoso:
Fixes #307 by removing the `kl` parameter and replacing it with `n_epochs_kl_warmup`, which controls how many epochs it will take for the training objective to reach the ELBO (in a linear warmup fashion).

The default is for no warmup. The `kl_weight` can no longer be larger than 1.

### DIFF
--- a/scvi/inference/inference.py
+++ b/scvi/inference/inference.py
@@ -16,8 +16,8 @@ class UnsupervisedTrainer(Trainer):
         :gene_dataset: A gene_dataset instance like ``CortexDataset()``
         :train_size: The train size, either a float between 0 and 1 or and integer for the number of training samples
          to use Default: ``0.8``.
-        :kl_epochs: Number of epochs for linear warmup of KL(q(z)||p(z)) term. After `kl_epochs`, the training
-            objective is the ELBO. This might be used to prevent inactivity of latent units, or to improve
+        :n_warmup_kl_epochs: Number of epochs for linear warmup of KL(q(z)||p(z)) term. After `n_warmup_kl_epochs`,
+            the training objective is the ELBO. This might be used to prevent inactivity of latent units, or to improve
             clustering of latent space, as a long warmup turns the model into something more of an autoencoder.
         :\*\*kwargs: Other keywords arguments from the general Trainer class.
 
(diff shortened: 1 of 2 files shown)


adamgayoso:
@romain-lopez Does the adapter trainer need access to the warmup?

scvi/inference/inference.py
```diff
@@ -16,6 +16,9 @@ class UnsupervisedTrainer(Trainer):
         :gene_dataset: A gene_dataset instance like ``CortexDataset()``
         :train_size: The train size, either a float between 0 and 1 or and integer for the number of training samples
          to use Default: ``0.8``.
+        :n_warmup_kl_epochs: Number of epochs for linear warmup of KL(q(z)||p(z)) term. After `n_warmup_kl_epochs`,
```
gabmis: `n_epochs_kl_warmup` sounds more natural to me.

gabmis:
@romain-lopez I'm leaving the choice of the default value to you, either 400 to keep the behaviour of the trainer as it was before. Or `None` to have no warmup by default

scvi/inference/inference.py
```diff
@@ -16,6 +16,9 @@ class UnsupervisedTrainer(Trainer):
         :gene_dataset: A gene_dataset instance like ``CortexDataset()``
         :train_size: The train size, either a float between 0 and 1 or and integer for the number of training samples
          to use Default: ``0.8``.
+        :n_epochs_kl_warmup: Number of epochs for linear warmup of KL(q(z)||p(z)) term. After `n_epochs_kl_warmup`,
```
romain-lopez: q(z |x) || p(z), for less confusion and stick to the paper notations ?

scvi/inference/inference.py
```diff
@@ -47,7 +50,10 @@ def loss(self, tensors):
...
 
     def on_epoch_begin(self):
-        self.kl_weight = self.kl if self.kl is not None else min(1, self.epoch / 400)  # self.n_epochs)
+        if self.n_epochs_kl_warmup is not None:
+            self.kl_weight = min(1, self.epoch / self.n_epochs_kl_warmup)
+        else:
+            self.kl_weight = 1.0
 
 
 class AdapterTrainer(UnsupervisedTrainer):
```
romain-lopez: @adamgayoso I believe the adapter trainer is only used to optimize the variational network at test time so we probably do not need to use any warmup on this

scvi/inference/inference.py
```diff
@@ -47,7 +50,10 @@ def loss(self, tensors):
         return loss
 
     def on_epoch_begin(self):
-        self.kl_weight = self.kl if self.kl is not None else min(1, self.epoch / 400)  # self.n_epochs)
+        if self.n_epochs_kl_warmup is not None:
+            self.kl_weight = min(1, self.epoch / self.n_epochs_kl_warmup)
+        else:
```
romain-lopez: I know that certain datasets really need a different warmup to get good clustering results. Is there a way to keep the same default than before ?

adamgayoso: Yes, basically just make the default value 400, which I will do.

scvi/inference/inference.py
adamgayoso: Sure

romain-lopez:
looks good to me

PR #381, opened 2019-06-25
[WIP] Achille/gimvi

### LINKED ISSUE #380: Implement spatial-sequencing joint model (gimVI)
Provide an implementation of the model described in:
 _A joint model of unpaired data from scRNA-seq and spatial transcriptomics imputing missing gene expression measurements_ (https://arxiv.org/abs/1905.02269)

Add datasets to test the model

ANazaret:
Closes #380


scvi/dataset/smfish.py
```diff
@@ -63,9 +64,9 @@ def preprocess(self):
                     new_labels.append(5)
 
             select = np.array(to_keep)
-            labels, cell_types = np.array(new_labels), np.array(ds.ca['ClusterName'])
+            labels = np.array(new_labels)
             labels = np.reshape(labels, (labels.shape[0], 1))
-            cell_types = np.reshape(cell_types, (cell_types.shape[0], 1))[select]
```
romain-lopez: cell_types is usually used in the codebase for the name of th ecell types and not the labels. Was this a glitch ?

scvi/dataset/starmap.py
```diff
@@ -0,0 +1,72 @@
...
+import loompy
+
+
+class StarmapDataset(GeneExpressionDataset):
+    r""" Loads the starmap dataset
+
+
+    """
+
+    def __init__(self, save_path='data/', file='starmap.loom', without_positions=False, scrna_seq_genes=None):
```
romain-lopez: Same thing here, can you write some docstrings about what scrna_seq_genes mean?

scvi/inference/__init__.py
```diff
@@ -11,6 +11,7 @@
     ClassifierTrainer
 )
 from .fish import TrainerFish
```
romain-lopez: shouldn't we delete trainer fish ?

scvi/inference/__init__.py
```diff
@@ -23,6 +24,7 @@
     'SemiSupervisedTrainer',
     'AlternateSemiSupervisedTrainer',
     'ClassifierTrainer',
+    'JVAETrainer',
```
romain-lopez: same thing around here probably

scvi/inference/inference.py
```diff
@@ -37,7 +39,10 @@ def __init__(self, model, gene_dataset, train_size=0.8, test_size=None, n_epochs
         if type(self) is UnsupervisedTrainer:
             self.train_set, self.test_set = self.train_test(model, gene_dataset, train_size, test_size)
             self.train_set.to_monitor = ['elbo']
-            self.test_set.to_monitor = ['elbo']
+            if len(self.test_set.indices) >= 5:
```
romain-lopez: good fix, it avoids bug downstream

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+    def elbo(self, verbose: bool = False) -> float:
+        elbo = compute_elbo(self.model, self, mode=self.mode)
+        if verbose:
+            logger.info("ELBO : %.4f" % elbo)
+        return elbo
+
+
+class JVAETrainer(Trainer):
+    r"""
+    The VariationalInference class for the unsupervised training of an autoencoder.
```
romain-lopez: VariationalInference class ?

romain-lopez: that might need to be updated

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+
+    :param model: A model instance from class ``JVAE``
+    :param discriminator: A model instance of a classifier (with logit output)
+    :param gene_dataset_list: list of gene_dataset instance like ``[CortexDataset(), SmfishDataset()]``
```
romain-lopez: does these dataset need to have a certain property ? (one included in the other ?)

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+    :param train_size: Train size on cells
```
romain-lopez: training set split ratio ? between 0 and 1

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+        self.n_epochs_kl_warmup = n_epochs_kl_warmup
+        self.kappa_target = kappa
+        self.all_dataset = [
+            self.create_posterior(
+                gene_dataset=gd, type_class=partial(JPosterior, mode=i)
+            )
+            for i, gd in enumerate(gene_dataset_list)
+        ]
+        self.n_dataset = len(self.all_dataset)
+        self.all_train, self.all_test = list(
```
romain-lopez: that's how this is extended to multiple datasets, I see

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+        self.n_epochs = None
+
+        self.track_disc = []
+
+    def on_epoch_begin(self):
+        if self.n_epochs_kl_warmup is not None:
+            self.kl_weight = min(1, self.epoch / self.n_epochs_kl_warmup)
+        else:
+            self.kl_weight = 1.0
+        self.kappa = (self.epoch / self.n_epochs) * self.kappa_target
```
romain-lopez: I did not remember you did that in the manuscript. what is self.n_epochs, do we have a docstring somehwere ?

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+
+    @property
+    def posteriors_loop(self):
+        return ["train_%d" % i for i in range(self.n_dataset)]
+
+    def data_loaders_loop(self):
+        posteriors = [self._posteriors[name] for name in self.posteriors_loop]
```
romain-lopez: With this, when we train on a dataset with 1000 samples for largest dataset and 100 for the smallest. 1 epochs for our model means 10 epochs for the smallest ? We might want to write this somewhere in the docstrings

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+                            latent_tensors = []
+                            for (i, (data, *_)) in enumerate(tensors):
+                                z = self.model.sample_from_posterior_z(
+                                    data, mode=i, deterministic=False
+                                )
+                                latent_tensors.append(z)
+
+                            # Train discriminator
+                            d_loss = self.loss_discriminator(
+                                [t.detach() for t in latent_tensors], True
```
romain-lopez: is the detach necessary ? you're supposed to only take gradients with respect to the parameters of the discriminator ?

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+                            )
+                            d_loss *= self.kappa
```
romain-lopez: The kappa is not useful here ? it just scales the gradients ?

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+                            d_optimizer.zero_grad()
+                            d_loss.backward()
+                            d_optimizer.step()
+
+                            # Train generative model to fool discriminator
+                            fool_loss = self.loss_discriminator(latent_tensors, False)
+                            fool_loss *= self.kappa
+                            g_optimizer.zero_grad()
+                            fool_loss.backward()
```
romain-lopez: here I understand the kappa better and why you cut the optimization of the composite loss

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+    ) -> Union[List[torch.Tensor], torch.Tensor]:
+        """
+        Compute the loss of the discriminator (either for the true labels or the fool labels)
+
+        :param latent_tensors: Tensors for each dataset of the latent space
+        :param predict_true_class: Specify if the loss aims at minimizing the accuracy or the mixing
+        :param return_details: Boolean used to inspect the loss values, return detailed loss for each dataset
+        :return: scalar loss if return_details is False, else list of scalar losses for each dataset
+        """
+        log_softmax = nn.LogSoftmax(dim=1)
```
romain-lopez: is naming this operator necessary ?

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+        for i, z in enumerate(latent_tensors):
+            cls_logits = log_softmax(self.discriminator(z))
+
+            if predict_true_class:
+                cls_target = torch.zeros(
+                    n_classes, dtype=torch.float32, device=z.device
+                )
+                cls_target[i] = 1.0
+            else:
+                cls_target = torch.ones(
```
romain-lopez: What is this for Predict_true_class ? To learn the optimal discriminator, you minimize the cross entropy between estimated and true class. To include in the adversarial loss (the H divergence) is it another term then ? I'm asking because it looks like this is where the error was last time we talked about it

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+            return losses
+
+        total_loss = torch.stack(losses).sum()
+        return total_loss
+
+    def loss(
+        self, tensors: Iterable[torch.Tensor], return_details: bool = False
+    ) -> Union[torch.Tensor, Tuple[List[torch.Tensor], List[torch.Tensor]]]:
+        """
+        Compute the loss of the scvi model
```
romain-lopez: scvi ?

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+            z = self.model.sample_from_posterior_z(data, mode=i, deterministic=True)
+            cls_z = nn.Softmax(dim=1)(self.discriminator(z)).detach()
+
+            if self.use_cuda:
+                cls_z = cls_z.cpu()
+            cls_z = cls_z.numpy()
+
+            row = cls_z.mean(axis=0)
+            confusion.append(row)
+        return np.array(confusion)
```
romain-lopez: that does not look like a confusion, more like an average prediction. I understand you probably want it as close to uniform as possible ? But it also depends on  the size of the datasets ?

romain-lopez: why do we need that ?

romain-lopez: It looks like this should be in the posterior class, as this is sending back some information about the data.

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+            for i in range(self.n_dataset):
+                total_reconstruction[i] += reconstruction_losses[i].item()
+                total_kl_divergence[i] += kl_divergences[i].item()
+                total_discriminator[i] += discriminator_losses[i].item()
+            if one_sample:
+                break
+
+        return total_reconstruction, total_kl_divergence, total_discriminator
+
+    def get_latent(self, deterministic: bool = True) -> List[np.ndarray]:
```
romain-lopez: Same for this ? Posterior and not trainer ?

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+                        sample_batch, mode, deterministic=deterministic
+                    )
+                )
+
+            latent = torch.cat(latent).detach().numpy()
+            latents.append(latent)
+
+        return latents
+
+    def get_imputed_values(
```
romain-lopez: Same here ?

scvi/models/classifier.py
```diff
@@ -4,13 +4,36 @@
...
+            ),
+            nn.Linear(n_hidden, n_labels),
+            nn.Softmax(dim=-1),
         )
 
     def forward(self, x):
         return self.classifier(x)
+
+
+class ClassifierLogit(nn.Module):
```
romain-lopez: Do we really need both ? We can't refactor this ?

scvi/models/jvae.py
```diff
@@ -0,0 +1,372 @@
...
+            if ql_m is not None:
+                library = ql_m
+        px_scale, px_r, px_rate, px_dropout = self.decode(
+            z, decode_mode, library, batch_index, y
+        )
+
+        return px_scale
+
+    # This is a potential wrapper for a vae like get_sample_rate
+    def get_sample_rate(self, x, batch_index, *_, **__):
```
romain-lopez: I'm not quite familiar with *_ and *__

scvi/models/jvae.py
```diff
@@ -0,0 +1,372 @@
...
+    ) -> torch.Tensor:
+        reconstruction_loss = None
+        if self.reconstruction_losses[mode] == "zinb":
+            reconstruction_loss = -log_zinb_positive(x, px_rate, px_r, px_dropout)
+        elif self.reconstruction_losses[mode] == "nb":
+            reconstruction_loss = -log_nb_positive(x, px_rate, px_r)
+        elif self.reconstruction_losses[mode] == "poisson":
+            reconstruction_loss = -torch.sum(Poisson(px_rate).log_prob(x), dim=1)
+        elif self.reconstruction_losses[mode] == "gaussian":
+            reconstruction_loss = -torch.sum(Normal(px_rate, 10).log_prob(x), dim=1)
```
romain-lopez: why is this still here ? Especially with a fixed std at 10

scvi/models/jvae.py
```diff
@@ -0,0 +1,372 @@
...
+            n_layers_shared=n_layers_decoder_shared,
+            n_cat_list=[self.n_batch],
+            dropout_rate=dropout_rate_decoder,
+        )
+
+        if self.dispersion == "gene":
+            self.px_r = torch.nn.Parameter(torch.randn(self.total_genes))
+        elif self.dispersion == "gene-batch":
+            self.px_r = torch.nn.Parameter(torch.randn(self.total_genes, n_batch))
+        elif self.dispersion == "gene-label":
```
romain-lopez: do we need the gene-label case ?

scvi/models/jvae.py
```diff
@@ -0,0 +1,372 @@
...
+            n_heads=len(dim_input_list),
+            n_input_list=dim_input_list,
+            n_output=self.n_latent,
+            n_hidden=dim_hidden_encoder,
+            n_layers_individual=n_layers_encoder_individual,
+            n_layers_shared=n_layers_encoder_shared,
+            dropout_rate=dropout_rate_encoder,
+        )
+
+        self.l_encoders = ModuleList(
```
romain-lopez: fun !

scvi/models/jvae.py
```diff
@@ -0,0 +1,372 @@
...
+        mode: int,
+        library: torch.Tensor,
+        batch_index: Optional[torch.Tensor] = None,
+        y: Optional[torch.Tensor] = None,
+    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
+        px_scale, px_r, px_rate, px_dropout = self.decoder(
+            z, mode, library, self.dispersion, batch_index, y
+        )
+
+        if self.dispersion == "gene-label":
```
romain-lopez: same question

scvi/models/modules.py
```diff
@@ -19,10 +53,21 @@ class FCLayers(nn.Module):
     :param n_layers: The number of fully-connected hidden layers
     :param n_hidden: The number of nodes per hidden layer
     :param dropout_rate: Dropout rate to apply to each of the hidden layers
+    :param n_batch_norm: number of conditional batch normalization (1 is standard)
```
romain-lopez: What does this number mean ? Can we be more explicit ?

scvi/models/modules.py
```diff
@@ -32,23 +77,33 @@ def __init__(self, n_in: int, n_out: int, n_cat_list: Iterable[int] = None,
...
         self.fc_layers = nn.Sequential(collections.OrderedDict(
             [('Layer {}'.format(i), nn.Sequential(
                 nn.Linear(n_in + sum(self.n_cat_list), n_out),
                 # Below, 0.01 and 0.001 are the default values for `momentum` and `eps` from
                 # the tensorflow implementation of batch norm; we're using those settings
                 # here too so that the results match our old tensorflow code. The default
                 # setting from pytorch would probably be fine too but we haven't tested that.
-                nn.BatchNorm1d(n_out, momentum=.01, eps=0.001) if use_batch_norm else None,
+                nn.BatchNorm1d(n_out, momentum=.01, eps=0.001) if n_batch_norm == 1 else None,
+                ConditionalInstanceNormalization(
```
romain-lopez: I thought we were not using this anymore ? Maybe at some point for ICML submission ?

scvi/models/modules.py
```diff
@@ -273,3 +330,129 @@ def forward(self, x: torch.Tensor, *cat_list: int):
...
+        self.mean_encoder = nn.Linear(n_hidden, n_output)
+        self.var_encoder = nn.Linear(n_hidden, n_output)
+
+    def forward(self, x: torch.Tensor, head_id: int, *cat_list: int):
+        q = self.encoders[head_id](x, *cat_list)
+        q = self.encoder_shared(q, *cat_list)
+
+        q_m = self.mean_encoder(q)
+        q_v = torch.exp(self.var_encoder(q))
+        latent = reparameterize_gaussian(q_m, q_v)
```
romain-lopez: do we really need this reparameterize gaussian ? not a critical point

romain-lopez:
Oh, and the code coverage seems low, we might want to add some unit tests but you might already be working on that :) Take my comments as a first high level review :)

scvi/dataset/cortex.py
```diff
@@ -76,7 +76,10 @@ def preprocess(self):
                 if gene_names[gene_cortex].lower() == gene_fish.lower():
                     additional_genes.append(gene_cortex)
 
-        selected = np.std(expression_data, axis=0).argsort()[-self.additional_genes:][::-1]
+        if self.additional_genes > 0:
```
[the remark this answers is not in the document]
ANazaret: It will be in the pull request #374, I just pushed it quickly.

scvi/dataset/dataset.py
```diff
@@ -182,7 +182,7 @@ def _cell_type_idx(self, cell_types):
 
     def _gene_idx(self, genes):
         if type(genes[0]) is not int:
-            genes_idx = [np.where(gene == self.gene_names)[0][0] for gene in genes]
+            genes_idx = [np.where(gene.lower() == self.gene_names)[0][0] for gene in genes]
```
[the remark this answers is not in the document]
ANazaret: It matters when you want to match the genes between the two datasets (you do string equality).
I had issues with one dataset being uppercase and one being a mix but mostly lowercase.

Once again, I guess we can do that in #374 because for now all datasets have different behaviors.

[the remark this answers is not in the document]
ANazaret: You're right uppercase seems to be standard (even though it always hurt my eyes).

scvi/dataset/dataset.py
```diff
@@ -292,7 +292,7 @@ def library_size(X):
         return local_mean, local_var
 
     @staticmethod
-    def get_attributes_from_matrix(X, batch_indices=0, labels=None):
+    def get_attributes_from_matrix(X, batch_indices=0, labels=None, subclusters=None):
```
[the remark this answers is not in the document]
ANazaret: Another example of the bad api of the current GeneExpressiondataset (where you have to change the base class prototype).
Subclusters are an attribute from one of the new dataset, I took the code from jules branch and not sure exactly.
It will be cleaned in #374

scvi/dataset/dropseq.py
```diff
@@ -0,0 +1,93 @@
...
+    those whose variance in the expression matrix were above a certain threshold and we also kept all the genes who
+    were studied in the Starmap experiment. We have a 71639*7611 gene expression matrix from which we subsample to
+    keep only `subsample` cells
+    """
+
+    def __init__(self, save_path='data/', genes_to_keep=[], subsample=15000):
+        self.genes_to_keep = genes_to_keep
+        self.subsample = subsample
+
+        self.url = ['https://github.com/YosefLab/scVI-data/raw/master/dropseq.loom']
```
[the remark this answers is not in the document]
ANazaret: Can I have the push access on this one?

scvi/dataset/dropseq.py
```diff
@@ -0,0 +1,93 @@
...
+                labels = np.array(ds.ca['Clusters'])
+                labels = np.reshape(labels, (labels.shape[0], 1))[select]
+
+            if 'CellTypes' in ds.attrs:
+                cell_types = np.array(ds.attrs['CellTypes'])
+
+            data = ds[:, select].T  # change matrix to cells by genes
+            ds.close()
+
+            np.random.seed(0)
```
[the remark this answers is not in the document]
ANazaret: Yes it does and it can be a problem you're right. 
The proper way to do it is to have a separate random generator:
`random_state = np.random.RandomState(seed=0)` and then use it exactly as `np.random` like
`random_state.poisson()` for instance.

scvi/dataset/dropseq.py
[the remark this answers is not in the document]
romain-lopez: sent access

scvi/dataset/smfish.py
ANazaret: Not sure I get your comment. Cell types in smfish were wrong because they were not the names indeed but an array for each cell. I am removing this wrong line in the commit.

scvi/inference/jvae_trainer.py
ANazaret: Thx

scvi/inference/jvae_trainer.py
ANazaret: Oh good one! I tried this recently and left it. I think I can remove this warmup which didn't make much a difference.

scvi/inference/jvae_trainer.py
ANazaret: True, I just think if you don't detach them, autograd will compute their gradients on backward which takes computation time, even if d_optimizer.step() will not use the gradient.

scvi/dataset/dataset.py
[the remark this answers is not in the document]
romain-lopez: ok got it

scvi/inference/jvae_trainer.py
ANazaret: Not really ...

scvi/dataset/dropseq.py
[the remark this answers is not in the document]
romain-lopez: ok let's fix that then

scvi/inference/jvae_trainer.py
ANazaret: The error in the fish branch was that the adversarial loss was `-CE(true_class, prediction)`. With this if condition, you compute the optimal loss `+CE(true_class, prediction)` and the adversarial loss with `+CE(false_classes, prediction)`.
Is it right for you?

scvi/dataset/smfish.py
romain-lopez: yes that's my point, ok it was a glitch

scvi/inference/jvae_trainer.py
romain-lopez: ok perfect

scvi/inference/jvae_trainer.py
romain-lopez: ok, the code might be more readable without it but it looks good to me

jeff-regier:
Looks good Achille! What do you have in mind for tests? Just adding a notebook that illustrates usage should bring the test coverage back up.

I'd still like to see some experiments with synthetic data. These experiments could serve as tests too if you were to do what Oscar and Pierre did here, using a `model_fit` boolean variable to keep the tests from taking too long on travis-ci:
https://github.com/YosefLab/scVI/blob/master/tests/test_zeros_synth.py

scvi/inference/__init__.py
ANazaret: Yes you're right!

scvi/inference/jvae_trainer.py
ANazaret: Yes it scales gradient, but that's the goal

scvi/models/modules.py
ANazaret: Not sure I get your question, yes it is the reparametrization trick for variational inference.

scvi/models/modules.py
ANazaret: I thought I could leave it to check hyperoptimization of vanilla scvi with this `ConditionalInstanceNormalization`. Maybe it could help scvi to correct batch effect. What do you think?

scvi/inference/jvae_trainer.py
ANazaret: Not necessarily. The user has to provide a mapping of the genes from each dataset to `Union(d.gene_names for d in datasets)`

scvi/inference/jvae_trainer.py
ANazaret: That's true. It's kind of bootstrapping the smaller datasets to reach the bigger one size.

scvi/inference/jvae_trainer.py
ANazaret: Indeed we want it as close to uniform as possible but it is still a confusion matrix.
It gives insights on the adversarial training, to check how the discriminator is fooled.

scvi/inference/jvae_trainer.py
ANazaret: It seems difficult to add this in Posterior as for now Posterior is bond to a specific GeneExpressionDataset and this function returns the latent space for all the datasets.

scvi/models/modules.py
romain-lopez: I mean, do we need the function called reparameterize_gaussian ? Instead of torch.Normal(...).rsample() ?

scvi/models/modules.py
romain-lopez: ok that works for me

scvi/inference/jvae_trainer.py
romain-lopez: great !

scvi/inference/jvae_trainer.py
romain-lopez: ok thanks

scvi/inference/jvae_trainer.py
romain-lopez: isn't it equivalent to change the learning rate ? I thought Adam would naturally do that.

scvi/inference/jvae_trainer.py
romain-lopez: ok !

scvi/inference/jvae_trainer.py
romain-lopez: cf. our conversation, got it

scvi/models/modules.py
ANazaret: Imo, it increases the code readability: the reader understand that this line is executed to "reparameterize the gaussian".

scvi/models/modules.py
ANazaret: I updated the docstring

romain-lopez: great

scvi/models/modules.py
romain-lopez: good to me

scvi/inference/jvae_trainer.py
ANazaret: Not exactly: you have 3 terms in the loss 
- A = `d_loss` (train discriminator)
- B = `fool_loss` (train encoder of generative model to fool discriminator)
- C = `g_loss` (train the generative model to model the data)

The kappa scales A,B
The learning rate of the discriminator scales A
The learning rate of the generator scales B, C

For fun, the matrix
> 1 1 0
> 1 0 0
> 0 1 1

is invertible so the system of parameters is not redundant :)

scvi/models/classifier.py
ANazaret: Done!

adamgayoso:
This looks great - one point is that we might not want to name the datasets by their protocol name (dropseq, starmap). Can we exchange with a more informative name?

ANazaret:
> This looks great - one point is that we might not want to name the datasets by their protocol name (dropseq, starmap). Can we exchange with a more informative name?

Yes I'm doing it!

jeff-regier:
Should we close this PR? (Is it superseded by #415?)

ANazaret:
Yes we should. It's outdated now that #404 has been merged

PR #384, opened 2019-06-27
feat ZI for generate

### LINKED ISSUE #379: Update posterior.generate() to reflect the model likelihood
We should use `vae.reconstruction_loss` in the following function so that generate actually generates the correct data. Right now it is set to only generate NB data.

https://github.com/YosefLab/scVI/blob/31e87f4de0baeca85f39f5e510de2baf6af1efc7/scvi/inference/posterior.py#L508

PierreBoyeau:
Fixes #379 

PR based on the work we did with @oscarclivio to sample from the Posterior Predictive

Some remarks:
- Right now the tests do not pass because I used the new conventions for the `Dataset` class for the implementation and in the test.
- I did everything using torch, but should I transfer the objects to CPU memory in the end? That is what I did right now but feel free to tell me what you think

@oscarclivio can you review just in case?

### DIFF
--- a/scvi/inference/posterior.py
+++ b/scvi/inference/posterior.py
@@ -543,18 +543,20 @@ class Posterior:
             # Clamping as distributions objects can have buggy behaviors when
             # their parameters are too high
             l_train = torch.clamp(l_train, max=1e8)
-            gene_expressions = distributions.Poisson(l_train).sample()
+            gene_expressions = distributions.Poisson(l_train).sample() # Shape : (n_samples, n_cells_batch, n_genes)
             if zero_inflated:
                 p_zero = (1.0 + torch.exp(-px_dropout)).pow(-1)
                 random_prob = torch.rand_like(p_zero)
                 gene_expressions[random_prob <= p_zero] = 0
 
+
+            gene_expressions = gene_expressions.permute([1, 2, 0]) # Shape : (n_cells_batch, n_genes, n_samples)
+
             x_old.append(sample_batch)
             x_new.append(gene_expressions)
 
-        x_old = torch.cat(x_old)  # Shape (n_cells, n_genes
-        x_new = torch.cat(x_new)  # Shape (n_samples, n_cells, n_genes)
-        x_new = x_new.permute([1, 2, 0])  # Shape (n_cells, n_genes, n_samples)
+        x_old = torch.cat(x_old)  # Shape (n_cells, n_genes)
+        x_new = torch.cat(x_new)  # Shape (n_cells, n_genes, n_samples)
         if genes is not None:
             gene_ids = self.gene_dataset._gene_idx(genes)
             x_new = x_new[:, gene_ids, :]
--- a/scvi/inference/posterior.py
+++ b/scvi/inference/posterior.py
@@ -2,7 +2,7 @@ import copy
 import os
 import logging
 
-from typing import List, Optional, Union
+from typing import List, Optional, Union, Tuple
 
 import numpy as np
 import pandas as pd
@@ -509,9 +509,9 @@ class Posterior:
     def generate(
         self,
         n_samples: int = 100,
-        genes: Union[list, np.array] = None,
+        genes: Union[list, np.ndarray] = None,
         batch_size: int = 128
-    ):
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
         """
         Create observation samples from the Posterior Predictive distribution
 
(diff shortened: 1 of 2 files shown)


scvi/inference/posterior.py
```diff
@@ -505,39 +506,58 @@ def imputation(self, n_samples=1):
         return imputed_list.squeeze()
 
     @torch.no_grad()
-    def generate(self, n_samples=100, genes=None):  # with n_samples>1 return original list/ otherwose sequential
+    def generate(
+        self,
+        n_samples: int = 100,
+        genes: Union[list, np.array] = None,
```
gabmis: `np.array` is a callable,  `np.ndarray` is what you are looking for.

scvi/inference/posterior.py
```diff
@@ -505,39 +506,58 @@ def imputation(self, n_samples=1):
...
-        :param n_samples:
-        :param genes:
-        :return:
+        Create observation samples from the Posterior Predictive distribution
+
+        :param n_samples: Number of required samples for each cell
+        :param genes: Indices of genes of interest
+        :param batch_size: Desired Batch size to generate data
+
+        :return: Tuple (x_new, x_old)
```
gabmis: python supports Typing for returned values, it is very useful when developing in smart IDEs like pycharm, please consider adding it here and elsewhere in general.

scvi/inference/posterior.py
```diff
@@ -505,39 +506,58 @@ def imputation(self, n_samples=1):
...
+                gene_expressions[random_prob <= p_zero] = 0
+
+            x_old.append(sample_batch)
+            x_new.append(gene_expressions)
+
+        x_old = torch.cat(x_old)  # Shape (n_cells, n_genes
+        x_new = torch.cat(x_new)  # Shape (n_samples, n_cells, n_genes)
+        x_new = x_new.permute([1, 2, 0])  # Shape (n_cells, n_genes, n_samples)
+        if genes is not None:
+            gene_ids = self.gene_dataset.gene_as_index(genes)
```
gabmis: As of now, this method doesn't exist, are you refering to the future `GeneExpressionDataset`api? If so, please use the current API and i'll make the change before merging the PR.

romain-lopez:
Looks good to me, @gabmis gave some good feedback 👍

scvi/inference/posterior.py
PierreBoyeau: Yes I was refering to the future API, I'll ping you when requested changes are done

scvi/inference/posterior.py
```diff
@@ -505,39 +506,60 @@ def imputation(self, n_samples=1):
...
+            Where x_new has shape (n_cells, n_genes, n_samples)
         """
-        original_list = []
-        posterior_list = []
-        batch_size = 128  # max(self.data_loader_kwargs["batch_size"] // n_samples, 2)  # Reduce batch_size on GPU
+        assert self.model.reconstruction_loss in ['zinb', 'nb']
+        zero_inflated = self.model.reconstruction_loss == 'zinb'
+        x_old = []
+        x_new = []
         for tensors in self.update({"batch_size": batch_size}):
```
adamgayoso: Why exactly is this necessary?  Why can't we iterate through self?

gabmis: ``for sample_batch, _, _, batch_index, labels in self.update({"batch_size": batch_size}):`` is more concise indeed.

adamgayoso: @gabmis I guess I was asking why we need to update the batch size? Doesn't that make a copy? If the tensors get put on GPU during every iteration, why do we need to change the batch size?

oscarclivio:
There was a shape error in the scVI reproductibility notebook when concatenating the generated tensors. It should have been fixed now !

oscarclivio:
Thanks @PierreBoyeau for having updated this for everyone !

Note : There was a shape error in the scVI reproductibility notebook when concatenating the generated tensors. It should have been fixed now.

oscarclivio:
Ok, basically Travis failed on the reproductibility notebook as it required the outputs of generate to be numpy arrays to plot in the cell after the one involving generate. A new push should have corrected that. Is it ok to output numpy arrays instead of (CPU) tensors ? Feel free to let us know!

jeff-regier:
Yes seem fine. If we're not keeping the data on the GPU, we might as well use a numpy format.

scvi/inference/posterior.py
gabmis: I am not sure I get your inquiry Adam. IMO, ``batch_size`` seems a like a reasonable parameter for this function. Besides, making a shallow copy of ``Posterior`` object should be rather inexpensive.

PR #385, opened 2019-06-28
[WIP] Criticism module with PPCs

adamgayoso:
I wanted to get some feedback on this. I think it would be great if we 

1) Have a general PPC class so that we can more efficiently get posterior predictive metrics
2) Disentangle some of the code in Posterior, which really involves model criticism such as batch entropy mixing (just one example)
3) Provide users a principled way to decide between any number of parameterizations through PPCs, etc.

If you think it's not worth it I will just keep this code privately for my own tests, so no big deal.

### DIFF
--- a/scvi/criticism/__init__.py
+++ b/scvi/criticism/__init__.py
@@ -1,0 +1,3 @@
+from .ppc import PosteriorPredictiveCheck
+
+__all__ = ['PosteriorPredictiveCheck']
--- a/scvi/criticism/ppc.py
+++ b/scvi/criticism/ppc.py
@@ -22,7 +22,7 @@ class PosteriorPredictiveCheck:
             n_samples (int, optional): Number of posterior predictive samples. Defaults to 10.
         """
         self.posteriors = posteriors_dict
-        self.dataset = posteriors_dict[next(iter(posteriors_dict.keys()))].dataset
+        self.dataset = posteriors_dict[next(iter(posteriors_dict.keys()))].gene_dataset
         self.raw_counts = None
         self.posterior_predictive_samples = {}
         self.n_samples = n_samples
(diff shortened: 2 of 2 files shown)


jeff-regier:
I'm interested in the idea. If we did add a new PPC class, what methods would have to remain in `Posterior` rather than joining the new `PPC` class?  `PPC` seems like a good abstraction if there's a clear distinction btw it and other classes; less so if there's ambiguity where some methods belong.

adamgayoso:
I could see there being at least three classes :

1) PPC
2) LatentSpace
3) Imputation

1) Allows users to test their model against other scVI models or simpler models using the PPC metrics like Oscar and Pierre used in the blog post.
2) Handles things like batch entropy mixing, clustering scores, knn_purity, etc.
3) Will be a bit trickier but at the very least will contain the post processing imputation metrics and plots.

This leaves Posterior with DE, as well as get_latent(), get_sample_scale() - the functions that help process and access parts of the model.

jeff-regier:
My sense is that there'd be too much overlap between `Posterior` and `PPC` for this new abstraction to be helpful, and that we'd have a hard time maintaining the division btw the class as new people join the project. I'm not sure of that, but so far, it's not clear to me, why, say, the differential expression code would belong in `Posterior` but not `PPC` whereas imputation would be the latter.

Would it work to add the new PPC methods to Posterior, rather the creating a new class?

adamgayoso:
I think that's a better idea for now. I will go ahead and do that.

adamgayoso:
That said, integrating it within posterior will be hard because the example I posted here allows for comparing multiple posteriors. In fact, I think that's really the distinction. 

`Posterior` should be for functions related to one model fit. Classes in criticism should relate to metrics that one would use for model selection.

adamgayoso:
I'm going to close this and put more thought into it. In the meantime, I am using this for my model, so I will place this code in my model file in `inference`.

PR #386, opened 2019-06-28
removing verbose comments

### LINKED ISSUE #368: remove verbose and use module-level logger
Rather than passing `verbose` flags around all over the place, we should be using `logger.debug(...)`. e.g.,
https://github.com/YosefLab/scVI/blob/4eea43e2fefa559271a8c6b5f340611faa90bf7d/scvi/inference/annotation.py#L26-L27

This issue is to get rid of all the `verbose` flags.

At the same time, we could switch to module level logging, rather than using the root logger.

romain-lopez:
fixes #368

### DIFF
--- a/scvi/inference/annotation.py
+++ b/scvi/inference/annotation.py
@@ -17,6 +17,8 @@ from scvi.inference import Trainer
 from scvi.inference.inference import UnsupervisedTrainer
 from scvi.inference.posterior import unsupervised_clustering_accuracy
 
+logger = logging.getLogger(__name__)
+
 
 class AnnotationPosterior(Posterior):
     def __init__(self, *args, model_zl=False, **kwargs):
@@ -26,7 +28,7 @@ class AnnotationPosterior(Posterior):
     def accuracy(self):
         model, cls = (self.sampling_model, self.model) if hasattr(self, 'sampling_model') else (self.model, None)
         acc = compute_accuracy(model, self, classifier=cls, model_zl=self.model_zl)
-        logging.debug("Acc: %.4f" % (acc))
+        logger.debug("Acc: %.4f" % (acc))
         return acc
 
     accuracy.mode = 'max'
@@ -40,7 +42,7 @@ class AnnotationPosterior(Posterior):
         all_y_pred_groups = np.array([self.model.labels_groups[y] for y in all_y_pred])
         h_acc = np.mean(all_y_groups == all_y_pred_groups)
 
-        logging.debug("Hierarchical Acc : %.4f\n" % h_acc)
+        logger.debug("Hierarchical Acc : %.4f\n" % h_acc)
         return acc
 
     accuracy.mode = 'max'
@@ -58,7 +60,7 @@ class AnnotationPosterior(Posterior):
     def unsupervised_classification_accuracy(self):
         all_y, all_y_pred = self.compute_predictions()
         uca = unsupervised_clustering_accuracy(all_y, all_y_pred)[0]
-        logging.debug("UCA : %.4f" % (uca))
+        logger.debug("UCA : %.4f" % (uca))
         return uca
 
     unsupervised_classification_accuracy.mode = 'max'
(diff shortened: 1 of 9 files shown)


jeff-regier:
Nice. Thanks for working in it!

@gabmis suggests switching to module-specific loggers too. That's the first part of the #368. Would you mind making that change in this PR too?

scvi/inference/fish.py
```diff
@@ -13,17 +13,15 @@
 class FishPosterior(Posterior):
 
     @torch.no_grad()
-    def elbo(self, verbose=False):
+    def elbo(self):
         elbo = compute_elbo(self.model, self, mode="smFISH")
-        if verbose:
-            logging.info("ELBO Fish: %.4f" % elbo)
+        logging.debug("ELBO Fish: %.4f" % elbo)
```
gabmis: As discussed, we should use a per-module logger ``logger = logging.getLogger(__name__)``
 declared right after the imorts and then use calls such ``logger.info(..)``

romain-lopez: Oh so no logger.debug ?

scvi/inference/trainer.py
```diff
@@ -94,20 +91,18 @@ def compute_metrics(self):
...
 
                 for name, posterior in self._posteriors.items():
                     logging.info_name = ' '.join([s.capitalize() for s in name.split('_')[-2:]])
                     if hasattr(posterior, 'to_monitor'):
                         for metric in posterior.to_monitor:
                             if metric not in self.metrics_to_monitor:
-                                if self.verbose:
-                                    logging.info(logging.info_name, end=' : ')
-                                result = getattr(posterior, metric)(verbose=self.verbose)
+                                logging.debug(logging.info_name, end=' : ')
```
romain-lopez: @gabmis logging.info_name should be replaced also by the logger ?

scvi/inference/fish.py
gabmis: yes yes, all logs should be made using the module logger, ie ``logger.debug`` , ``logger.warning``

jeff-regier:
Some of the print / logging statements we also may not need any more -- really depends what's in the notebooks. For debugging, it's generally better to run the code in a debugger and add breakpoints than to print / log.

gabmis:
> Some of the print / logging statements we also may not need any more -- really depends what's in the notebooks. For debugging, it's generally better to run the code in a debugger and add breakpoints than to print / log.

True, but when doing so logging level goes to debug and seeing the logs really helps you narrow down your search and set the right breakpoints

scvi/inference/trainer.py
romain-lopez: ok, I thought this was a useful property of the logger

romain-lopez:
@gabmis ready to merge for me

scvi/inference/trainer.py
jeff-regier: I have no idea

scvi/inference/annotation.py
```diff
@@ -17,32 +17,32 @@
...
         super().__init__(*args, **kwargs)
         self.model_zl = model_zl
 
-    def accuracy(self, verbose=False):
+    def accuracy(self):
         model, cls = (self.sampling_model, self.model) if hasattr(self, 'sampling_model') else (self.model, None)
         acc = compute_accuracy(model, self, classifier=cls, model_zl=self.model_zl)
-        if verbose:
-            logging.info("Acc: %.4f" % (acc))
+        logger.debug("Acc: %.4f" % (acc))
```
ANazaret: Hey, is there a reason why you use `debug` instead of `info`? It seems to me it's more an information about the model rather than a software engineering debug information.

romain-lopez: not really, just thought that would be a debugging level information (that we kind of never really want to read unless we have trouble with the model fit)

gabmis: I agree with Romain on this

PR #396, opened 2019-07-03
Use pre-commit to enforce flake8 before comitting and mention the black formatter

### LINKED ISSUE #395: We should have a stronger opinion on code style, and use tools to enforce that

gabmis:
fixes #395 

No more non-flake8 commits

We should all install and use pre-commit with the configuration file provided in the repo
When installed pre-commit runs hooks on each call to git commit and forbids the commit if the hooks don't pass.
We have a minimal configuration in pre-commit-configuration.yml containing just flake8, which we should refine in a timely manner.

For lack of being able to change the code style to [black](https://github.com/python/black) for the whole code base (would mess up the git blame) we should enforce its use for new code as much as possible. This means:
* It's mentioned and strongly recommended in CONTRIBUTING.rst
* flake8 config is black compatible


ANazaret:
Looks good to me!

romain-lopez:
How dark: flake8 did not pass on this PR 
> The command "flake8" exited with 1.

ANazaret:
Do you know if there will be any conflict with this file?
https://github.com/YosefLab/scVI/blob/7df62412f9fa120182b06232075a2d257df09695/setup.cfg#L17-L19

ANazaret:
Hey @gabmis, here are the errors ignored by default by flake8 version 3.7.7 found in their code:
`IGNORE = ("E121", "E123", "E126", "E226", "E24", "E704", "W503", "W504")`
Ref: https://github.com/PyCQA/flake8/blob/076dfeee372affb1dbe849d12a1cc20f8ea37d40/src/flake8/defaults.py#L15

I guess 2 options:
1) You add the default ignore in the `.flake8` file you introduced - but E226 and E241 are actually important to me so it's nice to enforce them
2) You fix the conflicting lines (there are not many lines in the end) - but this alters the blame

Hope it helps!

gabmis:
> Do you know if there will be any conflict with this file?
> https://github.com/YosefLab/scVI/blob/7df62412f9fa120182b06232075a2d257df09695/setup.cfg#L17-L19

@ANazaret thanks for spotting the config I had missed!


> Hey @gabmis, here are the errors ignored by default by flake8 version 3.7.7 found in their code:
> `IGNORE = ("E121", "E123", "E126", "E226", "E24", "E704", "W503", "W504")`
> Ref: https://github.com/PyCQA/flake8/blob/076dfeee372affb1dbe849d12a1cc20f8ea37d40/src/flake8/defaults.py#L15
> 
> I guess 2 options:
> 
> 1. You add the default ignore in the `.flake8` file you introduced - but E226 and E241 are actually important to me so it's nice to enforce them
> 2. You fix the conflicting lines (there are not many lines in the end) - but this alters the blame
> 
> Hope it helps!

It helps! They all seem important to me.

gabmis:
@jeff-regier good to merge when tests pass

PR #397, opened 2019-07-04
CellMeasurement fixes for new dataset

adamgayoso:
- Fixes `Dataset10x` handling of protein measurements
- Fixes how cell measurements get handled as cell attributes and when updating cells
- Adds access to `.uns` of `anndata` object after population

I identified a number of issues and propose fixes here. Feel free to reimplement them how you would like or let me know how I can help.

### DIFF
--- a/scvi/dataset/cite_seq.py
+++ b/scvi/dataset/cite_seq.py
@@ -74,7 +74,7 @@ class CiteSeqDataset(DownloadableDataset):
         )
         protein_names = np.asarray(adt.index).astype(np.str)
         protein_measurement = CellMeasurement(
-            name="adt_expression",
+            name="protein_expression",
             data=adt.T.values,
             columns_attr_name="protein_names",
             columns=protein_names
@@ -87,9 +87,9 @@ class CiteSeqDataset(DownloadableDataset):
         if not np.array_equal(np.asarray(adt_centered.index).astype(np.str), protein_names):
             raise ValueError("Protein names are not the same for raw and centered counts.")
         protein_measurement_centered = CellMeasurement(
-            name="adt_expression_centered",
+            name="protein_expression_clr",
             data=adt_centered.T.values,
-            columns_attr_name="protein_names",
+            columns_attr_name="protein_names_clr",
             columns=protein_names,
         )
 
--- a/scvi/dataset/dataset.py
+++ b/scvi/dataset/dataset.py
@@ -584,7 +584,7 @@ class GeneExpressionDataset(Dataset):
                     n_cells=self.nb_cells, n_attr=len_attribute
                 )
             )
-        setattr(self, attribute_name, np.asarray(attribute))
+        setattr(self, attribute_name, np.asarray(attribute) if type(attribute) is list else attribute)
         self.cell_attribute_names.add(attribute_name)
         if categorical:
             self.cell_categorical_attribute_names.add(attribute_name)
--- a/scvi/dataset/dataset.py
+++ b/scvi/dataset/dataset.py
@@ -577,7 +577,10 @@ class GeneExpressionDataset(Dataset):
                 )
             )
             attribute_name = valid_attribute_name
-        len_attribute = attribute.shape[0] if type(attribute) is not list else len(attribute)
+        try:
+            len_attribute = attribute.shape[0]
+        except AttributeError:
+            len_attribute = len(attribute)
         if not self.nb_cells == len_attribute:
             raise ValueError(
                 "Number of cells ({n_cells}) and length of cell attribute ({n_attr}) mismatch".format(
--- a/scvi/dataset/dataset.py
+++ b/scvi/dataset/dataset.py
@@ -593,7 +593,7 @@ class GeneExpressionDataset(Dataset):
             np.asarray(attribute)
             if not isinstance(attribute, sp_sparse.csr_matrix)
             else attribute,
-        )        
+        )
         self.cell_attribute_names.add(attribute_name)
         if categorical:
             self.cell_categorical_attribute_names.add(attribute_name)
--- a/scvi/dataset/dataset.py
+++ b/scvi/dataset/dataset.py
@@ -587,7 +587,7 @@ class GeneExpressionDataset(Dataset):
                     n_cells=self.nb_cells, n_attr=len_attribute
                 )
             )
-        setattr(self, attribute_name, np.asarray(attribute) if not isinstance(attribute, csr_matrix) else attribute)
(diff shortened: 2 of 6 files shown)


ANazaret:
> `numpy.asarray(x)` does nothing if `x` is already a `numpy.ndarray` , does it?

Nope it doesn't do anything.
Ref: https://docs.scipy.org/doc/numpy/reference/generated/numpy.asarray.html

scvi/dataset/dataset10X.py
```diff
@@ -186,7 +186,8 @@ def populate(self):
         else:
             gene_names = None
             for measurement_type in np.unique(measurements_info[2]):
-                measurement_mask = measurements_info[2] == measurement_type
+                # .values required to work with sparse matrices
+                measurement_mask = (measurements_info[2] == measurement_type).values
```
ANazaret: I'v never seen `.values` for sparse matrix neither for numpy array; is it something like `todense()`?. I might miss something but are you sure it's working? and compatible with numpy array?

scvi/dataset/dataset10X.py
```diff
@@ -195,11 +196,17 @@ def populate(self):
                     gene_expression_data = measurement_data
                     gene_names = measurement_names
                 else:
-                    Ys = Ys if Ys is None else []
+                    Ys = [] if Ys is None else Ys
```
ANazaret: Good catch!

scvi/dataset/dataset.py
```diff
@@ -577,13 +577,14 @@ def initialize_cell_attribute(self, attribute_name, attribute, categorical=False
...
+        len_attribute = attribute.shape[0] if type(attribute) is not list else len(attribute)
+        if not self.nb_cells == len_attribute:
             raise ValueError(
                 "Number of cells ({n_cells}) and length of cell attribute ({n_attr}) mismatch".format(
-                    n_cells=self.nb_cells, n_attr=len(attribute)
+                    n_cells=self.nb_cells, n_attr=len_attribute
                 )
             )
-        setattr(self, attribute_name, np.asarray(attribute))
+        setattr(self, attribute_name, np.asarray(attribute) if type(attribute) is list else attribute)
```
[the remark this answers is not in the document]
ANazaret: ... in order to understand that you make a special case for the sparse_matrices

scvi/dataset/cite_seq.py
```diff
@@ -87,9 +87,9 @@ def populate(self):
         if not np.array_equal(np.asarray(adt_centered.index).astype(np.str), protein_names):
             raise ValueError("Protein names are not the same for raw and centered counts.")
         protein_measurement_centered = CellMeasurement(
-            name="adt_expression_centered",
+            name="protein_expression_clr",
```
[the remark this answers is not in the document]
adamgayoso: `clr` means centered log ratio, it's really just the log counts subtracted by their mean on the log scale, or can be seen as log of counts divided by geometric mean.

scvi/dataset/dataset.py
```diff
@@ -577,13 +577,14 @@ def initialize_cell_attribute(self, attribute_name, attribute, categorical=False
                 )
             )
             attribute_name = valid_attribute_name
-        if not self.nb_cells == len(attribute):
+        len_attribute = attribute.shape[0] if type(attribute) is not list else len(attribute)
```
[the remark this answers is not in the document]
adamgayoso: yes, `len` is not defined for sparse matrices.

scvi/dataset/dataset10X.py
adamgayoso: `measurement_mask` is a pandas DF. For some reason (not sure why) if you use the slice of the data frame column to access columns of the sparse matrix, the shape returned was the same as it was before. `.values` is the pandas way of grabbing the underlying `np.ndarray` so this should be universal.

adamgayoso:
@gabmis @ANazaret is there a way to enforce "protein_expression" as the attribute name for the CITE-seq count matrix?

scvi/dataset/dataset.py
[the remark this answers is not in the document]
gabmis: can you use duck-typing instead of looking for type equality, which is somewhat restrictive.
```python
try:
   len_attribute = attribute.shape[0]
except AttributeError:
   len_attribute = len(attribute)
```

scvi/dataset/dataset.py
```diff
@@ -577,13 +577,14 @@ def initialize_cell_attribute(self, attribute_name, attribute, categorical=False
...
+        if not self.nb_cells == len_attribute:
             raise ValueError(
                 "Number of cells ({n_cells}) and length of cell attribute ({n_attr}) mismatch".format(
-                    n_cells=self.nb_cells, n_attr=len(attribute)
+                    n_cells=self.nb_cells, n_attr=len_attribute
                 )
             )
-        setattr(self, attribute_name, np.asarray(attribute))
+        setattr(self, attribute_name, np.asarray(attribute) if not isinstance(attribute, csr_matrix) else attribute)
```
gabmis: good catch.
In general what's the number of measured protein markers?
I get the feeling that that number is pretty small, so we probably want to densify them in the dataset classes.

scvi/dataset/dataset10X.py
gabmis: not sure why we need this?

scvi/dataset/dataset.py
adamgayoso: it's about 100, no need to be sparse, but this will be necessary in the future.

gabmis: my point is, we should come up with a criterion to densify protein measurements in the dataset classes, like in the anndata loader

scvi/dataset/dataset10X.py
adamgayoso:
```
In [144]: test = csr_matrix(np.zeros((100, 3)))                                                                                                                                                                                    

In [145]: mask = pd.DataFrame(data=np.array([True, False, True]))                                                                                                                                                                  

In [146]: test[:, mask[0]].shape                                                                                                                                                                                                   
Out[146]: (100, 3)

In [147]: test[:, mask[0].values].shape                                                                                                                                                                                            
Out[147]: (100, 2)
```

scvi/dataset/dataset.py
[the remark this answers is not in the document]
adamgayoso: pushed.

scvi/dataset/dataset10X.py
gabmis: uncool..

scvi/dataset/dataset.py
gabmis: Can you please add this kind of behaviour real quick? and fix the test, you have a missing `sp_sparse`. in front of `csr_matrix` somewhere apparently

PR #404, opened 2019-07-10
Refactor gene expression dataset (rebased)

### LINKED ISSUE #270: library local_var and local_mean not calculated by batch

### LINKED ISSUE #337: refactor GeneExpressionDataset

### LINKED ISSUE #383: brain large : bug in gene subset selection

### LINKED ISSUE #389: Numpy version is too old

### LINKED ISSUE #391: can we remove the run_benchmarks.py script? it seems rather useless

### LINKED ISSUE #392: can we remove the scmfish notebook? ultimately replaced by gimVI notebook

gabmis:
fixes #337 
fixes #270 
fixes #383
closes #389 
closes #391
closes #392
prepares fix for #325

See original PR (#374) for details


gabmis:
@jeff-regier @romain-lopez I did what I could, this is not particularly pretty but the number of commits has been limited, so less breaking commits, and the blame has been preserved. Just to be clear, this should be merged not squash and merged.

Issue #412, opened 2019-07-12
CortexDataset is buggy


ANazaret:
1) The gene names just loaded are not uppercase when compared with the `genes_to_keep` so the intersection is null
2) The intersection doesn't preserve the order given by `genes_to_keep`
3) The additional genes are supposed to be the most variable but the `argsort` is wrong (it's applied after the filtration so the indexing is not matching the initial shape)

Issue #411, opened 2019-07-11
`reorder_genes` doesn't keep the given order, and uses alphabetical order instead


ANazaret:
`np.intersect1d` sorts the values alphabetically, so the contract of the `reorder_genes` function is violated. The tests didn't catch the issue because it was unfortunately checking for a new order with genes in alphabetical order.

PR #413, opened 2019-07-12
Achille/dataset enhancements

### LINKED ISSUE #409: Possibilty to `reorder_cell_types`
It would be nice for visualization and coherence between datasets to be able to reorder cell-types.

### LINKED ISSUE #410: `smFishDataset` merges cell types incorectly
`populate_data` calls `.remap_categorical_attributes()` at the end so it mixes the order of cell-types to merge after. We should add a boolean flag to call this function or not.

ANazaret:
Closes #409
Fixes #410 #411 #412 
Add datasets from StarMAP and Drop-seq original experiments
Add tests


scvi/dataset/dataset.py
```diff
@@ -1090,6 +1117,21 @@ def map_cell_types(
...
+    def reorder_cell_types(self, new_order: Union[List[str], np.ndarray]):
+        """Reorder in place the cell-types. The cell-types provided will be added at the beginning of `cell_types`
+        attribute, such that if some existing cell-types are omitted in `new_order`, they will be left after the
+        new given order
+
+        :param new_order:
+        """
+        for cell_type in self.cell_types:
+            if cell_type not in new_order:
+                new_order.append(cell_type)
```
gabmis: you typed `new_order` as a potential `np.ndarray`. You need to type check before calling `append` and treat the `numpy` case accordingly

scvi/dataset/dataset.py
```diff
@@ -1090,6 +1117,21 @@ def map_cell_types(
...
+
+        cell_types = OrderedDict([((x,), x) for x in new_order])
+        self.map_cell_types(cell_types)
+        self.remap_categorical_attributes(["labels"])
```
gabmis: Is this call to `remap_categorical_attributes` really necessary? Isn't the remapping done in `map_cell_types` ?

scvi/dataset/loom.py
```diff
@@ -36,14 +38,17 @@ def __init__(
         url: str = None,
         batch_indices_attribute_name: str = "BatchID",
         labels_attribute_name: str = "ClusterID",
+        encode_labels_name_into_int=False,
```
gabmis: Can you please type, for the sake of uniformity in the codebase.

tests/dataset/test_cortex.py
```diff
@@ -2,9 +2,29 @@
...
 from .utils import unsupervised_training_one_epoch
+import numpy as np
 
 
 class TestCortexDataset(TestCase):
     def test_populate(self):
         dataset = CortexDataset(save_path="tests/data")
         unsupervised_training_one_epoch(dataset)
+
+    def test_variance_and_order_and_size(self):
```
gabmis: cool!

scvi/dataset/dataset.py
ANazaret: Good catch, thank you!  `numpy` arrays are not designed to be resized and for dynamic changes so the best is to convert it in a list at the beginning. Thanks!

scvi/dataset/dataset.py
ANazaret: No it's not done and I think it's better like that. If the user starts using `map_cell_types` it mean the user knows what he's doing with the cell types and we shouldn't change labels under its nose.

gabmis: Agreed

PR #415, opened 2019-07-12
[WIP] Implementation of GimVI

### LINKED ISSUE #380: Implement spatial-sequencing joint model (gimVI)

ANazaret:
Closes #380
Clean branch of previous PR #381


scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+class JPosterior(Posterior):
+    def __init__(self, *args, mode=0, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.mode = mode
+
+    def elbo(self, verbose: bool = False) -> float:
+        elbo = compute_elbo(self.model, self, mode=self.mode)
+        if verbose:
```
ANazaret: I'm removing this `verbose`

scvi/benchmark.py
```diff
@@ -63,33 +42,7 @@ def ldvae_benchmark(dataset, n_epochs, use_cuda=True):
...
-    pca_values_seq = gene_dataset_seq.X
-    pca_labels_seq = gene_dataset_seq.labels
-    pca_labels_fish = gene_dataset_fish.labels
-    _ = proximity_imputation(pca_latent_seq, pca_values_seq[:, 0], pca_latent_fish, k=5)
-    _, _, = compute_accuracy_nn(pca_latent_seq, pca_labels_seq.ravel(), pca_latent_fish,
-                                pca_labels_fish.ravel())
+def all_benchmarks(n_epochs=250, use_cuda=True, save_path="data/", show_plot=True):
+    cortex_benchmark(
+        n_epochs=n_epochs, use_cuda=use_cuda, save_path=save_path, show_plot=show_plot
+    )
```
jeff-regier: Could we delete the whole `benchmark.py` file, and the unit tests that call any of the remain benchmarks here? We never really developed a complete set of benchmarks. Seems like we've converged on using notebooks for everything -- for better or worse.

ANazaret: I'm having a look to see if the tests covered by `benchmark.py` are redundant or can be moved somewhere else.

scvi/inference/__init__.py
```diff
@@ -1,29 +1,26 @@
...
+    "UnsupervisedTrainer",
+    "AdapterTrainer",
+    "JointSemiSupervisedTrainer",
+    "SemiSupervisedTrainer",
+    "AlternateSemiSupervisedTrainer",
+    "ClassifierTrainer",
+    "JVAETrainer",
+    "auto_tune_scvi_model",
+    "hyperopt_worker",
+    "launch_workers",
```
jeff-regier: Does the lint program (flake8) check for single quotes now? Otherwise it's going to be hard to maintain these coding standards.

scvi/models/modules.py
```diff
@@ -1,13 +1,52 @@
...
+from torch.nn import ModuleList
 
 from scvi.models.utils import one_hot
 
 
+def reparameterize_gaussian(mu, var):
+    return Normal(mu, var.sqrt()).rsample()
+
+
+class ConditionalInstanceNormalization(nn.Module):
```
jeff-regier: Did this end up helping? If not, should we still add it to the codebase?

scvi/models/modules.py
```diff
@@ -19,10 +58,23 @@ class FCLayers(nn.Module):
     :param n_layers: The number of fully-connected hidden layers
     :param n_hidden: The number of nodes per hidden layer
     :param dropout_rate: Dropout rate to apply to each of the hidden layers
+    :param n_batch_norm: Number of categories to conditionally batch normalize to (1 is
+             equivalent to vanilla batch norm, 0 no batch norm)
+             See `n_instances` in ConditionalInstanceNormalization for more information
+    :param use_batch_norm: Deprecated, use n_batch_norm=0 instead
```
jeff-regier: Better to delete it rather than having a "deprecate" note...we're not at v 1.0 so better to keep the code simple than to maintain backward compatibility.

tests/notebooks/utils/gimvi_tutorial.py
```diff
@@ -0,0 +1,104 @@
+import umap
```
jeff-regier: Can we move these the source directory (i.e., `scvi`). Or what's our thinking on what goes where?

scvi/models/modules.py
ANazaret: It was not helping much for GimVI, maybe it could be useful for further scVI researches. I don't have any preferences for the codebase.

tests/notebooks/utils/gimvi_tutorial.py
ANazaret: In the repo, the notebooks are used both as tests as tutorials/demonstrations. For sake of readability, I refactored all the `utils` function of the `gimvi_tutorial` notebook in this file, rather than having huge cells of code. 
I thought it would be easier for the user to have demo notebooks and their supplementary materials in the same folder. What do you think?

scvi/inference/__init__.py
gabmis: De are gradually introducing black which does!

ANazaret: Nope it doesn't enforce that, but `black` does and its usage is recommended in `CONTRIBUTING.rst`. I let the formatter only affect this file because the git history is not very important for this file. Tell me if you think I should revert the changes.

scvi/benchmark.py
ANazaret: I removed `benchmark.py`, though I found a mention of the file in the auto-generated docs. https://github.com/YosefLab/scVI/blob/48740851f32e1bee8d4e87f3b18a2bec9ec97e9c/docs/scvi.rst#L16
@jeff-regier , do you know how to update these files?

jeff-regier: `make docs` does it I think

tests/notebooks/utils/gimvi_tutorial.py
jeff-regier: There are already some functions in the scvi source code folder that seem like they could just as easily go into this new `utils` folder. I'm concerned about having more than one place where helper functions could go. Not sure what the solution is.

jeff-regier: I agree with not having huge cells of code though!

jeff-regier: Is this utils code only useful for this one notebook? Or could it be useful for other notebooks too?

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,385 @@
...
+
+            if self.use_cuda:
+                cls_z = cls_z.cpu()
+            cls_z = cls_z.numpy()
+
+            row = cls_z.mean(axis=0)
+            confusion.append(row)
+        return np.array(confusion)
+
+    def get_loss_magnitude(
```
gabmis: What is the meaning of magnitude here? Why not `get_total_loss` or `get_aggregate_loss` ?

tests/notebooks/utils/gimvi_tutorial.py
ANazaret: Yes these are really specific pieces of codes for this one notebook with seq/spatial questions. I don't think they should go into the scvi source folder until we found some other usages in the future.

scvi/models/jvae.py
```diff
@@ -0,0 +1,372 @@
...
+from scvi.models.log_likelihood import log_zinb_positive, log_nb_positive
+from scvi.models.modules import Encoder
+from scvi.models.modules import MultiEncoder, MultiDecoder
+from scvi.models.utils import one_hot
+
+torch.backends.cudnn.benchmark = True
+
+
+class JVAE(nn.Module):
+    r"""Joint Variational auto-encoder
```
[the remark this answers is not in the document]
ANazaret: Indeed we don't! Thanks.
Most of the `r"""` in the codebase are useless actually.

scvi/models/jvae.py
```diff
@@ -0,0 +1,372 @@
...
+            dim=1
+        )
+        kl_divergence = kl_divergence_z
+
+        if self.model_library_bools[mode]:
+            kl_divergence_l = kl(
+                Normal(ql_m, torch.sqrt(ql_v)),
+                Normal(local_l_mean, torch.sqrt(local_l_var)),
+            ).sum(dim=1)
+            kl_divergence += kl_divergence_l
```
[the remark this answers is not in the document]
ANazaret: Thank you for spotting that!

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,385 @@
...
+
+        self.compute_metrics()
+        begin = time.time()
+
+        with torch.set_grad_enabled(True):
+            self.model.train()
+            self.discriminator.train()
+
+            d_params = filter(
+                lambda p: p.requires_grad, self.discriminator.parameters()
```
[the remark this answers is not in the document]
ANazaret: I can but it's done this way in the codebase, and using `filter` make the code more understandable. Why do you want to get rid of this `lambda`? 
It won't cause any problem with `pickle` if that's your concern, because the lambda is not stored in memory.

scvi/inference/jvae_trainer.py
ANazaret: I introduced this function to track the different losses of this model: 
- reconstruction
- kl divergences
- adversarial
The main goal is to make sure they have the same order of magnitude (so that you don't have a losses that are actually never optimized because too small), so I called the function that way.
Does it make sense, or do you have another name in mind?

scvi/inference/jvae_trainer.py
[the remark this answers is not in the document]
gabmis: That was my concern!

scvi/inference/jvae_trainer.py
gabmis: So you don't use this for optimization purposes?
Maybe add a docstring then?

scvi/models/jvae.py
```diff
@@ -0,0 +1,372 @@
...
+        :param reconstruction_losses: list of distributions to use in the generative process 'zinb', 'nb', 'poisson'
+        :param model_library_bools: bool list: model or not library size with a latent variable or use observed values
+        :param n_latent: dimension of latent space
+        :param n_layers_encoder_individual: number of individual layers in the encoder
+        :param n_layers_encoder_shared: number of shared layers in the encoder
+        :param dim_hidden_encoder: dimension of the hidden layers in the encoder
+        :param n_layers_decoder_individual: number of layers that are conditionally batchnormed in the encoder
+        :param n_layers_decoder_shared: number of shared layers in the decoder
+        :param dim_hidden_decoder_individual: dimension of the individual hidden layers in the decoder
+        :param dim_hidden_decoder_shared: dimension of the shared hidden layers in the decoder
```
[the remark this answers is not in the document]
ANazaret: I think it could be a future improvement but I'm not sure it would really improve the model much compared to all the additional parameters it introduces. It would also be a broader change in the codebase, not only for GimVI.

I guess you should open an issue to change first `FCLayers` and eventually adapt the different models.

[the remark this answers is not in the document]
gabmis: Yep it's already there, another time then.

scvi/inference/jvae_trainer.py
ANazaret: Nope, all the `get_...` functions are for user inspection only.
I added docstrings

scvi/inference/jvae_trainer.py
```diff
@@ -0,0 +1,385 @@
...
+        n_classes = self.n_dataset
+        losses = []
+        for i, z in enumerate(latent_tensors):
+            cls_logits = nn.LogSoftmax(dim=1)(self.discriminator(z))
+
+            if predict_true_class:
+                cls_target = torch.zeros(
+                    n_classes, dtype=torch.float32, device=z.device
+                )
+                cls_target[i] = 1.0
```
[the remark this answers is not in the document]
ANazaret: Thanks for pointing that out, I was not that familiar with this. However for GimVI, I am not sure the adversarial training is the limiting factor of model. It will take time to benchmark with this again so I keep that in mind for future work but I won't add it in the code.

PR #420, opened 2019-07-16
Fixes #419

### LINKED ISSUE #419: Do not track test metrics if data is too small
The test metrics are always tracked if the user specifies `frequency`. However one might want to train the model without a test set, this would raise a ZeroDivision Error as in #418 

This line should be changed.

https://github.com/YosefLab/scVI/blob/48740851f32e1bee8d4e87f3b18a2bec9ec97e9c/scvi/inference/inference.py#L40

ANazaret:
Closes #419 
* Do not track metrics of a Posterior if it has less than 5 cells
* Add `nb_cells` attribute to Posterior, faster than `len(self.indices)`
* Remove code duplication in `GeneExpressionDataset.__len__`

### DIFF
--- a/scvi/dataset/dataset.py
+++ b/scvi/dataset/dataset.py
@@ -513,7 +513,7 @@ class GeneExpressionDataset(Dataset):
     #############################
 
     def __len__(self):
-        return self.X.shape[0]
+        return self.nb_cells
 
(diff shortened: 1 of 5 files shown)


ANazaret:
> LTGM

Let's To Go, Master!

PR #423, opened 2019-07-17
feat change inference output

### LINKED ISSUE #422: Replace `inference` outputs
I think it could be helpful to replace the output of `inference` (which is currently a tuple) by a dictionnary.
That way it would be easier to use other models (that do not necessarily have the exact same outputs) with `Posterior`. It would maybe also help gain in clarity.
What do you think?

PierreBoyeau:
Fixes #422


PierreBoyeau:
Done!

PR #424, opened 2019-07-18
Refactor the autotune code and add a tutorial and blog post reproducibility notebook

gabmis:
Seeing the amount of logs we were getting with the new configuration I added a `set_verbosity` function and a `_settings` module at top level of the `scvi` package.

closes #325 
closes #333 
closes #350 
closes #429

PR #425, opened 2019-07-18
Add validation set option

adamgayoso:
This is useful if one would like to do early stopping on a separate validation set.

### DIFF
--- a/scvi/inference/annotation.py
+++ b/scvi/inference/annotation.py
@@ -102,17 +102,29 @@ class ClassifierTrainer(Trainer):
         >>> trainer.test_set.accuracy()
     """
 
-    def __init__(self, *args, train_size=0.8, test_size=None, sampling_model=None, sampling_zl=False, use_cuda=True, **kwargs):
+    def __init__(
+        self,
+        *args,
+        train_size=0.8,
+        test_size=None,
+        sampling_model=None,
+        sampling_zl=False,
+        use_cuda=True,
+        **kwargs
+    ):
         self.sampling_model = sampling_model
         self.sampling_zl = sampling_zl
         super().__init__(*args, use_cuda=use_cuda, **kwargs)
-        self.train_set, self.test_set, self.validation_set = self.train_test_validation(self.model, self.gene_dataset,
-                                                                                    train_size=train_size,
-                                                                                    test_size=test_size,
-                                                                                    type_class=AnnotationPosterior)
-        self.train_set.to_monitor = ['accuracy']
-        self.test_set.to_monitor = ['accuracy']
-        self.validation_set.to_monitor = ['accuracy']
+        self.train_set, self.test_set, self.validation_set = self.train_test_validation(
+            self.model,
+            self.gene_dataset,
+            train_size=train_size,
+            test_size=test_size,
+            type_class=AnnotationPosterior,
+        )
+        self.train_set.to_monitor = ["accuracy"]
+        self.test_set.to_monitor = ["accuracy"]
+        self.validation_set.to_monitor = ["accuracy"]
         self.train_set.model_zl = sampling_zl
         self.test_set.model_zl = sampling_zl
         self.validation_set.model_zl = sampling_zl
--- a/scvi/inference/annotation.py
+++ b/scvi/inference/annotation.py
@@ -84,6 +84,8 @@ class ClassifierTrainer(Trainer):
         :gene_dataset: A gene_dataset instance like ``CortexDataset()``
         :train_size: The train size, either a float between 0 and 1 or and integer for the number of training samples
             to use Default: ``0.8``.
+        :test_size: The test size, either a float between 0 and 1 or and integer for the number of test samples
+            to use Default: ``None``.
         :sampling_model: Model with z_encoder with which to first transform data.
         :sampling_zl: Transform data with sampling_model z_encoder and l_encoder and concat.
         :\**kwargs: Other keywords arguments from the general Trainer class.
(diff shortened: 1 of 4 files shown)


adamgayoso:
@gabmis Does this force autotune to compute early stopping with respect to the dev set? If not we should enforce this.

scvi/inference/inference.py
```diff
@@ -31,13 +34,16 @@ class UnsupervisedTrainer(Trainer):
     """
     default_metrics_to_monitor = ['elbo']
 
-    def __init__(self, model, gene_dataset, train_size=0.8, test_size=None, n_epochs_kl_warmup=400, **kwargs):
+    def __init__(self, model, gene_dataset, train_size=0.8, test_size=0.15, n_epochs_kl_warmup=400, **kwargs):
         super().__init__(model, gene_dataset, **kwargs)
         self.n_epochs_kl_warmup = n_epochs_kl_warmup
         if type(self) is UnsupervisedTrainer:
-            self.train_set, self.test_set = self.train_test(model, gene_dataset, train_size, test_size)
+            self.train_set, self.test_set, self.dev_set = self.train_test(model, gene_dataset, train_size, test_size)
```
ANazaret: I think you changed the name of the function to `train_test_dev`

adamgayoso: yes, thanks for catching.

adamgayoso:
@ANazaret can you double check what I did to `jvae_trainer`?

jeff-regier:
Standard terminology is train, validate, and test. Probably better to stick to that.

ANazaret:
> Will the old notebooks still work well if another 10% of the dataset is removed from the training set?
> 
> It's probably fine but perhaps worth checking if the default behavior is to carve off some of the training set both for validation and for testing.

Indeed, some mock datasets only have a few cells and it's very likely that some tests can fail because the validation set would only contain 1 cells per batch (the `local_variance` computation would fail). But maybe not, just saying ...

gabmis:
> @gabmis Does this force autotune to compute early stopping with respect to the dev set? If not we should enforce this.

It should if it's the new default behaviour of `trainer.EarlyStopping`

scvi/inference/trainer.py
```diff
@@ -225,23 +226,26 @@ def __setattr__(self, name, value):
...
+
         :param train_size: float, int, or None (default is 0.1)
         :param test_size: float, int, or None (default is None)
         """
         model = self.model if model is None and hasattr(self, "model") else model
         gene_dataset = self.gene_dataset if gene_dataset is None and hasattr(self, "model") else gene_dataset
         n = len(gene_dataset)
         n_train, n_test = _validate_shuffle_split(n, test_size, train_size)
-        np.random.seed(seed=seed)
+        np.random.seed(seed=self.seed)
```
ANazaret: Hey @adamgayoso can you change this `seed` here. It has an impact on all the code. You can use: 
`random_state = np.RandomState(seed=self.seed)` and then `permutation = random_state.permutation(n)`

scvi/inference/annotation.py
```diff
@@ -100,17 +102,22 @@ class ClassifierTrainer(Trainer):
...
-        self.train_set, self.test_set = self.train_test(self.model, self.gene_dataset,
-                                                        train_size=train_size,
-                                                        type_class=AnnotationPosterior)
+        self.train_set, self.test_set, self.validate_set = self.train_test_validate(self.model, self.gene_dataset,
+                                                                                    train_size=train_size,
+                                                                                    test_size=test_size,
+                                                                                    type_class=AnnotationPosterior)
         self.train_set.to_monitor = ['accuracy']
-        self.test_set.to_monitor = ['accuracy']
+        if self.test_set.nb_cells > 0:
```
ANazaret: This lines are useless/wrong as even if `self.test_set.nb_cells = 0`, the metrics will still fail. 
I did a recent PR to prevent calculation of metrics at `compute_metrics` time.
So I think you can remove the if statement.

scvi/inference/annotation.py
```diff
@@ -100,17 +102,22 @@ class ClassifierTrainer(Trainer):
...
+            self.test_set.to_monitor = ['accuracy']
+        if self.validate_set.nb_cells > 0:
```
ANazaret: Same :)

scvi/inference/inference.py
```diff
@@ -35,9 +38,12 @@ def __init__(self, model, gene_dataset, train_size=0.8, test_size=None, n_epochs
         super().__init__(model, gene_dataset, **kwargs)
         self.n_epochs_kl_warmup = n_epochs_kl_warmup
         if type(self) is UnsupervisedTrainer:
-            self.train_set, self.test_set = self.train_test(model, gene_dataset, train_size, test_size)
+            self.train_set, self.test_set, self.validate_set = self.train_test_validate(model, gene_dataset, train_size, test_size)
             self.train_set.to_monitor = ['elbo']
-            self.test_set.to_monitor = ['elbo']
+            if self.test_set.nb_cells > 0:
```
ANazaret: Same :)

ANazaret:
> Standard terminology is train, validate, and test. Probably better to stick to that.

The standard is more train, validation, test.

adamgayoso:
> Where do you specifies to the code whether to track the validation or the test set (for metrics or earlystopping?). Because earlystopping should now rely on the validation set right. What if it's empty?

There are a number of things which rely on the test set, including early stopping and things inside notebooks. It's not a great solution here, but with this someone can just use the validation set as their test set.

ANazaret:
I see you forced-push one commit, do you think you can amend the commit message to have a cleaner one?

ANazaret:
> > Where do you specifies to the code whether to track the validation or the test set (for metrics or earlystopping?). Because earlystopping should now rely on the validation set right. What if it's empty?
> 
> There are a number of things which rely on the test set, including early stopping and things inside notebooks. It's not a great solution here, but with this someone can just use the validation set as their test set.

Yes that's what I understood, it's nice to have it but the test and validation test are swap compared to usual denomination.

ANazaret:
Do you want to change `autotune` behavior in this PR @adamgayoso or should I merge it?

gabmis:
> > Standard terminology is train, validate, and test. Probably better to stick to that.
> 
> The standard is more train, validation, test.

@ Let me take a quick look

gabmis:
> > > Standard terminology is train, validate, and test. Probably better to stick to that.
> > 
> > 
> > The standard is more train, validation, test.
> 
> @ Let me take a quick look

@adamgayoso @ANazaret  I have made quite a lot of changes to autotune in #424 , i'll add it in there or make a standalone PR afterwards.

PR #440, opened 2019-07-25
Fixes around autotune's unstable behaviour

### LINKED ISSUE #437: Tox a single test
Right now, tox runs all the test, we should be able to run a specific one.

### LINKED ISSUE #438: Unstable behaviour of the autotune tests

gabmis:
`auto_tune_scvi_model` was randomly failing when called multiple times with `parallel=True` because we terminated the `mongod` process but this process actually catches `SIGTERM` and does a few things before actually finishing. One of these things is to release the mongod.lock file which ensures there's only one MongoDb agent (`mongod` process) running per database. If this lock wasn't released by the time the next `auto_tune_scvi_model` lauched its Mongo agent, this agent would fail which would trigger connection errors in the workers and cause fmin to wait `fmin_timeout` after all workers have died, failing with a FminTimeoutError.

Now we wait for the `mongod` process to actually finish, additionally we test that it is actually running using `pymongo` five seconds after making the `subprocess.Popen(["mongod", ..])`.
Additionally, we output the logs of this process (originally stored in a file because verbose and messing with the autotune progress bar and logs) if the connection fails when we test it or after terminating the mongod process when behaviour is normal.

I fixed that and added a few changes which were convenient for debugging:
* let the `autotune` logs propagate all to the top so they could be caught by travis/tox
* add `{posargs}` positional argument in the first toxenv so that we can run a single test using e.g
`tox -e py37 -- tests/test_notebook.py::test_notebooks_autotune
* reduce the value of fmin_timeout in the notebooks (test mode only) so we don't have to wait too much if `auto_tune_scvi_model` fails that way

closes #438 
closes #437

### DIFF
--- a/scvi/_settings.py
+++ b/scvi/_settings.py
@@ -4,6 +4,34 @@ from typing import Union
 logger = logging.getLogger(__name__)
 scvi_logger = logging.getLogger("scvi")
 
+autotune_formatter = logging.Formatter(
+    "[%(asctime)s - %(processName)s - %(threadName)s] %(levelname)s - %(name)s\n%(message)s"
+)
+
+
+class DispatchingFormatter(logging.Formatter):
+    """Dispatch formatter for logger and it's sub logger."""
+
+    def __init__(self, default_formatter, formatters=None):
+        super().__init__()
+        self._formatters = formatters if formatters is not None else {}
+        self._default_formatter = default_formatter
+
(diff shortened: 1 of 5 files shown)


romain-lopez:
I see, looks good to me. Happy to see this fixed ! Way to go Gabriel !

PR #446, opened 2019-07-27
[WIP] avoid library saturation

PierreBoyeau:
Hi,

I am not sure this is the way to go, but this solution allowed me to avoid getting NaNs for some model, cause by the variance of the library encoder exploding for an example. Clamping the library size directly was a bad idea in my code, I am not sure about `master`.
Anyway, do you think controlling the library could be a good idea for scVI generally speaking?


jeff-regier:
It'd be nice if we could do this without creating a special case, i.e. without introducing a special `prevent_saturation` flag.

Issue #452, opened 2019-08-07
scVI 0.4.1 requires sklearn 0.21


Munfred:
When importing scVI (as below) with sklearn 0.20.0 installed I got the error `ImportError: cannot import name '_get_n_jobs'`

The solution was to upgrade to sklearn version 0.21.3, so the scVI requirements should be updated.

Import statement:
```
from scvi.inference import UnsupervisedTrainer
from scvi.inference.posterior import Posterior
```
Error message:
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-9-9cec25c261a2> in <module>()
     13 from scvi.dataset import GeneExpressionDataset
     14 from scvi.models import VAE
---> 15 from scvi.inference import UnsupervisedTrainer
     16 from scvi.inference.posterior import Posterior

~/.local/lib/python3.6/site-packages/scvi/inference/__init__.py in <module>()
----> 1 from .posterior import Posterior
      2 from .trainer import Trainer
      3 from .inference import UnsupervisedTrainer, AdapterTrainer
      4 from .annotation import (
      5     JointSemiSupervisedTrainer,

~/.local/lib/python3.6/site-packages/scvi/inference/posterior.py in <module>()
     13 from matplotlib import pyplot as plt
     14 from scipy.stats import kde, entropy
---> 15 from sklearn.cluster import KMeans
     16 from sklearn.manifold import TSNE
     17 from sklearn.metrics import adjusted_rand_score as ARI

~/.local/lib/python3.6/site-packages/sklearn/cluster/__init__.py in <module>()
      4 """
      5 
----> 6 from .spectral import spectral_clustering, SpectralClustering
      7 from .mean_shift_ import (mean_shift, MeanShift,
      8                           estimate_bandwidth, get_bin_seeds)

~/.local/lib/python3.6/site-packages/sklearn/cluster/spectral.py in <module>()
     15 from ..metrics.pairwise import pairwise_kernels
     16 from ..neighbors import kneighbors_graph
---> 17 from ..manifold import spectral_embedding
     18 from .k_means_ import k_means
     19 

~/.local/lib/python3.6/site-packages/sklearn/manifold/__init__.py in <module>()
      4 
      5 from .locally_linear import locally_linear_embedding, LocallyLinearEmbedding
----> 6 from .isomap import Isomap
      7 from .mds import MDS, smacof
      8 from .spectral_embedding_ import SpectralEmbedding, spectral_embedding

[...]

PR #467, opened 2019-09-05
[WIP] Refactor travis tests

### LINKED ISSUE #466: Testing notebooks only as side tests
Hi, 

Here are a few remarks about the testing issues of the codebase. @jeff-regier @adamgayoso what do you think ?

** TESTING **

We are having all the requirements and testing issues mostly because we rely on UMAP, igraph, scanpy etc.. for unit tests while I believe that only the codebase should be tested. This is especially problematic since the unit tests by travis need even to compile igraph etc... which honestly does not make much sense.

Still, trying out the notebooks is important to make sure our API is still compatible with scanpy and other software and that a wide range of people will run the tutorial without any trouble. 

Hence, here is an idea. How about having two different testing platforms, composed of:
+ unit tests
+ notebook tests

Only the unit tests are necessary to merge a branch to master. However, the notebook tests are still run on the background for each PR and there is simply a badge on the readme that indicates whether notebook tests pass or fail (as it is the case for the docs for example). 

When notebook fail, this will be the job of the maintainer to update things but at least for quick developments we will not run into these issues. 

** VERSION **

About the version system, it looks like the unit test would from now on only need the only strictly necessary packages to be installed while the notebook tests (background) would need to install extras. 

Consequently, it seems reasonable to have a requirements.txt file that has ONLY the necessary files to run the unit tests and NOT the notebooks.

setup.py will parse requirements.txt and have all the extras requirements to run the notebooks.

conda should either use setup.py or also install the extras. 

** README **

The readme should make it clear that the extras are needed to run the notebook examples. Also, it should make it clear that MongoDB is necessary whenever hyperopt is used.

PierreBoyeau:
Fixes #466


PierreBoyeau:
Still have work in progress as I do not know Travis very well

romain-lopez:
Hi @PierreBoyeau ! @adamgayoso made an amazing work on caching some data from travis that makes problem #466 obsolete. Still, we are left with issues #452. The work left to do is therefore:

+ Use only setup.py for requirement file and mention everything as main requirements
+ Remove pinned versions and solve compatibility with numba, umap and sklearn

Did you make some progress on the compatibility problem ? (especially numba 0.45.1 and sklearn 0.21) How do you recommend approaching this ?

PierreBoyeau:
Hi Romain, 
If I remember correctly all you mentionned above is done, except the numba compatibility issue. I'll work on that and ping you when it's done!

PR #473, opened 2019-09-18
Generalization of parameters for Anndata dataset API

fbrundu:
Generalization of parameters for extract_data_from_anndata API. It may be necessary because scanpy allows for the definition of different labels for batch, cell type etc. This update allows the user to define which is the label of the column containing the batch data, for example, in the original data.
Old values of previous code are retained as default, though.

### DIFF
--- a/scvi/dataset/anndataset.py
+++ b/scvi/dataset/anndataset.py
@@ -13,7 +13,12 @@ from scvi.dataset.dataset import DownloadableDataset, GeneExpressionDataset
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


adamgayoso:
This is a nice contribution, thank you! Please be sure to install the pre commit hook (or run black on the file), which should make Travis pass.

adamgayoso:
@fbrundu are you able to fix this PR? Please let us know! Thanks.

fbrundu:
Hi @adamgayoso, unfortunately at the moment I'm not able to work on it. If possible I'll update it when I can.

Thanks, Francesco

PR #479, opened 2019-10-03
Romain lopez/notebooks

romain-lopez:
worked on the main notebook to make it more informative for people and added scanpy use in that one.


adamgayoso:
Looks great! Can we also add how you would set the `train_size` and why you would pick anything less than 1?

Issue #434, opened 2019-07-24
Travis spend 65% of test time on `pip install`


ANazaret:
Most of the test time is lost on `pip install` which I believe spend 90% of the time building `leidenalg`, `python-igraph` and `louvain`.

It's not nice to wait that long to see a failing test that could be triggered in 1 minute.
Moreover these libraries are only use in notebooks, I'm not sure they are mandatory for the tests. We could just skip them in the notebook and keep the same coverage.

adamgayoso:
I address this in #482. We can still tests notebooks on the side but I think I mostly fixed the issue.

PR #482, opened 2019-10-04
cache pip for faster travis

adamgayoso:


adamgayoso:
Ok so pip install requirements is down to 140 secs from ~500 secs. I'm going to go ahead and merge this in.

romain-lopez:
Hi Adam, some questions:
(1) Does it detect changes or do you need to clear the cache manually ?
(2) Do you recommend we stick to testing also the notebook all the time and include all dependencies at install time ? (numba, umap, scanpy etc..) ? 
This will determine which issues to close

romain-lopez:
If this is a reasonable approach, we might want to keep testing the notebooks all the time. Notifying @PierreBoyeau .

adamgayoso:
1) it detects changes to the cache
2) I think we should tests notebooks all the time and figure out the bottlenecks in the notebooks - for example we might have some large datasets in our test data
3) we need to figure out how to unpin every version in requirements.txt

romain-lopez:
solves #434

romain-lopez:
solves #466

PR #484, opened 2019-10-07
LFC and DE

### LINKED ISSUE #487: Refactoring the differential expression module
I did a WIP PR for the LFC/DE estimation based on current code (my code looked nothing like that)
More precisely I simplified (I hope) the codebase for DE generally speaking. I will take care of IAF later on (which will require carefulness to as it optimizes another formulation of the ELBO) and I added two very simple functions:
- `estimate_change` computes the posterior properties of the change variable (LFC or used specified)
- `estimate_de_probability` computes the probability of being DE with our LFC formulation with threshold \delta.

As a reminder, the core function based on Bayes factors is in `differential_expression_score`. I did not touch that method, except to refactor it slightly. I am pretty sure a lot can be done/changed/simplified but I just wanted to get your opinion right now on several points.

1. While `all_stats` used by the BF version makes sense to me (it is nice to provide side informations on genes in addition to de scores), I am not sure the `bayes1_permuted` `bayes2` `bayes2_permuted` scores are really useful and I think we could get rid of those.
2. `differential_expression_score` and `estimate_de_probability` should somehow be merged. A complication is that outputs will potentially be different. Maybe returning a pandas DataFrame all the time could be a solution?
3. I was also wondering about the following thing: currently we sample `n_sample` times from the aggregated posterior. While I like this formulation (because it is quite staightforward), I used in the code for my article something similar to `n_samples_per_cell` argument of modified function `sample_scale_from_batch`. I don't know if this parameter makes sense, but if we don't include it I think all de related methods should inform the user of the number of time each cell posterior is sampled from for instance, and warn him if it is not sampled from enough
4. I think we could move all functions in the end of `posterior.py` (which are not `Posterior` methods) in another file (`posterior_utils.py` ?) for clarity. What do you think?
Anyway, I just wanted to ping you about the WIP PR and get your point of view on what straightforward modifications/simplifications we should do for the DE part of scVI

### LINKED ISSUE #505: `linear_assignment` depreciation
`linear_assignment` (from `sklearn.utils.linear_assignment_`) used in `posterior.py` is deprecated and should be replaced by `scipy`'s `linear_sum_assignment`

PierreBoyeau:
Fixes #487 
Fixes #505
- Refactors `Posterior` differential expression
- Creates of `posterior_utils.py` to lighten `posterior.py`
- Replaces deprecated `sklearn` function for linear assignment by `scipy`'s
- Generalizes `differential_expression_score` and other DE methods to LFC-based DE
- Simplifies `px_scale` sampling in `Posterior`


adamgayoso:
@PierreBoyeau what's the status with this? Were you able to address what we discussed?

PierreBoyeau:
Yes! I replaced the hypothesis issue and batch big we mentioned!

scvi/inference/posterior.py
```diff
@@ -273,176 +286,416 @@ def differential_expression_stats(self, M_sampling=100):
...
+        return dict(scale=px_scales, batch=batch_ids)
+
+    def get_bayes_factors(
         self,
         idx1: Union[List[bool], np.ndarray],
         idx2: Union[List[bool], np.ndarray],
+        mode: Optional[str] = "vanilla",
         batchid1: Optional[Union[List[int], np.ndarray]] = None,
         batchid2: Optional[Union[List[int], np.ndarray]] = None,
         genes: Optional[Union[List[str], np.ndarray]] = None,
```
adamgayoso: I'm not so sure about this genes parameter. It doesn't cut down on any computational cost and is an issue for totalVI because we can use this function but not this genes parameter. I think we should remove it and users can filter genes downstream

scvi/inference/posterior.py
```diff
@@ -273,176 +286,416 @@ def differential_expression_stats(self, M_sampling=100):
...
+        n_samples: Optional[int] = None,
+        n_samples_per_cell: Optional[int] = None,
+        batchid: Optional[Union[List[int], np.ndarray]] = None,
+        genes: Optional[Union[List[str], np.ndarray]] = None,
+        selection: Union[List[bool], np.ndarray] = None,
+    ) -> dict:
+        """
+        :param n_samples: Number of samples in total per batch (fill either `n_samples_total`
+        or `n_samples_per_cell`)
+        :param n_samples_per_cell: Number of time we sample from each observation per batch
```
adamgayoso: I thought you were removing `n_samples_per_cell`?

scvi/inference/posterior_utils.py
```diff
@@ -0,0 +1,291 @@
...
+def proximity_imputation(real_latent1, normed_gene_exp_1, real_latent2, k=4):
+    knn = KNeighborsRegressor(k, weights="distance")
+    y = knn.fit(real_latent1, normed_gene_exp_1).predict(real_latent2)
+    return y
+
+
+def pairs_sampler(
+    arr1: Union[List[float], np.ndarray, torch.Tensor],
+    arr2: Union[List[float], np.ndarray, torch.Tensor],
+    sample_pairs: bool = True,
```
adamgayoso: It's confusing that this function is called `pairs_sampler` and that there's a `sample_pairs` option. Is it possible to clarify this? Like, `sample_pairs` activates the permutation, so why not call it `use_permutation`?

scvi/inference/posterior.py
adamgayoso: if anything, the genes subsetting should occur in `differential_expression_score`

adamgayoso:
@PierreBoyeau this is looking really nice and I think in the future where we have more conditions it won't be too hard to adapt this code.

scvi/inference/posterior.py
```diff
@@ -273,176 +286,416 @@ def differential_expression_stats(self, M_sampling=100):
...
+        m1_domain_fn: Optional[Callable] = None,
+        delta: Optional[float] = 0.5,
+    ) -> dict:
+        r"""
+        Unified method for differential expression inference.
+        # FUNCTIONING
+        Two modes coexist:
+            - the "vanilla" mode follows protocol described in arXiv:1709.02082
+            In this case, we perform hypothesis testing based on:
+                M_1: h_1 > h_2
```
romain-lopez: m1 m2 or m0 m1 ?

scvi/inference/posterior.py
```diff
@@ -229,6 +233,15 @@ def entropy_batch_mixing(self, **kwargs):
 
     entropy_batch_mixing.mode = "max"
 
+    def update_sampler_indices(self, idx: Union[List, np.ndarray]):
+        """
+        Updates the dataloader indices
```
adamgayoso: can we provide a sentence or two about what this is used for?

scvi/inference/posterior.py
```diff
@@ -273,125 +286,375 @@ def differential_expression_stats(self, M_sampling=100):
...
+            DE can then be based on the study of the Bayes factors:
+            log (p(M_1 | x_1, x_2) / p(M_2 | x_1, x_2)
+
+            - the "change" mode (described in bioRxiv, 794289)
+            consists in estimating an effect size random variable (e.g., log fold-change) and
+            performing Bayesian hypothesis testing on this variable.
+            The `change_fn` function computes the effect size variable r based two inputs
+            corresponding to the normalized means in both populations
+            Hypotheses:
+                M_1: r \in R_0 (effect size r in region inducing differential expression)
```
adamgayoso: R_1?

scvi/inference/posterior.py
```diff
@@ -273,125 +286,375 @@ def differential_expression_stats(self, M_sampling=100):
...
         To that purpose we sample the Posterior in the following way:
             1. The posterior is sampled n_samples times for each subpopulation
             2. For computation efficiency (posterior sampling is quite expensive), instead of
-                comparing element-wise the obtained samples, we can permute posterior samples.
+                comparing the obtained samples element-wise, we can permute posterior samples.
                 Remember that computing the Bayes Factor requires sampling
                 q(z_A | x_A) and q(z_B | x_B)
 
+        # BATCH HANDLING
+
```
adamgayoso: are you putting something here?

scvi/inference/posterior.py
```diff
@@ -273,125 +286,375 @@ def differential_expression_stats(self, M_sampling=100):
...
+                    use_permutation=use_permutation,
+                    M_permutation=n_samples_per_batch,
+                )
+                scales_1.append(scales_1_local)
+                scales_2.append(scales_2_local)
+            scales_1 = np.concatenate(scales_1, axis=0)
+            scales_2 = np.concatenate(scales_2, axis=0)
+        else:
+            # As it is, this function might work poorly when the different batch ids differ
+            warnings.warn("batchid1 != batchid2: batches are ignored")
```
adamgayoso: is it fair to say "batches are ignored"? Even if the sets aren't equal you still "put" them into the batches in the argument to this function. for example if batchid1 = (1, 2) and batchid2 = (3, 4), you're still comparing group 1 in batches 1 and 2 to group 2 in batches 3, 4. Say you want to test control versus condition and you have two control batches (1 and 2) and two condition batches (3 and 4). This seems to be doing what we'd want. 

I'd give the warning (also should we be using logger warning?) but maybe say exactly what happens in this case.

adamgayoso:
@romain-lopez can we remove `differential_expression_stats()`? I don't see anything that uses it in this repo, and it looks like @chenlingantelope uses a hacked version of it in the scanVI reproducibility?

One could use `get_sample_scale()`, which does the same as the `differential_expression_stats()`.

scvi/inference/posterior.py
PierreBoyeau: I think the current status is acceptable. I just left an optional `n_samples_per_cell` that can replace `n_samples` just in case. Do you think I should remove it?

PR #485, opened 2019-10-08
[WIP] Add AutoZI to the codebase

oscarclivio:
**Done so far (as of 10/9)**:

- Reimplemented AutoZI using the latest version of the codebase : VAE, trainer, functions in `log_likelihood.py` : AutoZI's VAE class inherits from the standard VAE class, UnsupervisedTrainer is used for both AutoZI and scVI.
- Tested it successfully on reproducibility procedures (to be shown in separate repo as in the AutoZI bioRxiv preprint)
- Added tutorial notebook

**Important:** the loss used in `UnsupervisedTrainer` for scVI averages the ELBO wrt cells instead of summing them. In AutoZI I had averaged the same ELBOs but then also multiplied them by the size of the training set to keep proportions compared to the KL divergence wrt global parameters. An alternative which would have generalized the loss for scVI would be to divide the KL divergence wrt global parameters by this size instead. But doing this did not make me recover results from the preprint yet. Instead, `UnsupervisedTrainer` now gives the choice between multiplying the ELBO wrt cells by the size of the training set or dividing the KL divergence wrt global parameters by this size and by default it is set as True when scVI is used and False when AutoZI is used, so results for both models are not disturbed.


scvi/inference/autozi_trainer.py
```diff
@@ -0,0 +1,28 @@
...
+            self.train_set, self.test_set, self.validation_set = self.train_test_validation(
+                model, gene_dataset, train_size, test_size
+            )
+            self.train_set.to_monitor = ["elbo"]
+            self.test_set.to_monitor = ["elbo"]
+            self.validation_set.to_monitor = ["elbo"]
+
+    def loss(self, tensors):
+        sample_batch, local_l_mean, local_l_var, batch_index, y = tensors
+        reconst_loss, kl_divergence, kl_divergence_global =\
```
romain-lopez: this kl_divergence global is needed only for AutoZI but could be 0 by default in VAE. In that case, you don't need a new trainer

oscarclivio: Indeed, sounds much better

scvi/models/autozivae.py
```diff
@@ -0,0 +1,283 @@
...
+                bernoulli_params_res = []
+                for sample in range(bernoulli_params.shape[0]):
+                    bernoulli_params_res.append(F.linear(one_hot_batch, bernoulli_params[sample]))
+                bernoulli_params = torch.stack(bernoulli_params_res)
+
+        return bernoulli_params
+
+
+
+    def inference(self, x, batch_index=None, y=None, n_samples=1, eps_log=1e-8):
```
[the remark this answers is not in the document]
oscarclivio: Looks like scANVI just calls parent inference in its forward and does the rest of the job in the latter. In fact we still may have to implement a new inference method as the dropout is rescaled and we may want to obtain it directly rescaled *but* we can easily call the inference from the VAE parent, and just modify the dropout accordingly (and add sampled Bernoulli parameters, may look better in log_likelihood.py functions)

scvi/models/log_likelihood.py
```diff
@@ -36,6 +36,36 @@ def compute_elbo(vae, posterior, **kwargs):
...
+    plus a term that is constant with respect to the variational distribution.
+    It still gives good insights on the modeling of the data, and is fast to compute.
+    """
+    # Iterate once over the posterior and compute the elbo
+    elbo = 0
+    for i_batch, tensors in enumerate(posterior):
+        sample_batch, local_l_mean, local_l_var, batch_index, labels = tensors[
+            :5
+        ]  # general fish case
+        reconst_loss, kl_divergence, kl_divergence_bernoulli = autozivae(
```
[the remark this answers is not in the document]
oscarclivio: Everything looks easily factorizable except `log_likelihood` though. Remember we do not perform Monte-Carlo in the same way : loop on cells then on MC samples for VAE, loop on MC samples then on cells for AutoZI.

oscarclivio:
Thanks Romain and Adam, now working on it!

scvi/inference/inference.py
```diff
@@ -41,28 +41,48 @@ def __init__(
...
+        self.normalize_loss = (
+            not (
+                hasattr(self.model, "reconstruction_loss")
+                and self.model.reconstruction_loss == "autozinb"
+            )
+            if normalize_loss is None
+            else normalize_loss
+        )
+
+        self.n_scale = 1.0
```
romain-lopez: Can you comment what this is what why this has to take different values (testing set and training set) ? Maybe refer to the equation in your paper or the AEVB paper ?

scvi/models/autozivae.py
```diff
@@ -0,0 +1,310 @@
...
+            alpha_prior = 1.0 - beta_prior
+
+        device = (
+            torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
+        )
+
+        # Create parameters for Bernoulli Beta prior and posterior distributions
+        # Each paramer, whose values are in (0,1), is encoded as its logit, in the set of real numbers
+
+        if self.dispersion == "gene":
```
romain-lopez: Why is this using the argument dispersion ? This should be a separate parameter called ZI_design or something similar ?

scvi/models/autozivae.py
```diff
@@ -0,0 +1,310 @@
...
+                if alpha_prior is None
+                else torch.Tensor([logit(alpha_prior)]).to(device)
+            )
+            self.beta_prior_logit = (
+                torch.nn.Parameter(torch.randn(1, self.n_labels))
+                if beta_prior is None
+                else torch.Tensor([logit(beta_prior)]).to(device)
+            )
+
+        else:  # gene-cell
```
romain-lopez: This does not make sense then ?

scvi/models/jvae.py
```diff
@@ -308,6 +308,9 @@ def decode(
 
         return px_scale, px_r, px_rate, px_dropout
 
+    def compute_global_kl_divergence(self):
```
romain-lopez: this does not look like it is called anywhere ?

scvi/models/vae.py
```diff
@@ -232,6 +233,9 @@ def inference(self, x, batch_index=None, y=None, n_samples=1):
             library=library,
         )
 
+    def compute_global_kl_divergence(self):
+        return 0.0
```
romain-lopez: same here ?

scvi/models/log_likelihood.py
```diff
@@ -132,6 +212,7 @@ def log_zinb_positive(x, mu, theta, pi, eps=1e-8):
     theta: inverse dispersion parameter (has to be positive support) (shape: minibatch x genes)
     pi: logit of the dropout parameter (real support) (shape: minibatch x genes)
     eps: numerical stability constant
+    return_gene_specific: sum on genes if False, return the log likelihood per gene if True
```
romain-lopez: you might run into a minor conflict with this once we merge Adam's TotalVI first

romain-lopez:
Ah, and the notebook is missing in order to complete the PR. The notebook will augment the coverage (has to be unit tested) and should be short as we discussed.

scvi/models/jvae.py
oscarclivio: We do call it in `compute_elbo` in its current refactorized form

scvi/models/vae.py
oscarclivio: We do call it in `compute_elbo` in its current refactorized form

scvi/models/autozivae.py
oscarclivio: Indeed, this was a bad confusion. I just chose `zero_inflation` to be consistent with `dispersion`.

oscarclivio:
Thanks for the feedback @romain-lopez , everything should have been addressed, except `compute_global_kl_divergence` in models different from AutoZI : we call this function in `compute_elbo`.

scvi/models/autozivae.py
```diff
@@ -0,0 +1,312 @@
...
+        self.minimal_dropout = minimal_dropout
+
+        # Parameters of prior Bernoulli Beta distribution : alpha + beta = 1 if only one is specified
+        if beta_prior is None and alpha_prior is not None:
+            beta_prior = 1.0 - alpha_prior
+        if alpha_prior is None and beta_prior is not None:
+            alpha_prior = 1.0 - beta_prior
+
+        device = (
+            torch.device("cuda") if torch.cuda.is_available() else torch.device("cpu")
```
adamgayoso: this should not be necessary these lines. See line 86 in trainer.py

scvi/models/log_likelihood.py
```diff
@@ -172,6 +253,7 @@ def log_nb_positive(x, mu, theta, eps=1e-8):
     mu: mean of the negative binomial (has to be positive support) (shape: minibatch x genes)
     theta: inverse dispersion parameter (has to be positive support) (shape: minibatch x genes)
     eps: numerical stability constant
+    return_gene_specific: sum on genes if False, return the log likelihood per gene if True
```
adamgayoso: please be sure to remove this `return_gene_specific` as it is now default

scvi/models/log_likelihood.py
```diff
@@ -33,6 +33,8 @@ def compute_elbo(vae, posterior, **kwargs):
         )
         elbo += torch.sum(reconst_loss + kl_divergence).item()
     n_samples = len(posterior.indices)
+    kl_divergence_global = vae.compute_global_kl_divergence()
```
adamgayoso: 1) I don't see why you need to call this here because shouldn't it be the term you returned by calling forward?
2) Doesn't the global KL need to be divided by sample size?

scvi/inference/inference.py
```diff
@@ -41,28 +41,51 @@ def __init__(
         train_size=0.8,
         test_size=None,
         n_epochs_kl_warmup=400,
+        normalize_loss=None,
```
adamgayoso: this should be a bool

scvi/models/log_likelihood.py
adamgayoso: so yeah, it looks like you should do the same normalization here as you do in the loss function.

adamgayoso:
Also, @oscarclivio, would you be able to add typing to your model file? See Achille's code or my `totalvi.py`. Thanks!

scvi/models/autozivae.py
oscarclivio: In fact, model.cuda() was not putting the prior alpha and beta on CUDA when they are fixed and when applying model.cuda() in trainer.py. But if we ever choose `use_cuda=False` when a GPU is available, we would obtain prior alpha and beta on CUDA and the rest on CPU. The best is to make a new `cuda` method tackling the discrepancy with prior alpha and beta directly.

scvi/models/log_likelihood.py
oscarclivio: 1. I thought it was ugly to use the global KL divergence returned by calling forward as it should always be the same and can be called only once, but doing so tackles Romain's previous point about unused `compute_global_kl_divergence` so this is the best idea.
2. We do divide it by sample size when we return `elbo / n_samples`

scvi/inference/inference.py
oscarclivio: Are we sure we want to fix it to a bool at default value ? So far, scVI only relied on `normalize_loss=True` and all results from AutoZI we now have rely on `normalize_loss=False`. `normalize_loss=None` by default and the processing done for this case helps us treat the two different values for scVI and AutoZI at the same time.

adamgayoso: can you just clarify in the docstring what the options are?

scvi/models/autozivae.py
```diff
@@ -0,0 +1,357 @@
...
+from scvi.models.vae import VAE
+from scipy.special import logit
+from scvi.models.utils import one_hot
+
+from typing import Dict, Optional, Tuple, Union
+
+torch.backends.cudnn.benchmark = True
+
+
+class AutoZIVAE(VAE):
```
adamgayoso: please add docstring here explaining your model and its parameters

scvi/models/autozivae.py
```diff
@@ -0,0 +1,357 @@
...
+            self.beta_prior_logit = (
+                torch.nn.Parameter(torch.randn(1, self.n_labels))
+                if beta_prior is None
+                else torch.Tensor([logit(beta_prior)])
+            )
+
+        else:  # gene-cell
+            raise Exception("Gene-cell not implemented yet for AutoZI")
+
+    def cuda(self, device: Optional[str] = None) -> torch.nn.Module:
```
adamgayoso: docstring here too please

adamgayoso:
@oscarclivio it looks like travis takes about 2 minutes longer with the branch. Are the test datasets you added very large?

PR #488, opened 2019-10-09
Fix requirements

PierreBoyeau:


setup.py
```diff
@@ -10,10 +10,12 @@
     history = history_file.read()
 
 requirements = [
+    "pip>=18.1"
     "numpy>=1.16.4",
     "torch>=0.4.1",
+    "Sphinx>=1.7.1"
     "matplotlib>=2.0",
-    "scikit-learn>=0.18, <0.20.0",
+    "scikit-learn>=0.18",
```
romain-lopez: apparently there is this bug on sklearn 0.20 ? Should we update all the minimal version on this list to be the exact same there is on the previous requirements.txt ? At least we know these one work. For example, sklearn can be >= 0.20.3 ? We can do the same for all the other packages.

setup.py
```diff
@@ -28,16 +30,18 @@
     "umap-learn>=0.3.7",
     "seaborn>=0.9.0",
     "hyperopt>=0.1.2",
-    "numba>=0.37, <0.45.1",
+    "numba>=0.37",
```
romain-lopez: maybe we need to != 0.45.1 here. It would be nice to check tox installs 0.46 and make sure this one works for UMAP. Maybe we should pin down with a comment that umap fails with this version of numba.

romain-lopez:
Also, I recommend using tox --recreate (in a python 3.7 virtual env). When this is done, I can try on my side an push this online to pypi with a new version :)

adamgayoso:
One thing we should do here is split out the reqs for documentation like scanpy does in their setup.py. There's no reason to req packages that are only used to build docs.

PierreBoyeau:
The queue issue we are all having is caused by some loom/numba incompatibility that causes an error in the `Custom objective function` section of the autotune notebook.
I'm trying to understand more precisely what happens

romain-lopez:
I see, so numba 0.45.1 is  incompatible with UMAP and numba 0.46.0 is incompatible with loompy in this parallel scenario. For this probably we can add one comment to the notebook, explaning here numba 0.45.0 is needed and disable parallel hoping only this one will not work ? Also, we can add in our setup.py numba >= 0.45.0 but numba != 0.45.1 so that people do not get their good numba uninstalled ?

.travis.yml
```diff
@@ -3,12 +3,15 @@ dist: xenial
 
 language: python
 
+services:
+  - mongodb
```
romain-lopez: is this useful or 
+ should we add pymongodb to tox 
or 
should we ask people to install mongodb on their laptop ?

This is important for the installation instructions on the readme

.travis.yml
```diff
@@ -17,7 +20,8 @@ notifications:
   email: false
 
 install:
-  - pip install --quiet -r requirements.txt
+  - pip install --quiet -e ".[notebooks]"
```
romain-lopez: should we just put all the requirements in the setup.py with no extras ?

romain-lopez: and use here pip install --quiet .

.travis.yml
PierreBoyeau: I think you could just get rid of it!

.travis.yml
PierreBoyeau: I like the idea of having a "simple" version of scVI, and another one which includes almost all the tools needed to run the notebooks.
It looks like installing the scVI package with the extras would be really easy (scanpy does something similar):

```
pip install scvi[notebooks]
```
That being said I can get rid of this distinction if you want (full installation is not that long and introducing extras may be confusing to the users?)

adamgayoso: I think the regular pip install should have the minimum number of requirements to make scVI work. We should have these extras for stuff required in the notebooks, testing, and building the docs.

In the readme we can add the different pip install commands. scVI technically does not require scanpy, louvian, etc. so it's a bit weird to force installation of that even if they are extremely useful downstream of scVI.

PR #491, opened 2019-10-14
fix suffix determination in dataset10X

adamgayoso:
if there are hidden files like `.temp.file` it messes with the suffix determination `.gz`, for example.


PierreBoyeau:
Looks good to me!

PR #495, opened 2019-10-17
don't make gene names uppercase

### LINKED ISSUE #450: populate_from_data edits gene_names
While trying to debug why a DE gene found using scvi could not be found in the original anndata object, I realized that `GeneExpressionDataset.populate_from_data()` called by `AnnDatasetFromAnnData` would capitalize all the gene names. Just wanted make a note of this in case anyone else has this problem. It makes senes to make sure the gene_names are of `str` type. But it does not seem to be necessary and can be confusing to capitalize gene_names in the backend, when the user provides the dataset with given gene names. 

``` python
        if gene_names is not None:
            self.initialize_gene_attribute(
                "gene_names", np.char.upper(np.asarray(gene_names, dtype="<U64"))
            )
```

### LINKED ISSUE #459: make gene names unique
We don't check if gene names are unique, which can pose a problem when merging datasets. We should look into how Seurat or scanpy handle this issue.

adamgayoso:
- Fixes #450 
- Fixes #459 
- Adds a new method `make_gene_names_lower()` to the dataset class
- Removes all instances of gene names case being changed throughout the repo


ANazaret:
Hi Adam, why don't you want to make gene uppercase? We need some sort of standardization because for spatial-sequencing experiments (GimVI for instance) then we match the genes between two datasets by their names. If the gene names are not all uppercase then the matching will fail.

adamgayoso:
> Hi Adam, why do you want to make gene uppercase? We need some sort of standardization because for spatial-sequencing experiments (GimVI for instance) then we match the genes between two datasets by their names. If the gene names are not all uppercase then the matching will fail.

I think the idea is that there are standards to gene names that people expect, and making them uppercase breaks that. I think we need to do a bit more research before implementing these changes.

adamgayoso:
@ANazaret how do you feel if the user uppercases the gene names themselves in each dataset object before merging?

ANazaret:
> @ANazaret how do you feel if the user uppercases the gene names themselves in each dataset object before merging?

I think they won't do it and it can fail. Probably the better thing is to change directly the function matching the names so that this function is not sensitive to the case.
That is replacing `g` by `g.lower()` for instance in this snippet. 
https://github.com/YosefLab/scVI/blob/05920d1f85daa362d4fb694e588ab090bc84e207/scvi/dataset/cortex.py#L73-L75

mukundvarma:
Is this PR going to be merged? Upper-casing gene names is an issue when working with non-human datasets.

adamgayoso:
@ANazaret I think we will add an option in `populate_from_data` that uppercases the gene names with the default being no uppercase. 

There will be some instances where this is not desirable, but downstream analysis (DE for example) is much more difficult when gene names are changed. We could do what you suggested but then how would we set the gene_names attribute of a concatenated dataset?

mukundvarma:
@adamgayoso 

I feel like changing feature names is something that should be done outside of a machine learning model, so it makes sense to me for this not to be a method on the model class, and instead be something users do on their own prior to dataset concatenation or alignment.

adamgayoso:
@romain-lopez any thoughts on this?

ANazaret:
Hey Adam,

I still think what I suggested in https://github.com/YosefLab/scVI/pull/495#issuecomment-556097875 could be a nice behavior.

The changes would be:
- [ ] **Remove** the `.upper()` in the different dataset files
- [ ] **Add a flag** `case_sensitive=False` in `dataset.reorder_genes` and in the constructors of the datasets accepting a ``genes_to_keep`` argument; such as `CortexDataset` or `FrontalCortexDropseqDataset` in loom.py
- [ ] **Change the behavior** of the previous functions to work accordingly: if `case_sensitive==True` then it should be the same code -- since now upper was removed -- and else we should do `.upper` only in the look_up dictionaries we use to match the gene names (see code below).

This way, gene names will never be changed but the `gene_to_keep` arguments will have a very handful argument to be case insensitive.

https://github.com/YosefLab/scVI/blob/c1299048cd25398d6df65547f7dcf3fba52a4d7d/scvi/dataset/dataset.py#L937-L938

adamgayoso:
I disagree that we should allow the code to be invariant to gene name capitalization as it seems like bad practice.

We could just throw a warning in populate from datasets if there are very few shared gene names suggesting the user checks gene names themselves .

ANazaret:
@adamgayoso I let you decide since you know more about the casual usage of gene names. 

However the uppercasing was introduced in the repo to avoid problems with case sensitive matching. With your approach, if two datasets have different case (notably some dataset with capitalized or uppercased genes), we will have to upper case the gene names with `make_gene_names_upper` although it was specifically the **thing we wanted** to avoid. Thus I still think a flag `case_sensitive: bool` is more appropriate since it will fix the original issue. Or rather use `make_gene_names_lower` since upper seems to be a problem.

In the end, the most important is to get the PR merged since some people need this to be fixed fast. Thanks for your changes and feel free to change at least the make_gene_lower!

adamgayoso:
@ANazaret I'll change it to lower. 

I think the solution proposed in the code is more generally suitable. Users will only have to make gene names have the same case in advanced usage situations.

romain-lopez:
I think we should at the very least do two things. (A) let the user know whether we changed gene names for merging or any other reasons, that should be done via logging (B) that should not be done by default. 

Achille's rule might not be perfectly suitable since even once you merged the gene names in a case unsensititive way, you need to pick one of the two gene names. So Adam's solution is surely arbitrary but not too bad as long as (A) we inform the user and (B) this is done exceptionally and not the default behavior

vals:
Not sure how you're doing with this, but spent some time wrestling with the consequences of this today looking at some mouse data...

In cases of e.g. integrating proteins and genes, or data from different organisms, which are cases where linked molecules have different capitalization/names, I think it needs to be up to the user to define the names in a consistent way.

PR #497, opened 2019-10-20
Ldvae updates

adamgayoso:
- Updates ldvae code to use more built in modules, adds "ln" latent distribution.


adamgayoso:
I know I will need to add a test for the "ln" dist

scvi/models/vae.py
```diff
@@ -337,9 +361,22 @@ def __init__(
...
             n_layers=n_layers,
             n_hidden=n_hidden,
+            use_batch_norm=use_batch_norm,
         )
 
     def get_loadings(self):
         """ Extract per-gene weights (for each Z) in the linear decoder.
         """
-        return self.decoder.factor_regressor.parameters()
+        with torch.no_grad():
```
romain-lopez: can't we use @torch.nograd ?

scvi/models/modules.py
```diff
@@ -266,35 +278,41 @@ def __init__(
         n_cat_list: Iterable[int] = None,
         n_layers: int = 1,
```
romain-lopez: Should we add in the doc the exact value that makes it linear ? If n_layers > 1 then it wouldn't be linear right ? Should we otherwise remove n_layers from this ?

tests/notebooks/linear_decoder.ipynb
```diff
@@ -23,17 +23,9 @@
    "cell_type": "code",
```
romain-lopez: <p>We should give credit here to Valentine. The best way to me seems to cite their paper :)</p><p>https://www.biorxiv.org/content/biorxiv/early/2019/08/16/737601.full.pdf</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/YosefLab/scVI/pull/497/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

scvi/models/modules.py
adamgayoso: I will rename to `n_layers_encoder` because decoder is hardcoded.

adamgayoso:
@vals how do you want to set the defaults?

tests/notebooks/linear_decoder.ipynb
adamgayoso: I'll add to the readme instead.

vals:
Hey @adamgayoso I just reproduced my main analysis with the new API.

So far the defaults looks all right, at least given the performance evaluation you had done.

I'll do some further exploration in the latent spaces with the different settings. I couldn't immediately see much structure with the default settings, but only looked at a couple of dimensions. Without batch-norm I saw some more clear cell populations. I'm curios to see what tSNE's of the ln-distributed latent space looks like.

adamgayoso:
@vals it could also be interesting to look at W in the `use_batch_norm=True` setting, without looking at BW. After training with `use_batch_norm=True`, you could set `use_batch_norm=False` before calling the loadings function just to see W. Also, for ln-distributed, a 2-d PCA of the full latent space looks cool.

PR #513, opened 2019-11-15
add tqdm for one vs all, use debug instead of info for get bayes factors

adamgayoso:


adamgayoso:
@PierreBoyeau what's the implication of using the warnings we implemented versus logger.warning?

PierreBoyeau:
@adamgayoso I used `warnings` without thinking it through. Based on what I could find on the official python docs:

"`warnings.warn()` in library code if the issue is avoidable and the client application should be modified to eliminate the warning

`logging.warning()` if there is nothing the client application can do about the situation, but the event should still be noted"

Based on that, I think our warnings make sense, right?

PR #514, opened 2019-11-16
feat transformed_batch option for imputation

### LINKED ISSUE #512: Extend `Posterior`'s `imputation`
#484 now provides an easy way to condition on a fixed batch in `VAE`. Based on discussion with Adam, we could extend the features of `imputation` so that conditioning could be performed on a specific (list of) batch(es) with the new implementation.

PierreBoyeau:
Fixes #512

### DIFF
--- a/scvi/inference/posterior.py
+++ b/scvi/inference/posterior.py
@@ -1003,16 +1003,37 @@ class Posterior:
         return de_res, de_cluster
 
     @torch.no_grad()
-    def imputation(self, n_samples=1):
-        imputed_list = []
-        for tensors in self:
-            sample_batch, _, _, batch_index, labels = tensors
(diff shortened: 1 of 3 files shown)


scvi/inference/posterior.py
```diff
@@ -1003,16 +1003,37 @@ def within_cluster_degenes(
...
-        return imputed_list.squeeze()
+    def imputation(
+        self,
+        n_samples: Optional[int] = 1,
+        transform_batch: Optional[Union[int, List[int]]] = None,
+    ) -> np.ndarray:
+        """
+        Imputes px_rate over self cells
+        :param n_samples:
+        :param transform_batch: Batches to condition on
```
adamgayoso: we should say what happens if you put in a list versus an int for this option.

PR #515, opened 2019-11-16
stop training when too many consecutive NaNs

### LINKED ISSUE #511: terminate training if nan elbo
we should terminate the training loop if the model has nan train loss and/or test loss for M consecutive iterations.

PierreBoyeau:
Fixes #511: warn the user when NaNs pop up and automatically stop training after N consecutive NaN training losses

### DIFF
--- a/scvi/inference/trainer.py
+++ b/scvi/inference/trainer.py
@@ -129,12 +129,7 @@ class Trainer:
         self.compute_metrics_time += time.time() - begin
 
     def train(
-        self,
-        n_epochs=20,
-        lr=1e-3,
-        eps=0.01,
-        max_number_of_consecutive_nans=10,
-        params=None,
+        self, n_epochs=20, lr=1e-3, eps=0.01, max_nans=10, params=None,
     ):
         begin = time.time()
         self.model.train()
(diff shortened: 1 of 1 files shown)


scvi/inference/trainer.py
```diff
@@ -171,6 +174,22 @@ def train(self, n_epochs=20, lr=1e-3, eps=0.01, params=None):
                 % (int(self.training_time), self.n_epochs)
             )
 
+    def check_training_status(self, loss):
+        loss_is_nan = torch.isnan(loss).item()
+        if loss_is_nan:
+            logger.warning("Model loss was NaN")
```
adamgayoso: I would say "Model training loss..." just to be clear

scvi/inference/trainer.py
```diff
@@ -171,6 +174,22 @@ def train(self, n_epochs=20, lr=1e-3, eps=0.01, params=None):
...
+            self.nan_counter += 1
+            self.previous_loss_was_nan = True
+        else:
+            self.nan_counter = 0
+            self.previous_loss_was_nan = False
+
+        if self.nan_counter >= 10:
```
adamgayoso: I'm not sure why, but this should probably be the choice of user (this threshold, probably in the train method?). With a default of 10 is fine.

scvi/inference/trainer.py
```diff
@@ -128,7 +128,14 @@ def compute_metrics(self):
                 self.model.train()
         self.compute_metrics_time += time.time() - begin
 
-    def train(self, n_epochs=20, lr=1e-3, eps=0.01, params=None):
+    def train(
+        self,
+        n_epochs=20,
+        lr=1e-3,
+        eps=0.01,
+        max_number_of_consecutive_nans=10,
```
adamgayoso: Can we just call this `man_nans`? And then explain it's consecutive in the docstring?

PR #517, opened 2019-11-20
Kl weighting by iterate

### LINKED ISSUE #507: KL warmup should be timed to the number of iterations, not epochs
For very large datasets the current implementation can result in too much warmup.

PierreBoyeau:
Fixes #507

### DIFF
--- a/scvi/inference/inference.py
+++ b/scvi/inference/inference.py
@@ -1,8 +1,12 @@
 import copy
+import warnings
+from typing import Union
 
 import matplotlib.pyplot as plt
 import torch
+from numpy import ceil
 
+from scvi.dataset import GeneExpressionDataset
 from scvi.inference import Trainer
 
 plt.switch_backend("agg")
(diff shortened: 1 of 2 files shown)


scvi/inference/inference.py
```diff
@@ -42,15 +46,26 @@ class UnsupervisedTrainer(Trainer):
...
+        n_iter_kl_warmup: int = None,
+        n_epochs_kl_warmup: int = 400,
+        normalize_loss: bool = None,
         **kwargs
     ):
         super().__init__(model, gene_dataset, **kwargs)
+
+        # Set up number of warmup iterations
+        n_reference_cells = min(10000, gene_dataset.nb_cells)
+        if n_iter_kl_warmup is not None:
```
adamgayoso: I'm not sure I understand this default. if you set a not None value that should be what is used. I was only suggesting 400 * 10000/batch_size as the default in the function signature.

scvi/inference/trainer.py
```diff
@@ -153,13 +161,14 @@ def train(
         ):
             self.on_epoch_begin()
             for tensors_list in self.data_loaders_loop():
+                self.on_iteration_begin()
                 if tensors_list[0][0].shape[0] < 3:
                     continue
-                loss = self.loss(*tensors_list)
-                self.check_training_status(loss, max_nans)
+                self.current_loss = loss = self.loss(*tensors_list)
```
adamgayoso: what's the implication of the training status now being checked AFTER the backward call?

PierreBoyeau: I don't think it has any impact: 

```
import torch

param = torch.tensor([10.], requires_grad=True)
optim = torch.optim.SGD([param], lr=0.1)
loss = param ** 2
print(loss)
loss.backward()
print(loss)

```

scvi/inference/inference.py
PierreBoyeau: The motivation for `n_reference_cells = min(10000, gene_dataset.nb_cells)` was to take into account the dataset size to set up the threshold iteration number.

adamgayoso: It's still a bit weird to me that you put a number for this and then it gets changed. So if I set this to be 100 iters, and my dataset is very large, then the trainer changes it to (10000/128) * 100?

I'm suggesting that if you set it to 100 iters that's all it should do. The default I'm suggesting is so that the performance of scVI doesn't change very much if doing iter warmup vs epoch warmup.

scvi/inference/trainer.py
adamgayoso: Sorry got confused with the iteration end and the epoch end and the early stopping stuff.

scvi/inference/inference.py
PierreBoyeau: Ok I think I get it!

PR #518, opened 2019-11-20
poisson conditional likelihood

romain-lopez:
added conditional likelihood for PPC check


PierreBoyeau:
I used Poisson likelihoods before and had to control/clamp the `px_rate` value.

This is linked to the fact that when `Poisson`'s rate is too high, the log proba is completely erratic. An example:

```
from torch.distributions import Poisson

print(Poisson(rate=1e12).log_prob(1e12))
>>> 2097152
```

I don't know to what extent this is important for us (I was working with synthetic data, and this is maybe why I had these problems), but this is maybe something we should be careful about!

PR #519, opened 2019-11-20
Posterior Predictive Module

adamgayoso:


adamgayoso:
@romain-lopez I made some changes, including keeping everything sparse to reduce memory consumption (need to use a new package called `sparse` to handle 3-d sparse matrices). I also moved this to a new "folder" called criticism and made it general enough for any posterior predictive samples, so it's a bit scVI agnostic. Finally, I added some testing for it. I still need to add docstrings and examples.

I think this is a good thing to have in our codebase for benchmarking our own models, and could potentially make it easier for others to compare their models to ours. What do you think?

romain-lopez:
I don't see the coverage running ?

romain-lopez:
The sparse implementation looks much better to me !

romain-lopez:
as well as not attaching it to the dic of posterior right away

setup.py
```diff
@@ -25,6 +25,7 @@
     "ipython>=7.1.1",
     "numba==0.45.0",  # numba 0.45.1 has a conflict with UMAP and numba 0.46.0 with parallelization in loompy
     "hyperopt==0.1.2",
+    "sparse>=0.8.0",
```
romain-lopez: can't we use scipy.sparse ?

scvi/criticism/ppc.py
```diff
@@ -0,0 +1,405 @@
...
+        )
+        model_corrs["gene_names2"] = np.repeat(
+            self.dataset.gene_names[self.gene_set],
+            len(self.dataset.gene_names[self.gene_set]),
+        )
+
+        df = pd.DataFrame.from_dict(model_corrs)
+        self.metrics["all gene-gene correlations"] = df
+
+    def calibration_error(self, confidence_intervals=None):
```
romain-lopez: add documentation, refer to a paper ?

scvi/criticism/ppc.py
```diff
@@ -0,0 +1,405 @@
...
+            df[m + "_mean"] = dr_mean.ravel()
+
+            dr_std = np.std(np.mean(samples == 0, axis=0), axis=-1).todense()
+            df[m + "_std"] = dr_std.ravel()
+
+        df["Raw"] = np.mean(self.raw_counts == 0, axis=0).todense()
+
+        self.metrics["dropout_ratio"] = df
+
+    def gene_gene_correlation(self, gene_indices=None, n_genes=1000):
```
romain-lopez: what is n_genes ?

adamgayoso:
> I don't see the coverage running ?

There was something wrong with the last merge (like 1/3 of travis fails because of autotune) so codecov didn't run properly on master.

setup.py
adamgayoso: SciPy.sparse only allows 2d. We need 3d. This is a pretty nice package. We can move this to criticism_requirements if necessary. It's a pretty easy install either way since these are the only requirements
numpy
scipy>=0.19
numba>=0.45

romain-lopez: I see, alright

romain-lopez:
tried to restart the build

romain-lopez:
@adamgayoso are we ready to merge this ? did you add a bit of docstring ?

romain-lopez:
To merge, we simply need to add a mock edit to trigger codecov

adamgayoso:
@romain-lopez I haven't had time yet to finish this. Probably need a few more weeks.

adamgayoso:
@romain-lopez I'm not entirely sure this belongs here. We could move it to the norm jam repo?

romain-lopez:
I'm fine either way. If we don't have some code explicitly using that I'm indeed not sure it's worth it to maintain it. It's still good code we can archive to the normjam.

romain-lopez:
@adamgayoso quick thought. Is there a way to use the sample function to perform posterior predictive checks as well as prior predictive checks? Should we add this to a scvi-tools default somehow? (cf. our work at normjam)

romain-lopez:
if this is too much work, or out of line, we may want to close this PR

adamgayoso:
Going to close this. I think it's out of scope for now.

PR #522, opened 2019-11-21
[WIP] notebook de modifs

PierreBoyeau:


PierreBoyeau:
Based on discussions with the team, it looks like differential expression modes presentation does not belong in `basic_tutorial.ipynb`.
I will do another PR which present differential expression procedures in another notebook.

PR #527, opened 2019-12-04
update readme

romain-lopez:


romain-lopez:
this should fix the codecov @adamgayoso @PierreBoyeau

PR #531, opened 2019-12-10
add anndata to main requirements

### LINKED ISSUE #530: Missing dependency?
Hi there!

I tried cloning the repository and installing locally in a fresh python 3.7.2 virtual environment and ran into a few problems. First when I tried `python setup.py install` I hit the following error:

```
error: prompt-toolkit 3.0.2 is installed but prompt-toolkit<2.1.0,>=2.0.0 is required by {'jupyter-console'}
```

Next I tried running `pip install -e .` from the root directory. This worked, but I had to install the package `anndata`. I think this package should perhaps be added to the requirements in `setup.py`.

adamgayoso:
- Fixes #530

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,6 +25,7 @@ requirements = [
     "ipython>=7.1.1",
     "numba==0.45.0",  # numba 0.45.1 has a conflict with UMAP and numba 0.46.0 with parallelization in loompy
     "hyperopt==0.1.2",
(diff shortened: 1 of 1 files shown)


adamgayoso:
@PierreBoyeau I think this is the only thing that is missing based on the issue. What do you think? Can we test base installation using travis?

PR #542, opened 2019-12-26
codecov settings

adamgayoso:
- Fixes the codecov issue with PRs not passing
- Sets a minimum coverage threshold


adamgayoso:
@romain-lopez I don't believe we need the patch commit status - this is a check of the actual PR itself.

PR #544, opened 2019-12-27
Colab notebooks

### LINKED ISSUE #498: basic tutorial batch correction
I think the tutorial is still a bit unclear. How do you let scVI correct batches for any dataset? If you dataset has non-zero batch indices will it automatically correct batches?

adamgayoso:
- Updates totalVI, basic tutorial, scanpy api, harmonization notebooks to have Colab links (tested to work in Colab) 
- Fixes #498


adamgayoso:
@romain-lopez @PierreBoyeau I think we should make notebooks a git submodule (make a new GitHub repository for it `scvi-notebooks` then treat this as a git submodule in this repo) and dump them there. The advantage of this is we can rerun notebooks in Colab and copy paste them without it looking like I added 3000 lines of code as it does here. Travis automatically grabs submodules so it shouldn't be an issue

romain-lopez:
This looks cool. I think we should add the colab badge in front of the notebooks name in the readme to feature this cool contribution!

PR #545, opened 2019-12-29
Notebooks submodule

adamgayoso:
Move notebooks to another repository and treat as a submodule here. This way we can update that repository freely without having to worry about crazy swings in line changes. We can also use that repository to archive older versions of notebooks with outputs.


adamgayoso:
Closing as this is out of scope for what we need.

PR #551, opened 2020-01-03
Add seurat v3 feature selection method

adamgayoso:
- Adds an alternative formulation of the Seurat v3 highly variable gene selection method.

I personally like this method the best as it's the most intuitive and does not require placing genes into bins.

### DIFF
--- a/tests/dataset/test_highly_variable_genes.py
+++ b/tests/dataset/test_highly_variable_genes.py
@@ -43,7 +43,7 @@ class TestHighlyVariableGenes(TestCase):
         )
         assert df["highly_variable"].sum() >= n_top
 
-        dataset.subsample_genes(n_top_genes=n_top)
+        dataset.subsample_genes(new_n_genes=n_top)
         new_genes = dataset.nb_genes
         assert n_genes > new_genes, "subsample_genes did not filter out genes"
         pass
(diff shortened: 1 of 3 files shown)


tests/dataset/test_highly_variable_genes.py
```diff
@@ -6,7 +6,7 @@
 
 class TestHighlyVariableGenes(TestCase):
     def test_sparse_no_batch_correction(self):
-        for flavor in ["seurat", "cell_ranger"]:
+        for flavor in ["seurat", "cell_ranger", "seurat_v3"]:
```
romain-lopez: seems like this should be seurat_v2 here

tests/dataset/test_highly_variable_genes.py
```diff
@@ -65,13 +60,17 @@ def test_dense_subsample_genes(self):
...
-        dataset.subsample_genes(new_n_genes=n_top)
-        assert dataset.nb_genes < n_genes
-        # For some reason the new number of genes can be slightly different than n_top
+        dataset.subsample_genes(new_n_genes=n_top, mode="cell_ranger")
+        assert dataset.nb_genes == n_top
 
         # With Seurat
         dataset = GeneExpressionDataset()
         dataset.populate_from_per_batch_list(data)
         dataset.subsample_genes(new_n_genes=n_top, mode="seurat")
```
romain-lopez: here too ?

PR #556, opened 2020-01-08
abstract trainer to reduce duplicate code in jvae

adamgayoso:
In the base trainer class I added some helper functions that make it easier to rely on the base train method for more advanced training scenarios - gimVI. I would be also able to use this new code in totalVI.

### DIFF
--- a/scvi/inference/jvae_trainer.py
+++ b/scvi/inference/jvae_trainer.py
@@ -117,7 +117,7 @@ class JVAETrainer(Trainer):
 
         return zip(*data_loaders)
 
-    def in_training_loop(self, tensors_list):
+    def on_training_loop(self, tensors_list):
 
         if self.train_discriminator:
             latent_tensors = []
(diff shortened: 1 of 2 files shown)


romain-lopez:
That looks like a good refactoring to me!

romain-lopez:
@ANazaret does this make sense to you ?

ANazaret:
It looks very good for refactorization!
Just one question: @adamgayoso there will be:
- in_training_loop
- on_training_begin
- on_training_end
- on_iteration_begin
- on_iteration_end
- on_epoch_begin
- on_epoch_end
- training_extras_init
- training_extras_end

Can you detail what is the difference between iteration/epochs/training exactly? Especially iterations/epochs, you could also add a comment in the code to explain.

Thanks

adamgayoso:
Thanks, @ANazaret. I think I made it more clear. Can you review now?

ANazaret:
Great, I understand the difference thank you. 
Maybe just change `in_training_loop`to `on_training_loop` to be consistent with others (the notion of `in` is in `loop`).

romain-lopez:
@adamgayoso should we merge this ?

PR #570, opened 2020-01-29
check X observation values when setting

### LINKED ISSUE #569: Verify X is counts
We should add some code to the X setter that checks if X values are in the natural numbers, not continuous, not negative, etc.

adamgayoso:
- Fixes #569

### DIFF
--- a/scvi/dataset/dataset.py
+++ b/scvi/dataset/dataset.py
@@ -544,6 +544,12 @@ class GeneExpressionDataset(Dataset):
                     n_dim
                 )
             )
+        valid_obs = check_nonnegative_integers(X)
+        if valid_obs is False:
+            logger.warning(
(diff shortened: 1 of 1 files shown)


scvi/dataset/dataset.py
```diff
@@ -1528,6 +1534,20 @@ def concatenate_arrays(arrays):
     return concatenation
 
 
+def check_nonnegative_integers(X: Union[np.ndarray, sp_sparse.csr_matrix]):
+    """Checks values of X to ensure it is count data"""
+
+    data = X if type(X) is np.ndarray else X.data
```
romain-lopez: how does this deal with sparsity ? is this why there is the X.data? If so, does it make it dense ? That would be a problem

romain-lopez: ah ok got it, sounds good to me

PR #576, opened 2020-02-04
pandas version

adamgayoso:


romain-lopez:
That does not sound like the right fix ? Let's not merge that. Is there a bug in 1.0 ? It sounds more like we should fix the code instead. If there is a real reason to reject this version then we should use != 1.0

PR #580, opened 2020-02-10
Added helper to get single gene normalized rates

vals:
I made a helper method for `Posterior` which lets you get expression rates at a normalized library size for a single gene of interest. This way you can e.g. plot the expression of a particular gene without needing to store the 'imputed' matrix for the entire dataset in RAM.


adamgayoso:
It seems like this could all be options in the already existing `get_sample_scale()` method. We should also add an option that averages the value over many samples. What do you think @romain-lopez ?

romain-lopez:
Hi Valentine, 

Thanks for contributing. Indeed, I think we could refactor this to be a modification of get_sample_scale by adding some options to it! 

Best, 
Romain

vals:
Sure! I'll take a stab at making a refactored version

scvi/inference/posterior.py
```diff
@@ -1144,9 +1144,31 @@ def get_stats(self):
...
+            This allows gene expression levels to be interpreted on a common scale of relevant
+            magnitude (default=1, no scaling).
+        :param batch_size: Size of minibatches of the data to use. When only accessing a single
+            gene a larger batch size (e.g. 10000) substantially reduces the time spent on GPU
+            data transfer (default=128).        
+        '''
+        if gene is None:
+            gene_idx = slice(None)
+        else:
+            gene_idx = [np.where(self.gene_dataset.gene_names == gene)[0][0]]
```
adamgayoso: Do you think it's helpful if we allow gene to be a subset of genes?

scvi/inference/posterior.py
```diff
@@ -1157,7 +1179,7 @@ def get_sample_scale(self, transform_batch=None):
                             y=labels,
                             n_samples=1,
```
adamgayoso: Should we allow many samples here and take the mean? In general it would be nice if these `get` methods could return samples and expected value estimates.

adamgayoso: If so, we could increase n_samples here (could lead to memory issues in some cases), or put a for loop around everything and take the average.

scvi/inference/posterior.py
vals: I was thinking about that. But I wasn't sure what a good way would be to relate which returned columns to which genes. (Related, there should probably be a check here to make sure 1) the gene is in the gene list, 2) the gene name is unique).

adamgayoso: So we have this function....
https://github.com/YosefLab/scVI/blob/4a5eef104fdd1c5eb548d7ab85704ba99f16a00f/scvi/dataset/dataset.py#L896-L900

We can use that `subset_genes` returned by this to subset the scale, and also to help define the return value, which could be a pandas dataframe in this case. This makes it easy to ignore the input order.

adamgayoso: and then it also doesn't matter if it's one gene or many genes. the input would be a list of genes.

vals: I think I arrived at a decent compromise: by default just return an array, but can specify to return a `DataFrame` with correct column names.

scvi/inference/posterior.py
vals: It's a bit messy but I added a number of options for getting samples.

vals:
It seems the tool `black` is failing on the CI. I'm not familiar with it, do you have any guidance on how I can address it?

adamgayoso:
@vals sorry I missed this. I'll give a review now, but perhaps upgrading your black will help fix this.

scvi/inference/posterior.py
```diff
@@ -1269,13 +1304,25 @@ def get_sample_scale(self, transform_batch=None):
...
                             transform_batch=transform_batch,
-                        )
+                        )[..., gene_mask] * library_size
                     ).cpu()
                 )
             ]
-        return np.concatenate(px_scales)
+
+        if n_samples > 1:
+            px_scales = np.concatenate(px_scales, axis=-2)
```
adamgayoso: Can't this one and the one below both be axis 0? If `n_samples > 1`, then isn't the shape `(samples, cells, genes)`?

scvi/inference/posterior.py
```diff
@@ -1258,9 +1258,44 @@ def get_stats(self):
...
+        :param return_mean: (optional) Whether to return the mean of the samples (default=True).
+        '''
+        if gene_list is None:
+            gene_mask = slice(None)
+        else:
+            gene_mask = self.gene_dataset._get_genes_filter_mask_by_attribute(gene_list, return_data=False)
+
         px_scales = []
-        for tensors in self:
+        for tensors in self.sequential(batch_size):
```
adamgayoso: To be consistent with the other functions, I'd leave this as `for tensors in self`, since the posterior itself can be updated. Also, this function is central to DE, so to be honest I'd have to clear up exactly how this affects everything else.

scvi/inference/posterior.py
adamgayoso: ah, no sorry -2 is cells, at least a comment would help us remember :)

adamgayoso:
@vals, we'd like to get 0.6.2 out soon. If you're able to address these items soon we could include this. No rush, but just wanted to keep you in the loop.

PR #581, opened 2020-02-11
totalVI updates

adamgayoso:


scvi/inference/posterior.py
```diff
@@ -381,7 +384,8 @@ def scale_sampler(
                 np.arange(len(self.gene_dataset))[selection], n_samples
             )
             self.update_sampler_indices(idx=idx)
-            px_scales.append(self.get_sample_scale(transform_batch=batch_idx))
+            px_scales.append(self.get_sample_scale(transform_batch=batch_idx, **kwargs))
+            batch_idx = batch_idx if batch_idx is not None else np.nan
```
romain-lopez: np.nan? That does not look very good

scvi/dataset/dataset.py
```diff
@@ -475,16 +490,27 @@ def populate_from_datasets(
...
-                    _, indices, _ = np.intersect1d(
+                    template = np.zeros(
+                        (
+                            getattr(gene_dataset, attribute_name).shape[0],
+                            len(columns_to_keep),
+                        ),
+                        dtype=getattr(gene_dataset, attribute_name).dtype,
+                    )
+                    if type(getattr(gene_dataset, columns_attr_name)) is not np.ndarray:
+                        template = sp_sparse.lil_matrix(template)
```
romain-lopez: can we add a short comment to explain what this is doing ?

scvi/inference/posterior.py
```diff
@@ -968,20 +984,19 @@ def within_cluster_degenes(
             cell_labels = self.gene_dataset.labels.ravel()
         de_res = []
         de_cluster = []
-        states = np.asarray([1 if x else 0 for x in states])
-        nstates = np.asarray([0 if x else 1 for x in states])
+        oppo_states = ~states
```
romain-lopez: can you comment the docstring or add one quick comment to explain what those things are doing?

romain-lopez: It looks like just refactoring but I'm not sure

scvi/inference/total_inference.py
```diff
@@ -432,27 +434,85 @@ def get_sample_mixing(self, n_samples: int = 1, give_mean: bool = True):
 
         return py_mixings
 
+    @torch.no_grad()
+    def get_sample_scale(
+        self,
+        transform_batch=None,
+        eps=0.5,
```
romain-lopez: docstring ? What is espsilon ?

romain-lopez: and all the other variables :)

scvi/inference/total_inference.py
```diff
@@ -432,27 +434,85 @@ def get_sample_mixing(self, n_samples: int = 1, give_mean: bool = True):
...
+            )
+            # prior count for proteins
+            scales += [torch.cat(model_scale, dim=-1).cpu().numpy()]
+        return np.concatenate(scales)
+
     @torch.no_grad()
     def get_normalized_denoised_expression(
-        self, n_samples: int = 1, give_mean: bool = True
+        self,
+        n_samples: int = 1,
```
romain-lopez: it is clearer here !

scvi/inference/total_inference.py
```diff
@@ -476,6 +536,175 @@ def get_normalized_denoised_expression(
...
+                px_dispersion = torch.ones_like(x) * px_["r"]
+            if len(py_["r"].size()) == 2:
+                py_dispersion = py_["r"]
+            else:
+                py_dispersion = torch.ones_like(y) * py_["r"]
+
+            dispersion = torch.cat((px_dispersion, py_dispersion), dim=-1)
+
+            # This gamma is really l*w using scVI manuscript notation
+            p = rate / (rate + dispersion)
```
romain-lopez: this appears now multiple times in the codebase. Should we have some basic code for sampling from NB ? These numpy comments and parameterization of the NB are always heavy

scvi/inference/total_inference.py
```diff
@@ -563,20 +972,45 @@ def __init__(
...
         pro_recons_weight=1.0,
-        n_epochs_back_kl_warmup=200,
-        n_epochs_kl_warmup=200,
-        **kwargs
+        n_epochs_kl_warmup=None,
+        n_iter_kl_warmup="auto",
+        early_stopping_kwargs=default_early_stopping_kwargs,
+        discriminator=None,
+        use_adversarial_loss=False,
+        kappa=0.1,
```
romain-lopez: I  thought kappa was annealed as well ?

scvi/inference/total_inference.py
```diff
@@ -618,13 +1052,97 @@ def loss(self, tensors):
             + self.pro_recons_weight * reconst_loss_protein
             + self.kl_weight * kl_div_z
             + kl_div_l_gene
```
romain-lopez: that simplifies a bit that this is the same

scvi/inference/total_inference.py
```diff
@@ -618,13 +1052,97 @@ def loss(self, tensors):
...
+
+        return loss
+
+    def _get_z(self, tensors):
+        (
+            sample_batch_X,
+            local_l_mean,
+            local_l_var,
+            batch_index,
+            label,
```
romain-lopez: is this function necessary ?

scvi/models/totalvi.py
```diff
@@ -87,21 +91,22 @@ def __init__(
...
+        self.protein_batch_mask = protein_batch_mask
 
         # parameters for prior on rate_back (background protein mean)
         if n_batch > 0:
             self.background_pro_alpha = torch.nn.Parameter(
                 torch.randn(n_input_proteins, n_batch)
             )
             self.background_pro_log_beta = torch.nn.Parameter(
-                torch.randn(n_input_proteins, n_batch)
+                torch.clamp(torch.randn(n_input_proteins, n_batch), -10, 1)
```
romain-lopez: if you need to clamp at the beginning, is it just because you did not initialize well ? There are some trunctated distributions to sample from for initialization

adamgayoso: It's not critical and is set to this in all the benchmarking code, so I think it's fine to leave for now?

scvi/inference/total_inference.py
adamgayoso: that code will need to be somewhere, be it in the `on_training_loop` or as a "private" function here.

scvi/inference/total_inference.py
adamgayoso: It is. Need to update default here.

scvi/inference/total_inference.py
adamgayoso: we should really write our own custom NB distribution, which subclasses pytorch objects.

adamgayoso: It's not really the place to do this here. We could also write our own Gamma distribution that uses our parameterization.

scvi/inference/posterior.py
adamgayoso: This is due to the line directly below

` batch_ids.append(np.ones((px_scales[-1].shape[0])) * batch_idx)`

This actually fixes the bug, because you can't multiply by None.

Basically, this is for the `use_observed_batches` mode. I agree this is not straightforward. Somehow we should consider a differential expression module that cleans up a lot of the code base.

scvi/inference/posterior.py
adamgayoso: this is the same code that's in master, it fixes a bug where only cell 0 and cell 1 were being compared because they were ints.

scvi/inference/total_inference.py
adamgayoso: Not sure what you mean?

adamgayoso:
I need to update all docstrings, which I'll do before merging.

adamgayoso:
I need to add more testing.

scvi/inference/posterior.py
romain-lopez: hum ok, I guess it's just unsual to assign something to NaN. Can we do something more readable?

scvi/inference/posterior.py
romain-lopez: I see OK

scvi/inference/total_inference.py
romain-lopez: yes agreed. We should write a github issue

scvi/inference/total_inference.py
romain-lopez: it's great that you're using the same parameter

scvi/inference/total_inference.py
romain-lopez: ok

scvi/models/totalvi.py
romain-lopez: ok

scvi/inference/posterior.py
adamgayoso: I propose we leave this now, and I will make an issue to fix at a later time

Issue #532, opened 2019-12-12
add info to the README


romain-lopez:
1. Description
	Functionalities
	list of sub-tools

2. Links to licence, documentation, discourse, blog

3. How to install: 

4. List of notebooks with a short descriptive sentence

6. List of papers

romain-lopez:
List of papers should include at least the LDVAE. Supposedly, it should also include all papers whose code is hosted in the codebase

romain-lopez:
I think though we should add a clarification to the readme to indicate that in order to run the notebooks, we recommend to use the install pip but with the notebooks extras

PR #590, opened 2020-02-21
README refactor

### LINKED ISSUE #562: Add colab badge to readme

PierreBoyeau:
Fixes #562  #532

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,12 +2,14 @@
 scVI
 ====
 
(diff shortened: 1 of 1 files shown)


romain-lopez:
This looks good! Also, it is actually not needed to add the badge in front of every single notebook. Good catch! 

One quick comment though, can we add some **text** somewhere in the README so people can realize we made the notebooks available on colab

A simple but probably suboptimal proposal:
-> Follow along with our Jupyter notebooks to quickly get familiar with scVI (also available on Colab)!

You can address this while refactoring the README (the other issue).

PR #591, opened 2020-02-22
posterior updates

### LINKED ISSUE #557: Potential issue in DE when using a Posterior that does not include entire dataset
It's hard for me to tell if this is a bug. If I make a Posterior object on some subset of cells, shouldn't the following lines be in reference to the indices I already am using in the Posterior?

https://github.com/YosefLab/scVI/blob/6c9f43e3332e728831b174c1c1f0c9127b77cba0/scvi/inference/posterior.py#L379-L385

### LINKED ISSUE #563: Methods to save and load the posterior
I think it would be helpful if we added methods to save and load a posterior object. Since the posterior points to the model and to dataset, it is indeed the functional object of scVI. We should brainstorm a bit how this can best be done.

### LINKED ISSUE #588: give_mean option to posterior sampling functions

### LINKED ISSUE #599: Check defaults
~Reminder to check the defaults for the DE functions before the release.~

Reminder to check ALL defaults before release

PierreBoyeau:
Fixes #557, Fixes #588, Fixes #563, Fixes #599


scvi/dataset/anndataset.py
```diff
@@ -30,6 +35,7 @@ def __init__(
         ctype_label: str = "cell_types",
         class_label: str = "labels",
         use_raw: bool = False,
+        cell_measurements_columns: Optional[Dict[str, str]] = None,
```
adamgayoso: Can we rename this as `cell_measurements_mappings`?

scvi/inference/posterior.py
```diff
@@ -345,11 +376,16 @@ def scale_sampler(
                 - n_bio_batches * n_cells * n_samples_per_cell
                 or
                  - n_samples_total
-            `batch`
+            `bat_doc_paramsch`
```
adamgayoso: what is this?

scvi/dataset/dataset.py
```diff
@@ -1237,6 +1238,44 @@ def reorder_cell_types(self, new_order: Union[List[str], np.ndarray]):
     #                           #
     #############################
 
+    def to_anndata(self) -> anndata.AnnData:
+        """
+            Converts the dataset to a anndata.AnnData object.
+            The obtained dataset can then be saved/retrieved using the anndata API.
+        """
+
```
adamgayoso: what happens to dataset.cell_types here? I think these mappings are in dataset.attribute_mappings?

scvi/dataset/anndataset.py
PierreBoyeau: I wanted to be consistent with the `Dataset` attribute. I agree that `cell_measurements_mappings` is more readable, but we then lose the name consistency. Do you still want the rename? If yes, should we rename the `Dataset` attribute?

scvi/dataset/dataset.py
PierreBoyeau: I manually fixed the cell_types situation.

PierreBoyeau: I manually fixed the cell_types situation.

scvi/inference/posterior.py
```diff
@@ -1504,11 +1534,63 @@ def apply_t_sne(latent, n_samples=1000):
...
+        Because of pytorch model loading usage, this function needs a scVI model object initialized with exact same parameters
+        that during training.
+
+        :param dir_path: directory containing the posterior properties to be retrieved.
+        :param model: scVI initialized model.
+        :param use_cuda: whether the model should use cuda.
+
+        Usage example:
+        1. Save posterior
+            >>> model = VAE(nb_genes, n_batches, n_hidden=64, n_latent=5)
```
adamgayoso: I would use default parameters here so no one thinks these are reasonable parameters

scvi/dataset/dataset.py
```diff
@@ -1237,6 +1240,47 @@ def reorder_cell_types(self, new_order: Union[List[str], np.ndarray]):
...
+        # obs/obsm contruction
+        obsm = dict()
+        obs = dict()
+        for key in self.cell_attribute_names:
+            if key not in ["local_means", "local_vars"]:
+                vals = getattr(self, key)
+                if key in self.cell_measurements_col_mappings:
+                    obsm[key] = vals
+                elif key == "labels" and self.cell_types is not None:
+                    obs["cell_types"] = self.cell_types[vals.squeeze()]
```
adamgayoso: I believe we need to make self.cell_types a np array for this to work, since it's just a list. I suppose this is not tested?

PR #595, opened 2020-02-26
ZINB and NB torch distributions

### LINKED ISSUE #582: Create pytorch distributions for gamma and NB/ZINB using our parameterizations
This will clean up a lot of sampling code.

PierreBoyeau:
Fixes #582


scvi/inference/posterior.py
```diff
@@ -41,7 +41,7 @@
     compute_marginal_log_likelihood_scvi,
     compute_marginal_log_likelihood_autozi,
 )
-
+from scvi.models.distributions import NB, ZINB
```
adamgayoso: Should we follow the pytorch naming conventions (e.g., NegativeBinomial?)

scvi/models/distributions.py
```diff
@@ -0,0 +1,196 @@
...
+    broadcast_all,
+    probs_to_logits,
+    lazy_property,
+    logits_to_probs,
+)
+
+from scvi.models.log_likelihood import log_nb_positive, log_zinb_positive
+
+
+def from_nb_params2_to_1(mu, theta, eps=1e-6):
```
adamgayoso: is there some way we can give these functions a more specific name? Also should we make them "private"? "_from..."

scvi/models/__init__.py
```diff
@@ -15,4 +16,6 @@
     "Classifier",
     "AutoZIVAE",
     "TOTALVI",
+    "ZINB",
```
adamgayoso: I'm not sure putting this here is necessary is it?, all your imports are from scvi.models.distributions

scvi/models/distributions.py
PierreBoyeau: tried my best but it's not that easy

PR #598, opened 2020-02-26
created a history file up to date

### LINKED ISSUE #546: Add contributions of collaborators to HISTORY.RST
I would like to add some traceback on who collaborated to scVI (using github pseudonames, which is easy to get) and recreate a short version of the history (Valentine for example and the LDVAE but this is not the only contribution. Also autotune and all the models...). We do not need to annotate everything but at least add the contributions.

romain-lopez:
- Fixes #546

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,19 +2,140 @@
 History
 =======
 
-0.1.0 (2018-06-12)
-------------------
-
-* First release on PyPI.
+0.6.0 
+-----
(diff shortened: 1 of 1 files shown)


romain-lopez:
#546

adamgayoso:
For 0.6.0, we should add Updates to LDVAE, totalVI from me. These will correspond to publication versions.

romain-lopez:
done

PR #646, opened 2020-04-02
add worm tutorial

### LINKED ISSUE #526: I made a notebook for vanilla DE with interactive plots
Hello, following after the basic scVI tutorial I made a notebook that performs vanilla DE and makes interactive plot. It uses Plotly instead of scanpy.

It makes the following plots for a dataset of 89k C. elegans cells:

- tSNE, UMAP and TSVD of latent space
- Volcano plot of two clusters, showing scVI expression scale vs absolute bayes factor
- Heatmap of top 10 genes, showing short gene descriptions on mouseover 

I tried to make it as clear as possible and include a discussion of the `vanilla` and the `change` DE modes. I thought that maybe this could be helpful for an upcoming tutorial on the `change` mode ;)

Here's the notebook on Colab (for some reason nbviewer formatting was broken). 
Colab has the advantadge that you can run the notebook right away, perhaps something to keep in mind when doing new tutorials. For this notebook, with 1000 genes and 89k cells I only do 5 epochs (7 min total) and it trains very quickly. The slowest part is the t-SNE with openTSNE!

https://colab.research.google.com/drive/1hF7KSujhhHcyxzWkzAHy9lazXLexainr

If you have any comments (or find something I did wrong/could be improved) I'd be happy to hear. Hopefully this can be useful!

adamgayoso:
- Closes #526

### DIFF
--- a/docs/contributed_tutorials/index.rst
+++ b/docs/contributed_tutorials/index.rst
@@ -1,0 +1,11 @@
+Contributed tutorials
+=====================
+
+The following tutorials have been contributed by members of the scVI user community!
+
+
+.. toctree::
+   :maxdepth: 1
+   :hidden:
+
+   scVI_DE_worm
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -21,6 +21,7 @@ These models are able to simultaneously perform many downstream tasks such as le
 
    installation
    tutorials/index
+   contributed_tutorials/index
    contributing
    history
    references


adamgayoso:
Hi @Munfred, could you take a look at this? I changed a few of the hyperparameters and added some text. Thanks!

Munfred:
I reviewed the content and seems to be in order, but I think the colab button is broken. It doesn't work when I try to click on it, doesn't find the notebook. And when I tried manually loading it in colab form the link https://github.com/YosefLab/scVI/blob/de_worm/docs/contributed_tutorials/scVI_DE_worm.ipynb it loads an empty notebook. 

However, when I downloaded the file and uploaded it to colab, it loaded everything correctly, including the interactive plot results. I'm not sure what causes the issue.

adamgayoso:
ah, that must be because the link is pointing to the ipynb from the stable branch, which is the correct link once I pull everything there.

PR #648, opened 2020-04-03
Python 3.6

adamgayoso:
Add back support for 3.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,12 +3,15 @@ dist: xenial
 
 language: python
 
-python:
-  - 3.6
-  - 3.7
-
 os:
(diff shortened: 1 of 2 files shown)


adamgayoso:
I figured since many single cell packages still rely on 3.6, and that we don't really have any 3.7 specific features, it would be easy enough to support 3.6. Also, Colab uses 3.6.

romain-lopez:
Looks good to me, we deleted this to use dataclasses and because it was outdated and it would make more maintenance for us. However, it's true some other people are still using it so let's go for it. One thing, I can see that codecov worked but it did not post the coverage here ?

PR #663, opened 2020-04-28
WIP: Adding fixes for numerical stability

mortonjt:
There are two majors types of fixes here

- Multiple exponentiation steps have been removed, and now most computations are only in log coordinates
- Negative Binomial only works with log inputs

Note that this is only on the Negative Binomial + LDVAE. There are likely a lot more issues downstream.

### DIFF
--- a/scvi/models/distributions.py
+++ b/scvi/models/distributions.py
@@ -43,7 +43,8 @@ def _convert_counts_logits_to_mean_disp(total_count, logits):
         :return: the mean and inverse overdispersion of the NB distribution.
     """
     theta = total_count
-    mu = logits.exp() * theta
+    # mu = logits.exp() * theta
+    mu = logits + torch.log(theta)
     return mu, theta
 
 
@@ -60,8 +61,8 @@ class NegativeBinomial(Distribution):
     one parameterization to another.
     """
     arg_constraints = {
-        "mu": constraints.greater_than_eq(0),
-        "theta": constraints.greater_than_eq(0),
+        "log_mu": constraints.real,
+        "log_theta": constraints.real,
     }
     support = constraints.nonnegative_integer
 
@@ -70,12 +71,12 @@ class NegativeBinomial(Distribution):
         total_count: torch.Tensor = None,
         probs: torch.Tensor = None,
         logits: torch.Tensor = None,
-        mu: torch.Tensor = None,
-        theta: torch.Tensor = None,
+        log_mu: torch.Tensor = None,
+        log_theta: torch.Tensor = None,
         validate_args=True,
     ):
         self._eps = 1e-8
-        if (mu is None) == (total_count is None):
+        if (log_mu is None) == (total_count is None):
             raise ValueError(
                 "Please use one of the two possible parameterizations. Refer to the documentation for more information."
             )
@@ -87,11 +88,11 @@ class NegativeBinomial(Distribution):
             logits = logits if logits is not None else probs_to_logits(probs)
             total_count = total_count.type_as(logits)
             total_count, logits = broadcast_all(total_count, logits)
-            mu, theta = _convert_counts_logits_to_mean_disp(total_count, logits)
+            log_mu, log_theta = _convert_counts_logits_to_mean_disp(total_count, logits)
         else:
-            mu, theta = broadcast_all(mu, theta)
-        self.mu = mu
-        self.theta = theta
+            log_mu, log_theta = broadcast_all(log_mu, log_theta)
+        self.log_mu = log_mu
+        self.log_theta = log_theta
         super().__init__(validate_args=validate_args)
 
     def sample(self, sample_shape=torch.Size()):
@@ -115,11 +116,13 @@ class NegativeBinomial(Distribution):
                     "The value argument must be within the support of the distribution",
                     UserWarning,
                 )
-        return log_nb_positive(value, mu=self.mu, theta=self.theta, eps=self._eps)
+        return log_nb_positive(value, log_mu=self.log_mu,
+                               log_theta=self.log_theta, eps=self._eps)
 
     def _gamma(self):
-        concentration = self.theta
-        rate = self.theta / self.mu
+        concentration = torch.exp(self.log_theta)
+
+        rate = torch.exp(self.log_theta - self.log_mu)
         # Important remark: Gamma is parametrized by the rate = 1/scale!
         gamma_d = Gamma(concentration=concentration, rate=rate)
         return gamma_d
--- a/scvi/models/distributions.py
+++ b/scvi/models/distributions.py
@@ -43,9 +43,8 @@ def _convert_counts_logits_to_mean_disp(total_count, logits):
         :return: the mean and inverse overdispersion of the NB distribution.
     """
     theta = total_count
-    # mu = logits.exp() * theta
-    mu = logits + torch.log(theta)
-    return mu, theta
+    log_mu = logits + torch.log(theta)
+    return log_mu, theta
 
 
 class NegativeBinomial(Distribution):
--- a/scvi/models/log_likelihood.py
+++ b/scvi/models/log_likelihood.py
@@ -5,6 +5,7 @@ import torch
 import torch.nn.functional as F
 from torch import logsumexp
 from torch.distributions import Normal, Beta
+from scvi.utils import logexpsum
 
 
 def compute_elbo(vae, posterior, **kwargs):
@@ -262,11 +263,13 @@ def log_nb_positive(x, log_mu, log_theta, eps=1e-8):
         )  # In this case, we reshape theta for broadcasting
 
     theta = torch.exp(log_theta)
-    res = (theta * log_theta + x * log_mu
-           + torch.lgamma(x + theta)
-           + torch.lgamma(x + theta)
-           - torch.lgamma(theta)
-    )
+    log_denom = logexpsum(log_theta + log_mu))
+    res = x * (log_mu - log_denom) + \
+        theta * (log_theta - log_denom) + \
+        torch.lgamma(x + theta) - \
+        torch.lgamma(theta) - \
+        torch.lgamma(x + 1)
+
     return res
 
 
--- a/scvi/models/log_likelihood.py
+++ b/scvi/models/log_likelihood.py
@@ -5,7 +5,7 @@ import torch
 import torch.nn.functional as F
 from torch import logsumexp
 from torch.distributions import Normal, Beta
-from scvi.utils import logexpsum
+from scvi.models.utils import logexpsum
 
 
 def compute_elbo(vae, posterior, **kwargs):
@@ -263,7 +263,7 @@ def log_nb_positive(x, log_mu, log_theta, eps=1e-8):
         )  # In this case, we reshape theta for broadcasting
 
     theta = torch.exp(log_theta)
-    log_denom = logexpsum(log_theta + log_mu))
+    log_denom = logexpsum(log_theta, log_mu)
     res = x * (log_mu - log_denom) + \
         theta * (log_theta - log_denom) + \
         torch.lgamma(x + theta) - \
--- a/scvi/models/log_likelihood.py
+++ b/scvi/models/log_likelihood.py
@@ -246,7 +246,7 @@ def log_zinb_positive(x, mu, theta, pi, eps=1e-8):
     return res
 
 
-def log_nb_positive(x, mu, theta, eps=1e-8):
+def log_nb_positive(x, log_mu, log_theta, eps=1e-8):
     """
     Note: All inputs should be torch Tensors
     log likelihood (scalar) of a minibatch according to a nb model.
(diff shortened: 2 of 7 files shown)


scvi/models/modules.py
```diff
@@ -314,7 +314,7 @@ def forward(
         raw_px_scale = self.factor_regressor(z, *cat_list)
         px_scale = torch.softmax(raw_px_scale, dim=-1)
         px_dropout = self.px_dropout_decoder(z, *cat_list)
-        px_rate = torch.exp(library) * px_scale
+        px_rate = library + px_scale
```
mortonjt, line 317: This is interesting. What are the actual units of library size here?

mortonjt: Looks like the library size is already in log units

adamgayoso:
Thank you for the contribution. I think what you're proposing here would require major global backend changes.  I think it might be more worthwhile to identify what is causing the nan NB log likelihood. My intuition is that the library size parameter is exploding. A simple thing to do would be to add a clamp to that parameter (note library is generally on the log scale in the scVI code). Otherwise, I'm sure there's some anomaly detection in pytorch.

scvi/models/distributions.py
```diff
@@ -43,7 +43,8 @@ def _convert_counts_logits_to_mean_disp(total_count, logits):
         :return: the mean and inverse overdispersion of the NB distribution.
     """
     theta = total_count
-    mu = logits.exp() * theta
+    # mu = logits.exp() * theta
+    mu = logits + torch.log(theta)
```
adamgayoso, line 47: We don't use this parameterization during optimization so I'm not sure how this will fix any of the numerical issues.

scvi/models/log_likelihood.py
```diff
@@ -256,21 +256,17 @@ def log_nb_positive(x, mu, theta, eps=1e-8):
...
-        theta * (torch.log(theta + eps) - log_theta_mu_eps)
-        + x * (torch.log(mu + eps) - log_theta_mu_eps)
-        + torch.lgamma(x + theta)
-        - torch.lgamma(theta)
-        - torch.lgamma(x + 1)
+    theta = torch.exp(log_theta)
+    res = (theta * log_theta + x * log_mu
+           + torch.lgamma(x + theta)
+           + torch.lgamma(x + theta)
+           - torch.lgamma(theta)
```
adamgayoso, line 268: is this correct? There should probably be a log sum exp for the `log_theta_mu_eps` term in the old code.

scvi/models/modules.py
adamgayoso: indeed library in the scVI code is on the log scale

scvi/models/vae.py
```diff
@@ -216,7 +216,7 @@ def get_reconstruction_loss(
             )
         elif self.reconstruction_loss == "nb":
             reconst_loss = (
-                -NegativeBinomial(mu=px_rate, theta=px_r).log_prob(x).sum(dim=-1)
+                -NegativeBinomial(mu=px_rate, theta=px_r).log_prob(x).mean(dim=-1)
```
adamgayoso, line 219: This will change a lot of things, the loss is averaged in the `loss` function in the trainer.

scvi/models/log_likelihood.py
mortonjt: The beauty here is that `log_theta_mu_eps` is no longer needed, since theta is already in log units, no need for the back and forth transform.

mortonjt:
@adamgayoso thanks for the review. 

The exploding library size issue is another discussion - I have a feeling that this can be replaced with a constant Poisson offset (see [here](https://arxiv.org/abs/1311.6139) for an example) (probably going to test this out later).

But regarding the log parameterization of NB, I've seen this issue reappear in multiple contexts at this point, in my own packages and in [Stan](https://mc-stan.org/docs/2_19/functions-reference/neg-binom-2-log.html). I realize that this is going to require a major overhaul, but it is likely going to worthwhile, since it'll make this package more robust.

scvi/models/log_likelihood.py
adamgayoso: This is the function we optimize, `r=theta` here and `m=mu`. Therefore you're going to have a `log(m + r)` term. If `mu` and `r` are on the log scale you'll have to do a logsumexp computation.
Screen Shot 2020-04-28 at 10 50 47 AM

mortonjt: Awesome. Thanks for the reference.

scvi/models/vae.py
mortonjt: ahh, I see. ok, then this change probably can be reverted then

romain-lopez:
Hi @mortonjt, 

Thanks for contributing to scVI,

A) I am not convinced that working on log space will help because of the log of sum that Adam pointed out, though I am not familiar with the parameterization you sent. Can you point out some ressources that show some improvement / evidence on other packages maybe? 

B) What seems more worth investigating to me is whether using l as observed makes the problem go away: 
> l = sum(UMIs in the given cell)
We had a lot of trouble with l, had to clamp it sometimes etc.. so we know that this is a recurring issue.

C) Also, it might be a learning rate or an initialization problem? It would help to have a detailed diagnostic of what is going wrong on your use case. 

Thanks,

mortonjt:
Hi @romain-lopez , I think that is reasonable.  This PR still has a bit of a ways to go (hence the WIP). 

Regarding the logexpsum, there are better numerical techniques for handling that (see this [blog post](https://leimao.github.io/blog/LogSumExp/)).  

Diagnostics will be up next.

scvi/models/distributions.py
jeff-regier: We probably should call it `log_mu` here rather than `mu` if we're going to starting storing it on the log scale. Could get confusing otherwise.

mortonjt:
Here is an example to reproduce the issue (after installing my fork with `pip install git+https://github.com/mortonjt/scVI.git@biom`

```python
    import numpy as np
    from scvi.dataset import BiomDataset
    from scvi.inference import UnsupervisedTrainer
    from scvi.models import LDVAE

    from mavi.utils import load_unsupervised, save_unsupervised

    url = r'https://github.com/biocore/American-Gut/blob/master/data/AG/AG_100nt.biom?raw=true'
    dataset = BiomDataset(filename='AG_100nt.biom', url=url, save_path="tests/data")
    train-size = 0.8 
    epochs = 300 
    n_latent = 16 
    n_layers_encoder = 2 
    n_hidden = 64 
    reconstruction_loss = 'nb' 
    learning_rate = 1e-4
    gpu = True
    vae = LDVAE(dataset.nb_genes, dataset.n_batches, dataset.n_labels,
                n_latent=n_latent,
                n_layers_encoder=n_layers_encoder,
                n_hidden=n_hidden,
                reconstruction_loss=reconstruction_loss,
                latent_distribution=latent_distribution)
    trainer = UnsupervisedTrainer(vae, dataset, train_size=train_size,
                                  use_cuda=gpu)
    trainer.train(n_epochs=n_epochs)

    posterior = trainer.create_posterior(
        trainer.model, dataset, indices=np.arange(len(dataset)))
```

Below is the error that I get 

```
[2020-05-11 18:35:23,274] INFO - scvi.dataset.dataset | Remapping labels to [0,N]
[2020-05-11 18:35:23,276] INFO - scvi.dataset.dataset | Remapping batch_indices to [0,N]
[2020-05-11 18:35:23,307] INFO - scvi.dataset.dataset | Computing the library size for the new data
[2020-05-11 18:35:23,315] INFO - scvi.dataset.dataset | Downsampled from 22742 to 22742 cells
[2020-05-11 18:35:29,896] INFO - scvi.inference.inference | KL warmup phase exceeds overall training phaseIf your applications rely on the posterior quality, consider training for more epochs or reducing the kl warmup.
[2020-05-11 18:35:29,896] INFO - scvi.inference.inference | KL warmup for 400 epochs
training:  89%|████████▊ | 266/300 [05:27<00:41,  1.23s/it][2020-05-11 18:40:57,285] WARNING - scvi.inference.trainer | Model training loss was NaN
/mnt/home/jmorton/venvs/mmvec-torch/lib/python3.6/site-packages/numba/np/ufunc/parallel.py:355: NumbaWarning: The TBB threading layer requires TBB version 2019.5 or later i.e., TBB_INTERFACE_VERSION >= 11005. Found TBB_INTERFACE_VERSION = 6103. The TBB threading layer is disabled.
  warnings.warn(problem)
Traceback (most recent call last):
  File "/mnt/home/jmorton/venvs/mmvec-torch/bin/mavi", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/mnt/home/jmorton/software/mavi/scripts/mavi", line 74, in <module>
    mavi()
  File "/mnt/home/jmorton/venvs/mmvec-torch/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/mnt/home/jmorton/venvs/mmvec-torch/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/mnt/home/jmorton/venvs/mmvec-torch/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/mnt/home/jmorton/venvs/mmvec-torch/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/mnt/home/jmorton/venvs/mmvec-torch/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/mnt/home/jmorton/software/mavi/scripts/mavi", line 53, in unsupervised
    trainer.train(n_epochs=n_epochs)
  File "/mnt/home/jmorton/software/scVI/scvi/inference/trainer.py", line 177, in train
    self.on_training_loop(tensors_list)
  File "/mnt/home/jmorton/software/scVI/scvi/inference/trainer.py", line 201, in on_training_loop
    self.current_loss = loss = self.loss(*tensors_list)
  File "/mnt/home/jmorton/software/scVI/scvi/inference/inference.py", line 109, in loss
    sample_batch, local_l_mean, local_l_var, batch_index, y
  File "/mnt/home/jmorton/venvs/mmvec-torch/lib/python3.6/site-packages/torch/nn/modules/module.py", line 547, in __call__
    result = self.forward(*input, **kwargs)
  File "/mnt/home/jmorton/software/scVI/scvi/models/vae.py", line 316, in forward
    reconst_loss = self.get_reconstruction_loss(x, px_rate, px_r, px_dropout)
  File "/mnt/home/jmorton/software/scVI/scvi/models/vae.py", line 219, in get_reconstruction_loss
    -NegativeBinomial(mu=px_rate, theta=px_r).log_prob(x).sum(dim=-1)
  File "/mnt/home/jmorton/software/scVI/scvi/models/distributions.py", line 95, in __init__
    super().__init__(validate_args=validate_args)
  File "/mnt/home/jmorton/venvs/mmvec-torch/lib/python3.6/site-packages/torch/distributions/distribution.py", line 36, in __init__
    raise ValueError("The parameter {} has invalid values".format(param))
ValueError: The parameter mu has invalid value
```
Still need to sanity check with @adamgayoso's comments - so more details to come.

scvi/models/utils.py
```diff
@@ -44,3 +44,10 @@ def batch(batch_size, label):
 
     batch_size = x.size(0)
     return torch.cat([batch(batch_size, i) for i in range(y_dim)])
+
+
+def logexpsum(logx, logy):
```
adamgayoso, line 49: I believe `torch.logsumexp` implements this

scvi/models/modules.py
adamgayoso: you will need logsoftmax here for px_scale

adamgayoso:
Hi @mortonjt, I took a look at the dataset you linked,

The L1 norm of each observation (library size) ranges from 1 to 7,233,319. 

The mean and variance of the log library size:
```
In [16]: np.var(np.log(X.sum(axis=1)))                                                                
Out[16]: 7.068304947329521

In [17]: np.mean(np.log(X.sum(axis=1)))                                                               
Out[17]: 3.248720784509202
```

It seems that taking \ell as observed will likely fix your issue, though this requires non-trivial coding to do. 

Are you planning on refactoring all models and modules to operate on the log scale? Right now it seems like you're only targeting the LDVAE?

scvi/models/utils.py
mortonjt: I'm assuming you're referring to this: https://pytorch.org/docs/master/generated/torch.logsumexp.html

That is a possible option (although the syntax is slightly different).

mortonjt:
Hi @adamgayoso thanks for taking a closer look!

I have applied the `torch.clamp` approach as suggested earlier, and that actually seems to be giving promising results. Below is an example on the Human Microbiome Project when distinguishing body site (this is expected to give obvious clusters).

image

I haven't quite gotten the same results with the numerical stability changes in the PR - so there are still some bugs in the pull request.  I wasn't planning on touching the other PRs (mainly because I'm not planning on using the other modules).  But I'll continue hacking away at this as more edge cases pop up.

adamgayoso:
I'm going to close this PR as all the code has changed. We are exploring this problem in the context of our new code.

mortonjt:
Sounds good. I don't think this has a problem any more.

PR #670, opened 2020-05-20
Poisson variable gene selection

vals:
Hi,

One of my main bottlenecks have been selection of highly variable genes.

Here I have implemented the feature selection method from https://github.com/tallulandrews/M3Drop based on increased number of observed zeros than expected from Poisson sampling for biologically variable genes.

This implementation makes use of a sparse count matrix, and utilizes PyTorch for computation. For 300,000 and 30,000 genes it needs about 2GB of memory, either in RAM or GPU.

In the original M3Drop package a Z-test is used to quantify significance for genes with increased zeros (if I read the R code right). Here I have used binomial distributions and sampling using PyTorch distributions to compute the posterior probability of enrichment of zeros for each gene.

For 300,000 cells and 30,000 genes the process takes about 10 seconds with 16 cores of CPU, or about 3 seconds with a Tesla T4 GPU.

I added the method as a `mode` called `'poisson_zeros'` in `GeneExpressionDataset`. I included it in the tests listed in `test_highly_variable_genes.py`.

For a demonstration of the method, see this notebook: https://nbviewer.jupyter.org/gist/vals/adeb32f2be95333166347173b03f185e

image

I hope we can add this to the code!

Best,
/Valentine


adamgayoso:
I can review this soon. Looks cool. Out of curiosity, have you tried the seurat v3 method with a sparse matrix? It doesn't sparsify X if it's already dense, so you'd have to do it manually. I believe it should be very fast. I also just updated this to use the loess implementation (that's also used in plotnine and written in C) and tested that it gives the same outputs as seurat v3.

scvi/dataset/dataset.py
```diff
@@ -1409,6 +1412,11 @@ def _highly_variable_genes(
 
         logger.info("extracting highly variable genes using {} flavor".format(flavor))
 
+        if flavor == "poisson_zeros":
+            var = poisson_gene_selection(self.X, n_top_genes, **highly_var_genes_kwargs)
+
+            return var
+
```
adamgayoso, line 1419: It makes sense to me to put this below seurat v3

adamgayoso: We can then let line 1426 `counts=self.X.copy()` be `counts=sp_sparse.csr_matrix(self.X)` (maybe need a copy if it's already sparse?)

and then your new function can use the adata with potential batch information.

I think all the functions will be faster using sparse matrices.

adamgayoso: though maybe it's overkill to have your new function take an anndata object? the seurat v3 one really just needs X and the batch info. I guess it only really depends if you think people will want to access this function outside of `subsample_genes`? If not maybe the anndata is fine for now. We have someone coming to work on the codebase and he can refactor this kind of stuff.

adamgayoso: then your function doesn't need to return anything and the return value of `adata.var` would be consistent with the other functions.

scvi/dataset/dataset.py
```diff
@@ -1468,6 +1476,105 @@ def get_batch_mask_cell_measurement(self, attribute_name: str):
...
+        torch.cuda.empty_cache()
+
+    # Create result DataFrame
+
+    df = pd.DataFrame({
+        'observed_fraction_zeros': obs_frac_zeros,
+        'expected_fraction_zeros': exp_frac_zeros,
+        'prob_zero_enrichment': prob_zero_enrichment,
+        'highly_variable': False
+    })
```
adamgayoso, line 1571: My other general thought is that it would be great if we could have this run per batch (RNA batch not gpu batch) and then sort by how many batches it’s highly variable followed by the prob zero enrichment. You can follow what is done in the seurat v3 code.

scvi/dataset/dataset.py
```diff
@@ -1468,6 +1476,105 @@ def get_batch_mask_cell_measurement(self, attribute_name: str):
...
+            of enrichment of zeros for each gene. Default: ``10000``.
+        :param silent: If ``True``, disables the progress bar. Default ``False``.
+        :param batch_size: Size of temporary matrix for incremental calculation. Larger
+            is faster but requires more RAM or GPU memory. (The default should be fine unless
+            there are hundreds of millions cells or millions of genes.) Default ``5000``.
+
+        :return: A ``pd.DataFrame`` with per-gene statistics.
+
+    """
+    from tqdm import tqdm
```
adamgayoso, line 1511: btw you can import tqdm like in trainer.py if you want it to adapt to jupyter notebooks (unless you like the simpler look). In trainer we also direct it to stdout to avoid the red stuff in jupyter.

adamgayoso:
also this is a general question, could this have been implemented with numba? And then you wouldn't need to store any matrix in memory? Everything would just be vectors?

scvi/dataset/dataset.py
vals: Oh that's how you get rid of the red stuff! Will change that. Yeah I prefer the text output because it doesn't go away when you re-open the notebook.. It always looks like something is wrong with the message about a missing javascript thing.

scvi/dataset/dataset.py
vals: > though maybe it's overkill to have your new function take an anndata object? the seurat v3 one really just needs X and the batch info. I guess it only really depends if you think people will want to access this function outside of `subsample_genes`? If not maybe the anndata is fine for now. We have someone coming to work on the codebase and he can refactor this kind of stuff.

I was thinking the overhead of creating the anndata wouldn't be worth it. Though it's probably very minor. I implemented it like this first and used it outside of scVI. But then I figured since scVI requires PyTorch it would fit well as an internal function, and I figured it wouldn't need more than the X.

scvi/dataset/dataset.py
vals: I was thinking about this too when I looked at the tests. I'll give it a try and see how it goes.

vals:
> I can review this soon. Looks cool. Out of curiosity, have you tried the seurat v3 method with a sparse matrix? It doesn't sparsify X if it's already dense, so you'd have to do it manually. I believe it should be very fast. I also just updated this to use the loess implementation (that's also used in plotnine and written in C) and tested that it gives the same outputs as seurat v3.

I always use sparse matrices, but I've never gotten the 'seurat_v3' mode to run without errors. (These might be related, I always use CSC matrices, could that be the problem?)


> also this is a general question, could this have been implemented with numba? And then you wouldn't need to store any matrix in memory? Everything would just be vectors?

I'm not so familiar with Numba. Can it do the summing of outer products without storing the blocks I create now?

Since scVI requires both Numba and PyTorch, which is preferred for computation? Personally I like using PyTorch since I'm always running on GPU instances and I like to take advantage of that.

I implemented the summary statistic calculation in NumPy as well, switching to PyTorch gave a ~5x speed increase on CPU (I guess because of multithreading).

I see the tests are failing, I think I know why: the indices for the minibatches are probably getting calculated wrong when the number of cells are smaller than the batch size. I'll see if I can fix it.

adamgayoso:
> I always use sparse matrices, but I've never gotten the 'seurat_v3' mode to run without errors. (These might be related, I always use CSC matrices, could that be the problem?)\

Have you tried with the latest release? I'm using the same code that I have in a PR at scanpy, which is passing their CI. There was a bug in the earlier code on sparse matrices (before 0.6.5), I also updated the loess implementation.


> I'm not so familiar with Numba. Can it do the summing of outer products without storing the blocks I create now?

Yeah it seems like since you're just taking the row sum after the outer product (if I'm reading correctly the outer product gives genes by cells?) you can use elementary for loops. Numba is pretty easy because you can just use numpy and for loops and numpy random binomial seems to work in numba. It's also pretty fun! You just add a decorator to the function.

```python
@numba.njit(cache=True)
def poisson_gene_selection(...)
```
as long as you use numpy for everything it will be happy.
```python
    expected_fraction_zeros = np.zeros(scaled_means.shape)
    for i in range(n_cells):
        expected_fraction_zeros += total_counts[i] * scaled_means
```
I don't think numba likes tqdm though.

> Since scVI requires both Numba and PyTorch, which is preferred for computation? Personally I like using PyTorch since I'm always running on GPU instances and I like to take advantage of that.

I'm not sure maybe @romain-lopez has a preference. I guess if numba actually reduces the memory footprint then that would be preferable no?

scvi/dataset/dataset.py
adamgayoso: Yeah I think we should change it back, I agree it's a bit more annoying than good

scvi/dataset/dataset.py
```diff
@@ -1468,6 +1476,105 @@ def get_batch_mask_cell_measurement(self, attribute_name: str):
...
+    # Calculate probability of zero for a Poisson model.
+    # Perform in batches to save memory.
+    n_batches = total_counts.shape[0] // batch_size
+
+    expected_fraction_zeros = torch.zeros(scaled_means.shape).to(dev)
+
+    for i in range(n_batches):
+        total_counts_batch = total_counts[i * batch_size:(i + 1) * batch_size]
+        # Use einsum for outer product.
+        expected_fraction_zeros += torch.exp(-torch.einsum('i,j->ij', [scaled_means, total_counts_batch])).sum(1)
```
adamgayoso, line 1535: @vals I'm realizing now... is this just a sum across axis1 of the full outer product? If so, couldn't you just multiply the column vector by the sum of the row vector to get the final result? If so, we don't have to worry about numba, etc.

vals: Hey Adam, I haven't had time to get to the other points yet, but I did a benchmark with numba and it was a fair bit slower even when cached. (~1 minute vs ~3 seconds for PyTorch with CPU)

The problem is I need to exponentiate before I sum (if I understand what you mean right). I was trying to find if there was some "opposite" function of `logsumexp` but didn't find anything.

adamgayoso: ah ok, perhaps what you have is fine then. Sorry for the sidetracking I didn't see the exp^{-x}.

adamgayoso:
@vals we are refactoring the dataset module to basically only take anndata as input. I'm going to hold off on this until we have that done, and then we can add this feature. I imagine this function will then have to be something that operates directly on anndata.

vals:
Sure, sounds good. Since commit 107f4aa it operates on an AnnData like the other methods. So should be quick to reroute what it calculates on.

scvi/dataset/dataset.py
```diff
@@ -1541,20 +1547,25 @@ def _highly_variable_genes(
...
                 **highly_var_genes_kwargs,
             )
+
         elif flavor == "seurat_v3":
             seurat_v3_highly_variable_genes(
                 adata, n_top_genes=n_top_genes, batch_key=batch_key
             )
+
+        elif flavor == "poisson_zeros":
+            poisson_gene_selection(adata, n_top_genes, **highly_var_genes_kwargs)
```
adamgayoso, line 1557:
```suggestion
            poisson_gene_selection(adata, n_top_genes, batch_key=batch_key, **highly_var_genes_kwargs)
```

Issue #660, opened 2020-04-24
Adding compatibility with biom-format


mortonjt:
Biom-format is a commonly used file format for microbiome studies.
It boils down to an hdf5 format of scipy.sparse.csr_matrix that also includes sample / gene metadata.

Paper of the file format is here: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3626512/

Let me know if there is any interest in this. I've already implemented the [torch.dataset](https://github.com/mortonjt/scVI/pull/1) and will be happy to push in a PR.

romain-lopez:
Hi,

Thanks for your contribution! I am not familiar with biomformat. Do people use this for scRNA-seq or is it some other kind of data? If so, why is it of interest to the users of scVI? Should we add microbiome datasets URLs to the codebase, as well as a notebook of how to analyze this data with scVI? 

I'd be happy to add this to the codebase provided this will be useful.

Best, 
Romain

mortonjt:
I haven't seen this used for scRNA-seq, but I see no reason why it can't be done.  The biom-format just provided a compressed format for storing sparse matrices (in addition to accompanying feature / sample metadata).

I don't see any problem including microbiome datasets + notebooks.  Is there a preferred location for storing datasets? (i.e. zenodo?)

romain-lopez:
Hi, 

Usually we add the datasets in the scVI-data github repo https://github.com/YosefLab/scVI-data but naturally if those datasets are already hosted online and can be fetched with a wget this is definitely the preferred method (you can look into the Downloadable dataset class). We do this for example for 10x datasets.

We now have a contributed notebook section in readthedocs, so if you want to show how a model of our codebase can be applied to microbiome we will have a look and post it to Colab. One example is https://scvi.readthedocs.io/en/stable/contributed_tutorials/scVI_DE_worm.html

Best, 
Romain

adamgayoso:
I would be happy to integrate this in scVI but a few points:

- we are going to refactor the dataset module a bit to have simple `read_biom(path)` functions
- we should not make biom a strict requirement, and can instead just catch an `ImportError`
- we should have a tutorial describing how to use scVI with microbiome data

As @romain-lopez said, it can be a contributed tutorial. @mortonjt, would you be able to make a simple tutorial like the LDVAE one? As well as make the PR after we refactor the dataset module (< 1 month time)?

mortonjt:
Hopefully it shouldn't take a month XD. 
[...]

PR #675, opened 2020-05-27
Biom format compatability

mortonjt:
This addresses #660 and adds compatibility to the biom-format.

The following have been addressed
- [x] Tutorial is added show inghow LDVAE can be used to analyze microbiome data.
- [x] biom is an optional dependency
- [x] a read_biom function is added to locally load biom tables

There are some slight hiccups regarding how to keep track of the sample names (see 2nd to last code block in the notebook). This will likely be a recurring issue for users and may need to be addressed in a future pull request.


galenxing:
Hey @mortonjt, 

Thanks for the PR! We're currently doing a major refactor of our Datasets so once it's finished, we can review and port this Biom compatibility over.

adamgayoso:
@mortonjt we are refactoring the dataset module to only use anndata. We can add a convenience method to read biom data and make it into anndata format. I'm just wondering, is this the most popular file type for microbiome data? Should the `scvi.dataset.read_biom(filename, backup_url=...)` function have any other functionality? Thanks!

mortonjt:
Yes, I'd say that biom is the most common file format for sparse microbiome count tables. It looks like the data structures behind anndata are very similar to biom - I'm almost wondering if a `to_anndata` method would be appropriate to add directly into the biom-format repository.

One thing that I noticed with previous versions of scvi is that scvi loses information about the sample name (in your case the cell id).  It would make it more seamless to retain both row / column identifiers, addition to row / column metadata.

adamgayoso:
> I'm almost wondering if a to_anndata method would be appropriate to add directly into the biom-format repository.

This seems like a better solution to me as it opens up the possibility of using many more single-cell algorithms with microbiome data. Anndata has the capability to preserve all row and column identifiers and associated metadata.

adamgayoso:
Hi @mortonjt, the time has now come. It would be extremely helpful if you refactored your tutorial to

1. Work in google colab (don't worry about it being tested with pytest)
2. Use the new 1.0 API
3. Submitted a PR to the contributed folder in [this repo](https://github.com/YosefLab/scvi_tutorials)

To see how the new API works please look at the quick start tutorial [here](https://scvi--750.org.readthedocs.build/en/750/user_guide/index.html).

Full API documentation is also available on that page. Note that for LDVAE it's now `scvi.models.LinearSCVI`.

mortonjt:
Great!!  I'll make sure to take another look at this (it'll probably happen
in ~3 weeks due to ongoing manuscript submissions).

adamgayoso:
Thanks! I'm going to close this PR for now.

mortonjt:
Hi @adamgayoso , I'm taking another look at this; in order for the tutorial to work, we'd still need to have a BiomDataset object right? The PR you made to Biom enables the conversion to anndata, but that object will still need to be read in through the biom api, right?

adamgayoso:
@mortonjt, the tutorial will not be tested, so it should work entirely in google colab. You can use the first cell in our user guide notebooks as an example. 

> we'd still need to have a BiomDataset object right? The PR you made to Biom enables the conversion to anndata, but that object will still need to be read in through the biom api, right?

Correct, the first lines of the tutorial would download the biom file and load it with the biom api and then convert to anndata, or whatever is the most conventional workflow for this type of data at first.

PR #680, opened 2020-06-05
modified models/scanvi.py for symmetrical KL option

chenlingantelope:


scvi/models/scanvi.py
```diff
@@ -204,18 +204,28 @@ def forward(self, x, local_l_mean, local_l_var, batch_index=None, y=None):
                 0.0,
             )
 
-        probs = self.classifier(z1)
+        probs = self.classify(x)
```
galenxing, line 207: Why the change to self.classify?

adamgayoso: It looks like it's for the label_groups thing, though it seems like the implementation is inefficient because it has to encode `x` again.

chenlingantelope: Yes this seems like a typo to me, because there is the classify function that deals with both when there is hierarchy information (weight the classifier results by the ancestor probabilities) or not (just returns the results of the classifier)

romain-lopez:
I am unsure whether this is needed anymore. Closed it for now.

PR #686, opened 2020-06-17
WIP: dataset refactor

galenxing:
This is the current status of the dataset refactor. 

The new workflow to train the model:
```
#Temporary example with cortex dataset, but should work with any anndata. 
#This will be refactored later
gene_dataset = CortexDataset(save_path= "data/", total_genes=None)
gene_dataset.subsample_genes(1000, mode="variance")
gene_dataset.make_gene_names_lower()

adata = gene_dataset.to_anndata()
setup_anndata(adata,batch_key='batch_indices', labels_key = 'precise_labels' )

vae = VAE(adata.shape[1], n_batch = adata.uns['scvi_summary_stats']['n_batch'])
trainer = UnsupervisedTrainer(
    vae,
    adata,
    train_size=0.90,
    n_epochs_kl_warmup = 2,
    use_cuda=False,
    frequency=5,
)
trainer.train(n_epochs =10)

#notice we pass in an anndata object to create_posterior now
full = trainer.create_posterior(trainer.model, adata, indices=np.arange(len(gene_dataset)))

#DE should work as usual like the intro notebook

```

Summary of changes:
- scVI will be working with anndata objects instead of boutique Dataset classes
- User just passes an anndata obj to Trainer, which under the hood,  creates a BioDataset which is a lightweight wrapper around an anndata object
- BioDataset `_getitem_()` returns a dict of tensors instead of list
- User can save an anndata after `setup_anndata()` and then load it into the model without setting it up again (unsure if we should encourage this behavior though)
- The posterior will check whether all the assets required by a model exist in the anndata data registry (not yet implemented)

Explanation of `setup_anndata()`:
- Data can be in anndata.X or anndata.layers (in which case you pass in `X_layers_key`)
- Calculates log mean and log var per batch
- if no batch_key provided, will assume everything is a single batch
- if no label_key provided, will assume everything is a single label
- "registers" the anndata by adding a dictionary mapping assets required by the models to their respective locations in the anndata object. This dictionary is saved to `anndata.uns['scvi_data_registry']`. Example dictionary: `{"X": (None, "X"), "batch" :("obs", "batch_idx")}`
- Saves some summary statistics to `anndata.uns['scvi_summary_stats']`


scvi/dataset/utils.py
```diff
@@ -0,0 +1,286 @@
...
+    log_counts = masked_log_sum.filled(0)
+    local_mean = (np.mean(log_counts).reshape(-1, 1)).astype(np.float32)
+    local_var = (np.var(log_counts).reshape(-1, 1)).astype(np.float32)
+    return local_mean, local_var
+
+
+def compute_library_size_batch(
+    adata,
+    batch_key: str,
+    local_l_mean_key: str = None,
```
adamgayoso, line 88:
```suggestion
    local_l_mean_key_added: str = None,
```

scvi/dataset/utils.py
```diff
@@ -0,0 +1,286 @@
...
+    local_l_var_key: str = None,
```
adamgayoso, line 89:
```suggestion
    local_l_var_key_added: str = None,
```

PR #689, opened 2020-06-23
Fix scanVI dataload

### LINKED ISSUE #688: Improper semi-supervised dataloader behavior
After a code formatting commit, the default behavior of scanVI was modified, causing the data labels to be used as observations (instead of rv) when computing the ELBO.

scanVI's trainer inherits from `UnsupervisedTrainer`. 
This implies that this change [here](https://github.com/YosefLab/scVI/blob/master/scvi/inference/inference.py#L123) feeds label information to the `forward` loop of scanVI's trainer even from unlabelled data.

In addition to fixing this bug, we should define a more proper behavior for `scanVI`'s forward method to avoid similar bugs in the future.

PierreBoyeau:
Fixes #688

### DIFF
--- a/scvi/inference/annotation.py
+++ b/scvi/inference/annotation.py
@@ -284,7 +284,7 @@ class SemiSupervisedTrainer(UnsupervisedTrainer):
         super().__setattr__(key, value)
 
     def loss(self, tensors_all, tensors_labelled):
-        loss = super().loss(tensors_all)
+        loss = super().loss(tensors_all, feed_labels=False)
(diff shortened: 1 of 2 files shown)


galenxing:
Hey @PierreBoyeau, not sure I entirely understand the issue, but I think the change should be implemented in scanVI's `forward` or `inference` functions. 

This change right here seems to break our other models.

romain-lopez:
I don't know what is blocking this test but this change is super important @galenxing can we fix this ASAP ?

PR #691, opened 2020-06-26
train_size fix, bump versions, remove old reproducibility notebook

### LINKED ISSUE #669: train_size=1 is ambiguous
Our docstring clearly states that if you give train_size an int, it will take that many cells in the training set, but 1 is an ambiguous case as 1.0 means train on the whole dataset and 1 means train on one cell. A caveat here is that scVI skips minibatches of < 3 cells. 

We should restrict this option to only allow float between 0.0 and 1.0. Also, we need a warning statement of the training size is really low.

galenxing:
fixes #669


scvi/inference/inference.py
```diff
@@ -78,13 +74,18 @@ def __init__(
         self,
         model,
         gene_dataset: GeneExpressionDataset,
-        train_size: Union[int, float] = 0.8,
+        train_size: Union[int, float] = 0.9,
```
adamgayoso, line 77: I don't have an issue with this as the default will be "good" but @romain-lopez what do you think about the bump from 0.8 to 0.9?

scvi/inference/inference.py
```diff
@@ -78,13 +74,18 @@ def __init__(
...
         test_size: Union[int, float] = None,
         n_iter_kl_warmup: Union[int, None] = None,
         n_epochs_kl_warmup: Union[int, None] = 400,
         normalize_loss: bool = None,
         **kwargs
     ):
+        train_size = float(train_size)
+        if train_size > 1 or train_size <= 0:
```
adamgayoso, line 85:
```suggestion
        if train_size > 1.0 or train_size <= 0.0:
```

romain-lopez:
Can we put the same value for the whole codebase (I see 0.8 and 0.9 in this PR)? and yes, both 0.8 and 0.9 are fine by me

tests/test_notebook.py
```diff
@@ -92,11 +92,7 @@ def load_module(self, fullname):
                             "save_path = '" + os.getcwd() + "'",
                             code,
                         )
-                        code = re.sub(
-                            'save_path = "data/"',
-                            "save_path = '" + os.getcwd() + "'",
-                            code,
-                        )
```
adamgayoso, line 99: I think both needed to be here because some notebooks use " while others use '. Sorry!

galenxing: oh woah did NOT notice that lol.

PR #692, opened 2020-06-26
bump versions

galenxing:
testing travis

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,18 +11,18 @@ with open("HISTORY.rst") as history_file:
 
 requirements = [
     "numpy>=1.16.2",
-    "torch>=1.0.1",
+    "torch>=1.3",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -11,18 +11,18 @@
...
     "h5py>=2.9.0",
-    "pandas>=0.24.2",
+    "pandas>=1.0.5",
     "loompy>=3.0.6",
     "tqdm>=4.31.1",
     "xlrd>=1.2.0",
     "hyperopt==0.1.2",
-    "anndata>=0.7",
+    "anndata>=0.7.3",
     "statsmodels",
```
adamgayoso, line 24:
```suggestion

```

adamgayoso: we don't need this anymore

PR #722, opened 2020-08-13
biodataset creates a data dict attr to reduce anndata access

adamgayoso:


scvi/dataset/_biodataset.py
```diff
@@ -104,6 +116,20 @@ def setup_getitem(self, getitem_tensors: Union[List[str], Dict[str, type]] = Non
 
         self.attributes_and_types = keys_to_type
 
+    def __getitem__(self, idx: int) -> Dict[str, torch.Tensor]:
+
+        data_numpy = {}
+        for key, dtype in self.attributes_and_types.items():
+            data = self.data[key]
```
adamgayoso, line 123:
```suggestion
            data_numpy[key] = self.data[key]
```

PR #726, opened 2020-08-16
Rich log

adamgayoso:
Screen Shot 2020-08-16 at 12 33 09 PM

### DIFF
--- a/scvi/_settings.py
+++ b/scvi/_settings.py
@@ -1,6 +1,6 @@
 import logging
-import sys
 from typing import Union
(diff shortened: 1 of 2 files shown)


adamgayoso:
On another note, @galenxing the warning about more genes than cells should probably come upon Posterior object creation because it will be model specific.

adamgayoso:
And in jupyter -- I also decided to get rid of the time

Screen Shot 2020-08-16 at 10 48 56 PM

PR #740, opened 2020-09-08
Assorted fixes

galenxing:
Scanvi predict, model init with view

### DIFF
--- a/scvi/core/models/__init__.py
+++ b/scvi/core/models/__init__.py
@@ -664,9 +664,7 @@ class BaseModelClass(ABC):
                 logger.info("Received view of anndata, making copy.")
                 adata = adata.copy()
             else:
-                raise ValueError(
-                    "Please run `adata = adata.copy()`"
-                )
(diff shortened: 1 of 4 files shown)


scvi/core/models/__init__.py
```diff
@@ -660,10 +660,14 @@ def _validate_anndata(
...
             adata = self.adata
         if adata.is_view:
-            logger.warning("Input anndata is a view.")
             if copy_if_view:
-                logger.info("Making copy of anndata.")
+                logger.info("Received view of anndata, making copy.")
                 adata = adata.copy()
+            else:
+                raise ValueError(
+                    "Please run `adata = adata.copy()` or set copy_if_view to be True."
```
adamgayoso, line 668:
```suggestion
                    "Please run `adata = adata.copy()`"
```
I think this is more catered toward users who will not have this option (copy_if_view)

adamgayoso:
Let's add an info statement to setup anndata to not modify after, registered keys should be debug level.

PR #742, opened 2020-09-09
use custom tqdm

### LINKED ISSUE #741: tqdm repeating itself
Seems related to this [issue](https://github.com/tqdm/tqdm/issues/375)

Screen Shot 2020-09-08 at 5 27 45 PM

adamgayoso:
Fixes #741 

Still a minor issue (see below), but much better than before.

### DIFF
--- a/scvi/_compat.py
+++ b/scvi/_compat.py
@@ -13,3 +13,15 @@ except ImportError:
 
         class Literal(metaclass=LiteralMeta):
             pass
+
+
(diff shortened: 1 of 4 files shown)


adamgayoso:
This prevents the issue, but the bar from the run that gets stopped (and originally causes the problem -- stopping in progress -- stays behind)

Screen Shot 2020-09-08 at 5 48 11 PM

galenxing (approved):

PR #749, opened 2020-09-12
Custom progress bar with rich/tqdm styles and config

adamgayoso:
1. scvi-specific progress bar with "rich" and "tqdm" style option.
2. `silent` option to silent the bar
3. ScviConfig to control global settings like this, random seed, and batch size after training to get quantities

Screen Shot 2020-09-12 at 3 06 31 PM

### DIFF
--- a/scvi/core/trainers/trainer.py
+++ b/scvi/core/trainers/trainer.py
@@ -196,10 +196,6 @@ class Trainer:
         self.training_extras_end()
 
         self.training_time += (time.time() - begin) - self.compute_metrics_time
-        logger.info(
-            "Training time:  %i s. / %i epochs"
-            % (int(self.training_time), self.n_epochs)
-        )
         self.on_training_end()
 
     def on_training_loop(self, tensors_dict):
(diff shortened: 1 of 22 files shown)


scvi/core/trainers/trainer.py
```diff
@@ -173,11 +172,8 @@ def train(self, n_epochs=400, lr=1e-3, eps=0.01, params=None, **extras_kwargs):
 
         self.on_training_begin()
 
-        for self.epoch in tqdm(
-            range(n_epochs),
-            desc="training",
-            disable=not self.show_progbar,
-            file=sys.stdout,
+        for self.epoch in track(
+            range(n_epochs), description="Training...", disable=not self.show_progbar
```
galenxing, line 176: Thoughts on changing `self.show_progbar` to `self.silent`?

galenxing (commented):
Looks good! Only this is when there are multiple functions that have rich output in the same cell in Jupyter, the newer output overwrites the previous output.

scvi/core/trainers/trainer.py
adamgayoso: done

galenxing (approved):
LGTM

PR #755, opened 2020-09-16
deprecation warning

galenxing:
Screen Shot 2020-09-16 at 8 59 17 AM
Screen Shot 2020-09-16 at 9 00 19 AM


adamgayoso:
LGTM, the build error is just black, which was updated and we already took care of in scvi-tools.

adamgayoso (approved):

PR #772, opened 2020-09-19
initial fdr implementation

adamgayoso:


adamgayoso:
@romain-lopez is there anything stopping us from merging this?

romain-lopez:
I think this is great that a preliminary version is on the codebase and ready to go, so that other people can try it. Let's just have Pierre send some feedback on it and we can merge

PierreBoyeau (approved):
Hi Adam, thanks so much! This looks good to me. The name `fdr` is a bit confusing, and `fdr_target` makes more sense to me. I worked on developing new ways to better control FDR, but they still need further empirical prooftesting before we can include them to the codebase.

PR #773, opened 2020-09-19
adding .items() to iterate through dict in scvi/data/_utils.py line 191

TheAustinator:
Error raised during `model = scvi.model.SCVI(adata)`


adamgayoso:
Thanks for catching this, perhaps we should add a test in which extra covariates are setup and then an scvi model fit is performed?

adamgayoso (approved):
LGTM, we can address the testing when we actually implement this.

PR #793, opened 2020-09-24
use np.ndarray instead of pd.Index for categorical

### LINKED ISSUE #792: Error when saving h5ad
I think this may be a bug and apologize if I miss anything. I am trying to save the adata (adata.write("...")) after running through the scvi pipeline and got this error. Deleting adata.uns['_scvi'] solves the issue.

```pytb
NotImplementedError: Failed to write value for uns/_scvi/categorical_mappings/_scvi_batch/mapping, since a writer for type <class 'pandas.core.indexes.base.Index'> has not been implemented yet.

Above error raised while writing key 'uns/_scvi/categorical_mappings/_scvi_batch/mapping' of <class 'h5py._hl.files.File'> from /.
```

adamgayoso:
Fixes #792 

Index cannot be written with anndata

### DIFF
--- a/scvi/data/_anndata.py
+++ b/scvi/data/_anndata.py
@@ -519,7 +519,7 @@ def _setup_extra_categorical_covs(
             possible_cats = category_dict[key]
             cat = cat.astype(CategoricalDtype(categories=possible_cats))
         else:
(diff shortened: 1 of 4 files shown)


galenxing (approved):
LGTM

PR #798, opened 2020-09-25
Make data csr and C_CONTIGUOUS if sparse or  dense np respectively

### LINKED ISSUE #787: Ensure protein and X are "C" order if dense
I already did this if using a layer, but we should always check whenever it's a dense numpy array.

In the case of proteins, if it's a dataframe we may need to overwrite the obsm with a new dataframe.

galenxing:
Fixes #787

### DIFF
--- a/scvi/data/_anndata.py
+++ b/scvi/data/_anndata.py
@@ -319,13 +319,13 @@ def _verify_and_correct_data_format(adata, data_registry):
             data = np.asarray(data, order="C")
             _set_data_in_registry(adata, data, k)
         elif isinstance(data, pd.DataFrame) and (
-            data.values.flags["C_CONTIGUOUS"] is False
+            data.to_numpy().flags["C_CONTIGUOUS"] is False
         ):
             logger.debug(
                 "{} is not C_CONTIGUOUS. Overwriting to C_CONTIGUOUS.".format(k)
             )
             index = data.index
-            vals = data.values
+            vals = data.to_numpy()
             columns = data.columns
             data = pd.DataFrame(
                 np.ascontiguousarray(vals), index=index, columns=columns
--- a/scvi/data/_anndata.py
+++ b/scvi/data/_anndata.py
@@ -308,7 +308,7 @@ def _verify_and_correct_data_format(adata, data_registry):
 
     for k in keys:
         data = get_from_registry(adata, k)
-        if isspmatrix(data) and (data.getformat() == "csc"):
+        if isspmatrix(data) and (data.getformat() != "csr"):
             data = data.tocsr()
             _set_data_in_registry(adata, data, k)
         elif isinstance(data, np.ndarray) and (data.flags["C_CONTIGUOUS"] is False):
--- a/scvi/data/_anndata.py
+++ b/scvi/data/_anndata.py
@@ -251,6 +251,22 @@ def setup_anndata(
 
 
 def _set_data_in_registry(adata, data, key):
+    """
+    Sets the data associated with key in adata.uns['_scvi']['data_registry'].keys() to data.
+
+    Note: This is a dangerous method and will change the underlying data of the user's anndata
+    Currently used to make the user's anndata C_CONTIGUOUS and csr if it is dense numpy
+    or sparse respectively.
+
+    Parameters
+    ----------
+    adata
+        anndata object to change data of
+    data
+        data to change to
+    key
+        key in adata.uns['_scvi]['data_registry'].keys() associated with the data
+    """
     use_raw = adata.uns["_scvi"]["use_raw"]
     data_loc = adata.uns["_scvi"]["data_registry"][key]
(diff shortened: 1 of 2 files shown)


tests/dataset/test_anndata.py
```diff
@@ -99,6 +100,41 @@ def test_transfer_anndata_setup():
...
+    old_X = adata.X
+    old_pro = adata.obsm["protein_expression"]
+    old_obs = adata.obs
+    adata.X = np.asfortranarray(old_X)
+    adata.obsm["protein_expression"] = np.asfortranarray(old_pro)
+    assert adata.X.flags["C_CONTIGUOUS"] is False
+    assert adata.obsm["protein_expression"].flags["C_CONTIGUOUS"] is False
+
+    setup_anndata(adata, protein_expression_obsm_key="protein_expression")
+    assert adata.X.flags["C_CONTIGUOUS"] is True
```
adamgayoso, line 116: let's make sure using get from registry gives the same results

scvi/data/_anndata.py
```diff
@@ -240,12 +241,77 @@ def setup_anndata(
...
+    adata
+        anndata to check
+    data_registry
+        data registry of anndata
+    """
+    for k in data_registry.keys():
+        data = get_from_registry(adata, k)
+        if isspmatrix(data):
+            data = data.tocsr()
+        elif isinstance(data, np.ndarray):
```
adamgayoso, line 310: I think we are missing the case if it's a DataFrame here, like for the proteins

scvi/data/_anndata.py
```diff
@@ -240,12 +241,77 @@ def setup_anndata(
...
+            attribute.loc[:, attr_key] = data
+        else:
+            attribute[attr_key] = data
+        setattr(tmp_adata, attr_name, attribute)
+
+    if use_raw is True and attr_name in ["X", "var"]:
+        adata.raw = tmp_adata
+
+
+def _verify_and_correct_data_format(adata, data_registry):
```
adamgayoso, line 293: I feel like we should only care about X and proteins for now, maybe a parameter here for which keys to check? If none, check all?

tests/dataset/test_anndata.py
adamgayoso: I think I had noticed a weird issue where you set adata.X of a registered object and get from registry still references the old object

scvi/data/_anndata.py
```diff
@@ -240,12 +241,92 @@ def setup_anndata(
...
+        anndata to check
+    data_registry
+        data registry of anndata
+    """
+    keys_to_check = [_CONSTANTS.X_KEY, _CONSTANTS.PROTEIN_EXP_KEY]
+    keys = [key for key in keys_to_check if key in data_registry.keys()]
+
+    for k in keys:
+        data = get_from_registry(adata, k)
+        if isspmatrix(data) and (data.getformat() == "csc"):
```
adamgayoso, line 311:
```suggestion
        if isspmatrix(data) and (data.getformat() != "csr"):
```

scvi/data/_anndata.py
```diff
@@ -240,12 +241,92 @@ def setup_anndata(
...
+    Will iterate through all the keys of data_registry.
+
+    Parameters
+    ----------
+    adata
+        anndata to check
+    data_registry
+        data registry of anndata
+    """
+    keys_to_check = [_CONSTANTS.X_KEY, _CONSTANTS.PROTEIN_EXP_KEY]
```
adamgayoso, line 306: what happens if protein exp isn't registered?

tests/dataset/test_anndata.py
adamgayoso: I would still add one line to the test to make sure.

scvi/data/_anndata.py
galenxing: The next line will check if its in data_registry

adamgayoso: ah, didn't see

scvi/data/_anndata.py
```diff
@@ -240,12 +241,98 @@ def setup_anndata(
...
+            data = data.tocsr()
+            _set_data_in_registry(adata, data, k)
+        elif isinstance(data, np.ndarray) and (data.flags["C_CONTIGUOUS"] is False):
+            logger.debug(
+                "{} is not C_CONTIGUOUS. Overwriting to C_CONTIGUOUS.".format(k)
+            )
+            data = np.asarray(data, order="C")
+            _set_data_in_registry(adata, data, k)
+        elif isinstance(data, pd.DataFrame) and (
+            data.values.flags["C_CONTIGUOUS"] is False
```
adamgayoso, line 322: values or to_numpy?

galenxing: seems like values is deprecated. just changed to to_numpy()

adamgayoso (approved):

PR #799, opened 2020-09-25
Cleanup distributions, add NegativeBinomialMixture

adamgayoso:
- Exposes our custom distributions to public through `scvi.core.distributions`
- New NegativeBinomialMixture distribution


scvi/core/distributions/_negative_binomial.py
```diff
@@ -0,0 +1,523 @@
...
+    Parameters
+    ----------
+    total_count
+        Number of failures until the experiment is stopped.
+    probs
+        The success probability.
+    mu
+        Mean of the distribution.
+    theta
+        Inverse dispersion.
```
galenxing, line 256: Lets have docstings for all the params

scvi/core/distributions/_negative_binomial.py
```diff
@@ -0,0 +1,523 @@
...
+    zi_logits
+        Logits scale of zero inflation probability.
```
galenxing, line 354: same, lets document all the params

PR #809, opened 2020-09-30
[WIP]: refactor fclayers for scarches

adamgayoso:


adamgayoso:
@M0hammadL, @Koncopd, @Cottoneyejoe95, this is what I was envisioning earlier. Now we can just load the state dict with `strict=False` and add a few more nodes to the parameter list. The implementation is a bit slower than our current code though and I'm trying to figure out why.

adamgayoso:
@M0hammadL, @Koncopd, @Cottoneyejoe95 please see #813 instead. I think using the hooks could be better as it doesn't disrupt a normal model training. It could be a little slower for the query dataset update as gradients will be computed on every linear layer that receives the categoricals

M0hammadL:
Hi Adam, thx a lot for pushing.   @Koncopd, @Cottoneyejoe95 would you please check?

 I will also look in detail at the weekend.

M0hammadL:
@adamgayoso  is this cover totalVI and gimVI too?

Koncopd:
@M0hammadL yes, i'll check.

adamgayoso:
@M0hammadL, @Cottoneyejoe95  thanks for sharing the data. The PR will cover totalVI eventually. gimVI is a bit more difficult and not sure we should implement it here.

PR #813, opened 2020-10-01
extension for online updates with scarches

adamgayoso:
See [this](https://colab.research.google.com/drive/1-LE3xgoh-O9mggbhutEKCb32iTEQsl2H?usp=sharing) colab notebook for an example with scVI and scANVI. See [this](https://colab.research.google.com/drive/1MSzfbaGCTOmuBGCRzvlGBg6RioxthTBa?usp=sharing) notebook for a totalVI example.

- [x] scvi
- [x] scanvi
- [x] totalvi
- [x] properly save var names


Cottoneyejoe95:
@adamgayoso , just had a quick look at your code changes and found some lines that werent exactly clear for me:

- It seemed like you added "inject-covariates" as a variable in FcLayers, which is similar to our "only_in_first" variable, but that variable doesnt get called in any parent class (e.g. Encoder, Decoder etc.) so I also cant call it in SCVI model.

- For us it would be also good if we had the option to add Condition weights (categories in your code) also for the encoder model. But the VAE class has no n_cat_list option for encoder or l_encoder. Additionally to that we then need changes for the get_latents function to also pass batch_index there as input and following to that also in your posterior evaluation class.

Koncopd:
@adamgayoso thank you very much for this.
Also if i understand correctly it is now impossible to use `protein_batch_mask` in `totalvi` for query data because [here](https://github.com/YosefLab/scvi-tools/blob/6b21a010e8c6d39b50d63eadbf2e552424a35dad/scvi/core/modules/totalvae.py#L478) it will infer uncorrect batches

if we have `batch_index=[3,3,3,4]`, then `np.arange(len(torch.unique(batch_index))) = [0, 1]` and for `b in [0, 1]` we won't have anything useful in `(batch_index == b)` or `self.protein_batch_mask[b]`

adamgayoso:
@Koncopd this is a great catch, I think I just fixed it. It would be helpful if you guys checked over all the new code!

Koncopd:
@adamgayoso 
Can't commit here so i made a fork
https://github.com/Koncopd/scvi-tools/tree/fclayers_hooks
There some fixes to get_latent and so on.

Also i don't understand why trainer doesn't save any history after `load_query_data` and `train`, `trainer.history` is empty.

https://github.com/theislab/scarchesp/blob/in_scvi/notebooks/scANVI_scvi_tools_surgery.ipynb - example with scanvi. For some reason results in lower overall accuracy than our pipeline - https://github.com/theislab/scarchesp/blob/in_scvi/notebooks/scANVI_surgery_pipeline.ipynb

adamgayoso:
@Koncopd this is helpful! would you be able to make a PR with this branch as the base branch?

adamgayoso:
> Also i don't understand why trainer doesn't save any history after load_query_data and train, trainer.history is empty.

I have to double check but I believe you need to set the frequency parameter to an integer.

Koncopd:
@adamgayoso pr - https://github.com/YosefLab/scvi-tools/pull/818

Koncopd:
Ok, setting frequency solved the problem with trainer history.

Koncopd:
@adamgayoso it seems that with `deeply_inject_covariates=False` the freezing doesn't work correctly. Only the conditional weights for the first layer should be unfrozen (and hooked). Now all weight even without conditions require grad.

I had to freeze them manually in the notebook.
https://github.com/theislab/scarchesp/blob/in_scvi/notebooks/scANVI_scvi_tools_surgery_1st_cond_layer.ipynb

The correct freezing conditions for scvi and scanvi are listed [here](https://github.com/theislab/scarchesp/blob/6b2c103a51d24307c05bd8a7b7af5be44d17e5b7/scarchest/surgery/scvi.py#L130)

Still with and without manual freezing i can't recreate working transfer learning comparing to
https://github.com/theislab/scarchesp/blob/master/notebooks/scANVI_surgery_pipeline.ipynb
Accuracy goes down in scANVI_scvi_tools_surgery_1st_cond_layer.ipynb even with correct manual freezing.

For now i have no ideas why we can't recreate our transfer learning example.

Koncopd:
Also forward pass is broken in fc_layer with `deeply_inject_covariates=False`.
fixed here - https://github.com/YosefLab/scvi-tools/pull/818/commits/5ffcd9e00292974953a8635d711d94024769458b

Cottoneyejoe95:
@adamgayoso in addition to @Koncopd it would be very good if you could use the following freezing option for testing your implementation:
https://github.com/theislab/scarchesp/tree/Lib_with_cond

The Architecture of the network is the following: 
Conditions in first layer of encoder/L-encoder/px-decoder. 
Everything freezed except the condition weights, so also the batchnorm freezed.

I think that would be a good debugging script as well since every new addition is tested in there and the result could be easily validated, since the performance for the reference data shouldnt change during retraining in this surgery option if everything is correct.

adamgayoso:
@Koncopd 
> it seems that with deeply_inject_covariates=False the freezing doesn't work correctly. Only the conditional weights for the first layer should be unfrozen (and hooked). Now all weight even without conditions require grad.

I'll be slow to respond this week, but I think as of right now, requires grad will be true for the subsequent layers, but the hooks will be set on these matrices (layers after 0) so it shouldn't affect anything.

I'm sure there are a few things that I'm not freezing correctly still. I'll look into your links soon.

> Also forward pass is broken in fc_layer with deeply_inject_covariates=False.

I should have fixed this.

adamgayoso:
@Cottoneyejoe95 

>  Everything freezed except the condition weights, so also the batchnorm freezed.

Do you change the affine or running stats options?

> I think that would be a good debugging script as well since every new addition is tested in there and the result could be easily validated, since the performance for the reference data shouldnt change during retraining in this surgery option if everything is correct.

Actually, since we use L2 regularization with adam, and are using hooks, the params require grad and might change a bit because of the L2 reg. We could set weight decay to 0 on the retrain with query to mitigate this.

Cottoneyejoe95:
> Do you change the affine or running stats options?

@adamgayoso This is how Im doing batch freezing, however the line  that checks for requires_grad does not work with your hooks, since its always true even when freeze_expression is True for layers with conditions.

def on_training_loop(self, tensors_dict):
        if self.model.freeze:
            for name, module in self.model.named_modules():
                if isinstance(module, nn.BatchNorm1d):
                    if not module.weight.requires_grad:
                        module.affine = False
                        module.track_running_stats = False
        self.current_loss = loss = self.loss(*tensors_dict)
        self.optimizer.zero_grad()
        loss.backward()
        self.optimizer.step()

> Actually, since we use L2 regularization with adam, and are using hooks, the params require grad and might change a bit because of the L2 reg. We could set weight decay to 0 on the retrain with query to mitigate this.

I had no problem with weight decay with my cond_layers class, but if the results change with the hooks when its not zero maybe it is necessary.

adamgayoso:
Ok, so @Cottoneyejoe95 basically to handle batchnorm we

1. Set requires grad = False to it's parameters in the modules we want to freeze
2. Under the req grad = False condition, also do

```
module.affine = False
module.track_running_stats = False
```

?

Cottoneyejoe95:
@adamgayoso yes, at least for layers without conditions that should work. For layers with conditions we have to think of another idea to tell the trainer that it should set affine and running_stats to False.

Also 

    module.affine = False
    module.track_running_stats = False

has to be set again after every call of model.train(), since that resets those stats, thats also why I put it in your on_loop_function

adamgayoso:
@Cottoneyejoe95 , does `.affine` do anything if requires grad was set to False for the params?

It seems like setting `momentum` to 0 could also give the same effect as `module.track_running_stats = False` without having to touch the trainer.

Cottoneyejoe95:
@adamgayoso the batchnorm layers were only freezed correctly when I set both to False. I didnt experiment with setting momentum to 0.

adamgayoso:
I really need to invite you guys to slack...

@Koncopd I fixed the hooks a bit in [`4ab9183`](https://github.com/YosefLab/scvi-tools/pull/813/commits/4ab9183946847186d00b28e0dd123da0731581ca). In the notebook in the very first comment of the PR, I updated it with `deeply_inject... = False` and `encode_covariates=True`. I get 86% accuracy when treating the query as unlabeled.

I'm sure there are still a few things not being freezed exactly the same as your implementation; I'll have to address it next week unless you all want to make a PR here again.

M0hammadL:
> I really need to invite you guys to slack...
> 
That would be great Adam
> @Koncopd I fixed the hooks a bit in [`4ab9183`](https://github.com/YosefLab/scvi-tools/pull/813/commits/4ab9183946847186d00b28e0dd123da0731581ca). In the notebook in the very first comment of the PR, I updated it with `deeply_inject... = False` and `encode_covariates=True`. I get 86% accuracy when treating the query as unlabeled.
> 
@Cottoneyejoe95  that is even higher to what you got [here](https://github.com/theislab/scarchesp/blob/master/notebooks/scANVI_surgery_pipeline.ipynb) right? 


> I'm sure there are still a few things not being frozen exactly the same as your implementation; I'll have to address it next week unless you all want to make a PR here again.
what else is still there?
a lot of engineering involved here :)

Cottoneyejoe95:
@M0hammadL I also had runs with 86% accuracy and even higher depending the surgery method. Normally its a trade off between query accuracy and full data accuracy. But the method I mentioned above I got about 84-87% accurarcy on query with no accuracy loss for full data.

M0hammadL:
> @M0hammadL I also had runs with 86% accuracy and even higher depending the surgery method. Normally its a trade off between query accuracy and full data accuracy. But the method I mentioned above I got about 84-87% accurarcy on query with no accuracy loss for full data.

but I consider what Adam got good results right? compared to what Sergey got 74 or sth?

Cottoneyejoe95:
@M0hammadL that 74% was also full data accuracy. Sergei also got 86% on query.

M0hammadL:
> @M0hammadL that 74% was full data accuracy. Sergei also got 86% on the query.

I donnow how much accuracy @adamgayoso  got on the whole data?

Koncopd:
@adamgayoso `scanvi_model.train` will always use semisupervised trainer, yes? So then it treated the query as labeled (but i set 0 for n_labelled_samples_per_class). If it is incorrect, how i use unsupervised  trainer for scanvi? Shouldn't setting n_labelled_samples_per_class to 0 help? Or should i set all labels to 'Unknown'?

Koncopd:
@adamgayoso could you please resend slack invitation to koncopd@gmail.com?

Koncopd:
https://github.com/theislab/scarchesp/blob/in_scvi/notebooks/scANVI_scvi_tools_surgery_1st_cond_layer.ipynb
I updated the notebook. Now the surgery seems to work fine.
image
One suspicious thing is that the nonconditional parts still change a bit. Why can it happen?

adamgayoso:
@Koncopd it can happen due to the weight decay (L2 regularization) in Adam, and that I use hooks. You should be able to pass `weight_decay` through to the trainer using our higher-level `train` method.

PR #818, opened 2020-10-06
Some fixes for fclayers_hooks

Koncopd:


scvi/model/scanvi.py
```diff
@@ -259,6 +259,10 @@ def train(
         self.trainer.labelled_set = self.trainer.create_scvi_dl(
             indices=self._labeled_indices
         )
+
+        for scdl in [self.trainer.labelled_set, self.trainer.unlabelled_set]:
+            scdl.to_monitor = ["reconstruction_error", "accuracy"]
+
```
adamgayoso, line 265: isn't this already done?

https://github.com/YosefLab/scvi-tools/blob/b519575a5469f1ec71250e06a8856e155bbf5c91/scvi/core/trainers/annotation.py#L187-L188

scvi/core/modules/scanvae.py
```diff
@@ -176,8 +176,8 @@ def classify(self, x, batch_index=None):
             w_y = self.classifier(z)
         return w_y
 
-    def get_latents(self, x, y=None):
-        zs = super().get_latents(x)
+    def get_latents(self, x, y=None, batch_index=None):
```
adamgayoso, line 179: These functions are not actually used anywhere, we could also consider removing them now.

scvi/model/scanvi.py
Koncopd: The problem is that it is done in trainer and then scanvi assigns datasets after trainer init without `to_monitor`.

scvi/core/modules/scanvae.py
Koncopd: We use them, but they can be replaced ofc.

scvi/model/scanvi.py
adamgayoso: lol, we really need to refactor all the semisupervised trainers

scvi/core/modules/scanvae.py
adamgayoso: are you guys able to use instead the API as shown in the colab notebooks I linked in the PR?

PR #821, opened 2020-10-07
Save/load fixes

### LINKED ISSUE #812: Loading models with correct adata in new API: option to save adata?
Hello, with the new API when loading a model, it is necessary to provide it with an adata file. In the old API, saving a model would also save the adata and it would be loaded automatically. Because for doing the gene selection now I restrict my adata to the genes that I use, the most failsafe way tt not mixing up the genes is to save this new adata for loading at a later point. 

Saving and loading a model in the new API (as per the documentation) is done with these two lines:
```
model.save("my_model/")
model = scvi.model.SCVI.load(adata, "my_model/", use_cuda=True)
```

This creates the folder `my_model` which has two files: `model_params.pt` and `attr.pkl`. 

What I end up doing is I also save the adata in the `my_model` folder as something like `my_adata.h5ad`. Is there a better way of doing this? It is only two extra lines of code, but I sometimes forget and then I can't reload my model because the if a single gene is different in gene selection process (which is probabilistic). 

If saving a separate adata is the recommended strategy, is there a reason why `model.save` doesn't have an option to automatically save the adata? It would be convenient if there was a flag `save_adata=True` what would save `my_model/adata.h5ad`

Thanks!

### LINKED ISSUE #816: Error when doing DE `AttributeError: 'NoneType' object has no attribute 'var_names'` from loaded model
Hello, I ran into the error below when trying to do DE from a model that was loaded using these 3 lines:

```
adata=anndata.read('./'+ model_name+'.model/adata.h5ad')
model = scvi.model.SCVI.load(adata, model_name+'.model', use_cuda=True)
de = model.differential_expression( groupby="leiden")
```

After some poking around, I copied these lines from the intro example, and then the DE ran with no problem:
```
adata.layers["counts"] = adata.X.copy() # preserve counts
adata.raw = adata # freeze the state in `.raw`
```
However, I find it odd that I had to run these lines again, because the anndata that I saved was the one that used to train the model. I had not realized that having the adata itself stored in `adata.raw` was necessary, but given the error, this seems to be the case?

image

```
AttributeError                            Traceback (most recent call last)
<ipython-input-125-eb1e0fcfe1d1> in <module>
      1 de = model.differential_expression(
----> 2     groupby="leiden",
      3 )
      4 de.head()

~/anaconda3/lib/python3.7/site-packages/scvi/core/models/rnamixin.py in differential_expression(self, adata, groupby, group1, group2, idx1, idx2, mode, delta, batch_size, all_stats, batch_correction, batchid1, batchid2, **kwargs)
    179         adata = self._validate_anndata(adata)
    180 
--> 181         col_names = _get_var_names_from_setup_anndata(adata)
    182         model_fn = partial(
    183             self.get_normalized_expression,

~/anaconda3/lib/python3.7/site-packages/scvi/model/_utils.py in _get_var_names_from_setup_anndata(adata)
    119         adata.var_names
    120         if adata.uns["_scvi"]["use_raw"] is False
--> 121         else adata.raw.var_names
    122     )
    123 

AttributeError: 'NoneType' object has no attribute 'var_names'
```

galenxing:
fixes #816 
fixes #812 

also fixes issue where cuda=False when trained with GPU issue
also removed use_raw support


scvi/core/models/base.py
```diff
@@ -150,7 +149,7 @@ def _get_init_params(self, locals):
         }
         return user_params
 
-    def save(self, dir_path: str, overwrite: bool = False):
+    def save(self, dir_path: str, overwrite: bool = False, save_anndata: bool = True):
```
adamgayoso, line 152: let's default this to False

scvi/core/models/base.py
```diff
@@ -179,23 +180,41 @@ def save(self, dir_path: str, overwrite: bool = False):
                     dir_path
                 )
             )
-        torch.save(self.model.state_dict(), os.path.join(dir_path, "model_params.pt"))
-        with open(os.path.join(dir_path, "attr.pkl"), "wb") as f:
+
+        if save_anndata:
+            self.adata.write(os.path.join(dir_path, "adata.h5ad"))
```
adamgayoso, line 185: might want to add compression, or anndata_write kwargs or something

scvi/core/models/base.py
```diff
@@ -179,23 +180,41 @@ def save(self, dir_path: str, overwrite: bool = False):
...
+
+        model_save_path = os.path.join(dir_path, "model_params.pt")
+        attr_save_path = os.path.join(dir_path, "attr.pkl")
+        varnames_save_path = os.path.join(dir_path, "var_names.pkl")
+
+        var_names = self.adata.var_names
+        with open(varnames_save_path, "wb") as fp:
+            pickle.dump(var_names, fp)
```
adamgayoso, line 193: should we use h5 instead of pickle for this and the user attrs? I think you can put it in a dataframe and save hdf5 with pandas API. it might be cleaner

adamgayoso: I'd keep the model save file and the anndata files separate either way.

scvi/core/models/base.py
```diff
@@ -210,14 +229,42 @@ def load(cls, adata: AnnData, dir_path: str, use_cuda: bool = False):
...
+
         if "init_params_" not in attr_dict.keys():
             raise ValueError(
-                "No init_params_ were saved by the model. Check out the developers guide if creating custom models."
+                "No init_params_ were saved by the model. Check out the "
+                "developers guide if creating custom models."
             )
         # get the parameters for the class init signiture
         init_params = attr_dict.pop("init_params_")
+        use_cuda = use_cuda and torch.cuda.is_available()
```
adamgayoso, line 266: I'd add a comment here why we need this so we don't forget

scvi/core/models/base.py
```diff
@@ -179,23 +180,41 @@ def save(self, dir_path: str, overwrite: bool = False):
...
-        with open(os.path.join(dir_path, "attr.pkl"), "wb") as f:
+
+        if save_anndata:
+            self.adata.write(os.path.join(dir_path, "adata.h5ad"))
+
+        model_save_path = os.path.join(dir_path, "model_params.pt")
+        attr_save_path = os.path.join(dir_path, "attr.pkl")
+        varnames_save_path = os.path.join(dir_path, "var_names.pkl")
+
+        var_names = self.adata.var_names
```
adamgayoso, line 191: I think you should use the get_var_names_from_setup_anndata function because if using raw the names might be different. This will require some changes in load.

adamgayoso: Alternatively, should we just remove all use raw functionality? It seems like it's causing headaches for little value.

adamgayoso: most people use raw for the library size log normalized data (like in our tutorials), not the raw counts, and this is kind of the convention in scanpy.

adamgayoso (changes requested):

scvi/core/models/base.py
galenxing: I just played around with saving it as an h5. You need to cast the dtype of Index to save and theres no easy way to load and convert back to the user's dtype. It gets weird if you have like pd.Index(1, 2, '3') which pickle has no problem with.

galenxing: We could also just always cast to str and when checking, cast the 2nd adata.var_names to str as well and just compare like that. Edge case this will fail in is: pd.Index(0, '0')

scvi/core/models/base.py
```diff
@@ -210,14 +239,45 @@ def load(cls, adata: AnnData, dir_path: str, use_cuda: bool = False):
...
+            adata = read(adata_path)
+        elif not os.path.exists(adata_path) and adata is None:
+            raise ValueError(
+                "Save path contains no saved anndata and no adata was passed."
+            )
+
+        with open(varnames_path, "rb") as handle:
+            var_names = pickle.load(handle)
+
+        if len(set(adata.var_names) - set(var_names)) != 0:
```
adamgayoso, line 255: there's a pd.Index method to check if two Index objects are equal

scvi/data/_utils.py
```diff
@@ -32,7 +32,7 @@ def _compute_library_size_batch(
     local_l_mean_key: str = None,
     local_l_var_key: str = None,
     layer=None,
-    use_raw=False,
+    # use_raw=False,
```
adamgayoso, line 35: why a comment?

scvi/data/_utils.py
```diff
@@ -148,9 +144,10 @@ def _check_anndata_setup_equivalence(adata_source, adata_target):
     adata = adata_target
 
     stats = _scvi_dict["summary_stats"]
-    use_raw = _scvi_dict["use_raw"]
+    # use_raw = _scvi_dict["use_raw"]
```
adamgayoso, line 147: comment

scvi/data/_utils.py
```diff
@@ -148,9 +144,10 @@ def _check_anndata_setup_equivalence(adata_source, adata_target):
...
 
-    target_n_vars = adata.shape[1] if not use_raw else adata.raw.shape[1]
+    # target_n_vars = adata.shape[1] if not use_raw else adata.raw.shape[1]
```
adamgayoso, line 149: comment

scvi/model/_utils.py
```diff
@@ -122,8 +122,8 @@ def _get_var_names_from_setup_anndata(adata):
     """Gets var names by checking if using raw."""
     var_names = (
         adata.var_names
-        if adata.uns["_scvi"]["use_raw"] is False
-        else adata.raw.var_names
+        # if adata.uns["_scvi"]["use_raw"] is False
```
adamgayoso, line 125: we can keep this util, but let's remove these comments

adamgayoso (changes requested):

scvi/core/models/base.py
adamgayoso: I think it's save to assume the var names are strs. regardless, you can maybe do instead `var_names.to_numpy()` and save using numpy's save method. Then you can make a new index on load from the numpy array and to check if they are equal, the index equals method.

galenxing: Okay, I'll just cast to str then and save as h5, since np.save just saves as pickle anyways.

scvi/data/_utils.py
galenxing: woops

scvi/core/models/base.py
```diff
@@ -179,23 +189,43 @@ def save(self, dir_path: str, overwrite: bool = False):
...
+                os.path.join(dir_path, "adata.h5ad"), **anndata_write_kwargs
+            )
+
+        model_save_path = os.path.join(dir_path, "model_params.pt")
+        attr_save_path = os.path.join(dir_path, "attr.pkl")
+        varnames_save_path = os.path.join(dir_path, "var_names.h5")
+
+        var_names = self.adata.var_names
+        with h5py.File(varnames_save_path, "w") as f:
+            f.create_dataset("var_names", data=var_names.to_numpy().astype("S"))
```
adamgayoso, line 204: why not use https://numpy.org/doc/stable/reference/generated/numpy.save.html? then you don't have to worry about the type

adamgayoso (changes requested):

adamgayoso (approved):

PR #829, opened 2020-10-13
Update Codacy coverage reporter script

lolgab:
This changes codacy-coverage-reporter to the latest version
using the native binary for linux. It also adds the codacy cache
directory to travis cache so Travis can reuse the binary instead
of re-downloading every time.


lolgab:
Happy to help 🙂

PR #834, opened 2020-10-14
Staging for official release

### LINKED ISSUE #835: Support for either CSC or CSR?
Sparse matrices are currently cast to CSR matrices when setting up the anndata.

Would it be possible to optionally support CSC format?

I am often doing downstream analysis of the AnnData files with Julia, which for some reason only supports CSC matrices. Currently I need to convert adata.X to CSC and save a copy of adata for analysis with Julia. It would make my data organization a bit easier if scVI could also work on CSC matrices.

adamgayoso:
Fixes #835


galenxing (approved):
Looks good!

PR #842, opened 2020-10-27
[WIP] fix semi-supervised trained

### LINKED ISSUE #841: scANVI does not compute the conditioned ELBO
The current implementation of scANVI only computes the ELBO without relying on the cell states' knowledge for labeled examples (see [here](https://github.com/YosefLab/scvi-tools/blob/master/scvi/core/trainers/annotation.py#L200)). 
We should fix this behavior and treat labeled and unlabelled examples separately.

There are two options to consider to code this. The first one would use only one data loader, and compute the proper ELBO for each data-point. In the other option, in phase with the current codebase, we would compute the marginalized ELBO for a minibatch of unlabelled cells, and the conditioned ELBO for labelled cells. The overall ELBO would be weighted out based on the labelled fraction of the training data. I would opt for the second option.

However, I don't know which default behavior and training options we should provide. Should the current behavior still be an option?

#### Versions:

> VERSION

PierreBoyeau:
Fixes #841


adamgayoso:
This looks reasonable to me, don't worry about codacy... @romain-lopez is this what you want?

romain-lopez:
Thanks Pierre, that looks good to me!

romain-lopez:
where are we standing on this? is this already done in the coming pytorch-lightning version?

adamgayoso:
What should we do about this @romain-lopez @PierreBoyeau ?

romain-lopez:
What code is being used for the annotation project with @galenxing ?

galenxing:
@romain-lopez We're using what is on master right now. Based on our benchmarking results, theres not a big difference between current master and the non-pytorch-lightning version. 

If we want this conditioned elbo, we can rebase this branch off master and include conditioned elbo as an option (instead of default).

romain-lopez:
That sounds like a great option!

adamgayoso:
Closing this as it uses the old API

PR #848, opened 2020-11-02
DE Interface Fixes

galenxing:
fixes if group1 is an int
fixes if groupby isn't categorical


adamgayoso:
@galenxing can you just add some tests?

PR #852, opened 2020-11-11
Model refactor + pytorch lightning + multiple covariates

### LINKED ISSUE #364: Continuous covariate for batch in scVI
In the original publication, s is supposed to be any type of random variable even though it is introduced with only batch effects. Unfortunately, the codebase only works for discrete values of s. There would be actually quite some changes involved as the batch id is transformed far inside the module into a one-hot 
https://github.com/YosefLab/scVI/blob/master/scvi/models/modules.py#L57

Being able to condition on a continuous, multi-dimensional measurement (which includes the discrete case if adding the onehot transformation) would be really useful for people in the YosefLab and the greater comp bio community.

### LINKED ISSUE #390: Misleading function name in Posterior.py
The `update` function of the `Posterior` is not updating the Posterior object but is copying the Posterior and changes its data loader - which is the appropriate behavior.
However the name `update` suggests the modification is performed inplace.
https://github.com/YosefLab/scVI/blob/66cc63ead0d1a219f8d0ba1018d6f610f2b91679/scvi/inference/posterior.py#L117-L122

**The name of this function should be changed.** 

_Ideas_: 
- `copy_update`
- `with_new_data_loader` - such that the function is called with`my_posterior.with_new_data_loader(...)`

### LINKED ISSUE #393: Use pytorch early stopping and reduce LR implementation
It'd be better to use Pytorch's `ReduceLROnPlateau` function rather than our own implementation of it. `ReduceLROnPlateau` lowers the learning rate multiple times rather than just once, so it probably improves on what we do now. 

As part of this issue, let's delete all this code:
https://github.com/YosefLab/scVI/blob/1f399e75068bd80aeb785a1810b30cdd34bfd14b/scvi/inference/trainer.py#L259-L352

### LINKED ISSUE #693: pytorch-lightning
@adamgayoso @romain-lopez @PierreBoyeau --

I've had a pretty good experience on another project with pytorch-lightning, so I thought I'd mention it here. It might work even better for scVI. See
https://pytorch-lightning.readthedocs.io/en/stable/new-project.html

I like how lightning puts each dataloader in the same class as the model that it trains. It's also nice that boilerplate code (e.g. the training loop) can be removed entirely. Might help with #654

### LINKED ISSUE #851: stereoscope implementation
Hi @romain-lopez ,

following almaan/stereoscope#18, I finally got time to check out your implementation of stereoscope in `romain/spatial` branch. Don't have much to comment, it's great!
I really like the decoupling between scmodel and spatial model, making it super convenient to fit one model for e.g. an atlas and reuse it multiple times for visium slides. 
In terms of default `n_epochs`, they made a lot of sense also for the data I tried it on. Maybe I would be more relaxed with the spatial model (I increased it to 7k) but of course arg very dependent to data/settings.

Do you still plan to merge it to master? I'm sure many would use it.
Thank you again! 🎉

adamgayoso:
- Fixes #693, Fixes #364, Fixes #393, Fixes #390, Fixes #851


scvi/lightning/__init__.py
```diff
@@ -0,0 +1,4 @@
+from ._trainer import Trainer
+from ._vaetask import AdvesarialTask, SemiSupervisedTask, VAETask
```
romain-lopez, line 2: Just pointing out this typo!

scvi/lightning/_vaetask.py
```diff
@@ -0,0 +1,487 @@
...
+
+
+class VAETask(pl.LightningModule):
+    """
+    Lightning module task to train scvi-tools modules
+
+    Parameters
+    ----------
+    vae_model
+        A model instance from class ``AbstractVAE``.
```
romain-lopez, line 22: For terminology, that's not really a model. Let's be careful to keep this in mind

romain-lopez: It's confusing because of what we call model (the high-level API)

PR #859, opened 2020-11-16
More arches params

### LINKED ISSUE #849: Dumb Question - project new data onto existing data with scVI
If I've gone through the trouble of making a scVI --> UMAP representation I'd *like* if I could "keep" the exisiting UMAP projection if I add a (relatively) small number of cells. This is possible with UMAP (https://cran.r-project.org/web/packages/umap/vignettes/umap.html). But this would require the new data to have equivalent scVI latent dimensions.

Is there a path towards re-using an existing scVI model to incorporate new data?

adamgayoso:
- Fixes #849


adamgayoso:
@M0hammadL, @Koncopd, @cottoneyejoe95, please let me know if this works for you.

Cottoneyejoe95:
Hi Adam, when I use your new parameter "freeze_classifier=True" then I get the following the error:

optimizer got an empty parameter list

I think we already discussed that problem and I think that was the reason why you didnt freeze the classifier in this way in the past. Do maybe have another way of freezing the classifier?

adamgayoso:
@Cottoneyejoe95 I removed that parameter and will merge this to add the other functionality.

PR #879, opened 2021-01-06
scanvi updates

galenxing:

PR #883, opened 2021-01-09
Weighted random sampling by label proportions for scANVI

dibidave:
This is an alternative solution to the class imbalance problem than downsampling the labels to only a fixed number per label.

I've created an alternative sampler, the WeightedRandomSampler, that generates an iterator over samples based on user-defined weights for each sample (cell). For each batch it generates, it randomly samples cells, with replacement. Unlike batch samplers, this means that each epoch will present different cells, and some cells will be presented multiple times within an epoch, and even within batches.

I've updated ScviDataLoader to accept a new argument, sample_weights, which, if provided, will trigger the use of the WeightedRandomSampler instead of the BatchSampler. The default argument of None keeps ScviDataLoader functionality unchanged.

I've also updated the scANVI train method to have an additional optional argument, balanced_sampling, that, when set to True, generates weights for each label that are inversely proportional to the frequency of that label. These weights are then passed on to the ScviDataLoader, and therefore the WeightedRandomSampler. This makes it so that each cell label will be sampled in equal proportions. Again, the default argument of balanced_sampling being set to False keeps scANVI functionality unchanged.

From my testing, this improves accuracy on imbalanced datasets, and prevents the problem of having low probability classes from dropping out entirely.

I installed the pre-commit hook and the commit passes the black and flake8 tests, and I've also run the model test (tests/models/test_models.py), and all tests pass.

I'm new to this community, and this is my first pull request, so I'm not entirely sure if this is the right process for adding features. Any feedback is welcome, and if I'm doing something completely wrong here by even suggesting a pull request for an unrequested feature, please let me know! Also happy to provide some data to show performance on imbalanced classes.


dibidave:
Also - which branch is the best one to suggest pull requests into for new features by default - dev?

galenxing:
Hi @dibidave, thanks for this contribution! I agree, resampling cells during training is a good alternative to just downsampling and has also produced better results in our testing as well. 

We're in the middle of a major code refactor right now (#852) where we are transitioning to Pytorch lightning. So unfortunately you will have to reimplement this off the `pytorch_lit` branch. But, fortunately, it should be a lot less code.

I did recently implement a similar sampling regime in #879, but w/o per cell weights and w/o replacement. I think including the option of per cell weights and oversampling low probability classes would be a great addition. 

To quickly orient you, we [added a callback to the scanvi trainer](https://github.com/YosefLab/scvi-tools/blob/9726e520bbefde667d3f2cb6e1a89b3aad9b70ca/scvi/model/_scanvi.py#L345). The [callback](https://github.com/YosefLab/scvi-tools/blob/9726e520bbefde667d3f2cb6e1a89b3aad9b70ca/scvi/lightning/_sampling.py#L4-L10) is run before each epoch and calls the [resample_labels]( https://github.com/YosefLab/scvi-tools/blob/9726e520bbefde667d3f2cb6e1a89b3aad9b70ca/scvi/dataloaders/_concat_dataloader.py#L78-L99
) method which does the resampling and then sets the indices used by the dataloader for the labelled data.

Thus, we could probably add `sample_weights` as a class variable to `SemiSupervisedDataLoader`, which if not None, `subsample_labels` would sample according to your regime. 

The scanVI code is in a bit of flux right now but will hopefully be merged in by the end of the week, so once that's merged in, it would be a good time for you to implement the changes.

Let me know what you think and thanks again for contributing!

dibidave:
This sounds good, thanks for the info! Happy to re-implement this in `subsample_labels` post-refactor. If I start working on it now, would it be best to work off of the `scanvi_training` branch - is that closest to the refactor?

galenxing:
@dibidave sweet, and yea you can branch off `scanvi_training`

adamgayoso:
@dibidave our `master` branch is now mostly stable, so feel free to make a new PR off of `master`. This is an interesting approach and we can likely add it as an option, but we would probably need some sort of benchmarking to show that it does something useful, at least compared to the approach we have now that @galenxing implemented.

adamgayoso:
@dibidave I'm going to close this, but feel free to reach out if you'd like to discuss this further.

PR #890, opened 2021-01-22
[WIP] Cell assign

wukathy:


scvi/external/cellassign/_model.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Examples
+    --------
+    >>> # TODO
+    """
+
+    def __init__(
+        self,
+        sc_adata: AnnData,
+        cell_type_markers: pd.DataFrame,
```
adamgayoso, line 35: this will be a binary matrix, genes by cell types (rho)

`sc_adata` --> fancy pandas dataframe, gene names are in `sc_adata.var.index`

check that genes are the same in both objects, reorder `cell_type_markers` according to order of genes in `sc_adata.var.index`

scvi/external/cellassign/_model.py
```diff
@@ -0,0 +1,70 @@
...
+        self._model_summary_string = (
+            "CellAssign Model with params: \nn_genes: {}, n_labels: {}"
+        ).format(
+            self.n_genes,
+            self.n_labels,
+        )
+        self.init_params_ = self._get_init_params(locals())
+
+    def predict(self) -> np.ndarray:
+        """Predict soft cell type assignment probability for each cell."""
```
adamgayoso, line 61: take a look at `vaemixin.py` `get_latent_representation` function.

scvi/external/cellassign/_model.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+        self._model_summary_string = (
+            "CellAssign Model with params: \nn_genes: {}, n_labels: {}"
+        ).format(
+            self.n_genes,
+            self.n_labels,
+        )
+        self.init_params_ = self._get_init_params(locals())
+
+    def predict(self) -> np.ndarray:
```
adamgayoso, line 60:
```suggestion
    def predict(self, adata: AnnData) -> np.ndarray:
```

scvi/external/cellassign/_model.py
adamgayoso: return a dataframe, index is `adata.obs.index`, columns are the rows (cell types) of `cell_type_markers` df

adamgayoso: values are probability assignment to cell type (gamma)

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+            torch.exp(theta_log) + self.THETA_LOWER_BOUND
+        )
+        delta_log_prior = Normal(self.delta_log_mean, self.delta_log_variance)
+        summed_delta_log = torch.sum(self.delta_log)
+        delta_log_prob = -torch.sum(delta_log_prior.log_prob(summed_delta_log))
+        prior_log_prob = theta_log_prob
+        if self.shrinkage:
+            prior_log_prob += delta_log_prob
+
+        return SCVILoss(loss, Q, prior_log_prob, 0.0)
```
adamgayoso, line 179:
```suggestion
        return LossRecorder(loss, Q_per_cell, torch.zeros_like(Q_per_cell), prior_log_prob)
```
Q_per_cell should be the per cell (no outside sum over n) of this value.

And therefore, loss would be `torch.mean(Q_per_cell) * n_obs + prior_log_prob`
Screen Shot 2021-01-23 at 10 10 30 AM

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        # compute Q
+        # gamma_fixed = torch.empty((None, self.n_labels))
+        # Q = -torch.einsum("nc,cn->", gamma_fixed, p_y_unorm)
+        Q = p_y_norm
+
+        # second term in SCVILoss is Q per cell without prior terms. shape is (n_cells,)
+        loss = torch.sum(p_y_norm)
+
+        # third term is log prob of prior terms in Q
+        log_softmax = torch.nn.LogSoftmax()
```
adamgayoso, line 166: torch nn functional logsoftmax

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        generative_outputs,
+        n_obs: int = 1.0,
+    ):
+        # generative_outputs is a dict of the return value from `generative(...)`
+        # assume that `n_obs` is the number of training data points
+        p_y_norm = generative_outputs["p_y_norm"]
+
+        # compute Q
+        # gamma_fixed = torch.empty((None, self.n_labels))
+        # Q = -torch.einsum("nc,cn->", gamma_fixed, p_y_unorm)
```
adamgayoso, line 159: I'm assuming this or similar is correct for the components of Q that do not correspond to priors.

Recall we want to take the mean (or if einsum works, divide by number of cells in minibatch, and multiply by n_obs)

and then add prior terms.

adamgayoso (changes requested):

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        n_obs: int = 1.0,
+    ):
+        # generative_outputs is a dict of the return value from `generative(...)`
+        # assume that `n_obs` is the number of training data points
+        p_y_unorm = generative_outputs["p_y_unorm"]
+        gamma = generative_outputs["gamma"]
+
+        # compute Q
+        # take mean of number of cells and multiply by n_obs (instead of summing n)
+        q_per_cell = torch.mean(torch.sum(gamma * p_y_unorm, 1)) * n_obs
```
adamgayoso, line 160:
```suggestion
        q_per_cell = torch.sum(gamma * p_y_unorm, 1)
```

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+
+        delta_rho = delta * self.rho
+        delta_rho_e = delta_rho.expand(
+            s.shape[0], self.n_genes, self.n_labels
+        )  # (n, g, c)
+        log_mu_ngc = base_mean_e + delta_rho_e
+        mu_ngc = torch.exp(log_mu_ngc)  # (n, g, c)
+
+        # compute basis means for phi - shape (B)
+        basis_means_fixed = np.linspace(torch.min(x), torch.max(x), self.B)
```
adamgayoso, line 107: let's use torch linspace and use the device argument of linspace to be the device of `x`, this allows us to write code that can be used on gpu or cpu

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        basis_means = torch.tensor(basis_means_fixed)  # (B)
```
adamgayoso, line 108: this line would then not be necessary

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        super().__init__()
+        self.n_genes = n_genes
+        self.n_labels = n_labels
+        self.rho = rho
+
+        self.register_buffer("cell_type_markers", rho)
+
+        # perform all other initialization
+        self.LOWER_BOUND = 1e-10
+        self.THETA_LOWER_BOUND = 1e-20
```
adamgayoso, line 45: Not sure these need to be class attributes? They could just be near the imports?

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        self.B = 10
```
adamgayoso, line 46: same here

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+    @auto_move_data
+    def inference(self):
+        return {}
+
+    @auto_move_data
+    def generative(self, x, y):
+        # x has shape (n, g)
+        # n = 128
+        # g = 100
+        # c = 5
```
adamgayoso, line 87: this is the case in our test data, but we don't need these comments

scvi/external/cellassign/_module.py
adamgayoso: basically in case I miss it, anywhere you create a torch Tensor in this particular function, you need to specify the device using the device of the inputs

adamgayoso: and regarding `basis_means`, below we should also use torch functions everywhere

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        basis_means_e = basis_means.expand(
+            s.shape[0], self.n_genes, self.n_labels, self.B
+        )  # (n, g, c, B)
+        phi = (  # (n, g, c)
+            torch.sum(a_e * torch.exp(-b_e * torch.square(mu_ngcb - basis_means_e)), 3)
+            + self.LOWER_BOUND
+        )
+
+        # compute gamma
+        nb_pdf = NegativeBinomial(probs=mu_ngc, total_count=phi)
```
adamgayoso, line 129:
```suggestion
        nb_pdf = NegativeBinomial(mu=mu_ngc, theta=phi)
```

adamgayoso: our custom NB dist supports mean and inverse dispersion

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        y_ = x.unsqueeze(-1).expand(s.shape[0], self.n_genes, self.n_labels)
+        y_log_prob_raw = nb_pdf.log_prob(y_)  # (n, g, c)
+        theta_log_e = theta_log.expand(s.shape[0], self.n_labels)
+        p_y_unorm = torch.sum(y_log_prob_raw, 1) + theta_log_e  # (n, c)
```
adamgayoso, line 133: this is really p(y, c), so let's rename the var to `p_y_c`

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        p_y_norm = torch.logsumexp(p_y_unorm, 1)
```
adamgayoso, line 134: rename var to `normalizer_over_c`

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        p_y_norm_e = p_y_norm.unsqueeze(-1).expand(s.shape[0], self.n_labels)
+        gamma = torch.exp(p_y_unorm - p_y_norm_e)  # (n, c)
+
+        return dict(
+            mu=mu_ngc,
+            phi=phi,
+            gamma=gamma,
+            p_y_unorm=p_y_unorm,
```
adamgayoso, line 142: again `p_y_c`

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        p_y_unorm = generative_outputs["p_y_unorm"]
+        gamma = generative_outputs["gamma"]
+
+        # compute Q
+        # take mean of number of cells and multiply by n_obs (instead of summing n)
+        q_per_cell = torch.sum(gamma * p_y_unorm, 1)
+
+        # third term is log prob of prior terms in Q
+        theta_log = F.log_softmax(self.theta_logit)
+        theta_log_prior = Dirichlet(torch.tensor(self.dirichlet_concentration))
```
adamgayoso, line 164: same here with torch.tensor, when you define dirichelt_concentration in the init, use the `register_buffer` like for phi, this allows a call of `model.cuda()` to put everything on the gpu.

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,188 @@
...
+        theta_log_prob = -theta_log_prior.log_prob(
+            torch.exp(theta_log) + self.THETA_LOWER_BOUND
+        )
+        delta_log_prior = Normal(self.delta_log_mean, self.delta_log_variance)
+        summed_delta_log = torch.sum(self.delta_log)
+        delta_log_prob = -torch.sum(delta_log_prior.log_prob(summed_delta_log))
+        prior_log_prob = theta_log_prob
+        if self.shrinkage:
```
adamgayoso, line 172: is there a logical issue if `self.shrinkage` is False, then `delta_log_prior = Normal(self.delta_log_mean, self.delta_log_variance)` would give a runtimeerror

adamgayoso (changes requested):

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,225 @@
...
+
+        # compute theta
+        self.theta_logit = torch.nn.Parameter(torch.randn(self.n_labels))
+
+        # compute delta (cell type specific overexpression parameter)
+        self.delta_log = torch.nn.Parameter(
+            torch.FloatTensor(self.n_genes, self.n_labels).uniform_(
+                np.log(self.min_delta), 2
+            )
+        )
```
adamgayoso, line 75: This should be initialized from -2 to 2

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,225 @@
...
+        return input_dict
+
+    @auto_move_data
+    def inference(self):
+        return {}
+
+    @auto_move_data
+    def generative(self, x, design_matrix=None):
+        # x has shape (n, g)
+        delta = torch.exp(self.delta_log)  # (g, c)
```
adamgayoso, line 117: before this line, you should create a clamped delta, https://pytorch.org/docs/stable/generated/torch.clamp.html

clamp delta_log with a minimum value of log (min_delta)

adamgayoso (changes requested):

scvi/external/cellassign/_model.py
```diff
@@ -0,0 +1,98 @@
...
+            **model_kwargs,
+        )
+        self._model_summary_string = (
+            "CellAssign Model with params: \nn_genes: {}, n_labels: {}"
+        ).format(
+            self.n_genes,
+            rho.shape[1],
+        )
+        self.init_params_ = self._get_init_params(locals())
+
```
adamgayoso, line 74:
```suggestion
@torch.no_grad()
```

scvi/external/cellassign/_model.py
```diff
@@ -0,0 +1,98 @@
...
+    def predict(self, adata: AnnData) -> np.ndarray:
+        """Predict soft cell type assignment probability for each cell."""
+        adata = self._validate_anndata(adata)
+        scdl = self._make_scvi_dl(adata=adata)
+        latent = []
+        for tensors in scdl:
+            generative_inputs = self.module._get_generative_input(tensors, None)
+            # x, y = self.model.inference(**inference_inputs)
+            outputs = self.module.generative(self, generative_inputs)
```
adamgayoso, line 83:
```suggestion
            outputs = self.module.generative(**generative_inputs)
```

adamgayoso: double check

scvi/external/cellassign/_model.py
```diff
@@ -0,0 +1,98 @@
...
+        adata = self._validate_anndata(adata)
+        scdl = self._make_scvi_dl(adata=adata)
+        latent = []
+        for tensors in scdl:
+            generative_inputs = self.module._get_generative_input(tensors, None)
+            # x, y = self.model.inference(**inference_inputs)
+            outputs = self.module.generative(generative_inputs)
+            gamma = outputs["gamma"]
+            latent += [gamma.cpu()]
+        return np.array(torch.cat(latent))
```
adamgayoso, line 86: turn this into a pandas dataframe of cells (with index equal to adata.obs.index, and columns equal to the cell types of the inputted marker matrix)

scvi/external/cellassign/_model.py
```diff
@@ -0,0 +1,98 @@
...
+        """Predict soft cell type assignment probability for each cell."""
+        adata = self._validate_anndata(adata)
+        scdl = self._make_scvi_dl(adata=adata)
+        latent = []
+        for tensors in scdl:
+            generative_inputs = self.module._get_generative_input(tensors, None)
+            # x, y = self.model.inference(**inference_inputs)
+            outputs = self.module.generative(generative_inputs)
+            gamma = outputs["gamma"]
+            latent += [gamma.cpu()]
```
adamgayoso, line 86:
```suggestion
            predictions += [gamma.cpu()]
```

scvi/external/cellassign/_model.py
adamgayoso, line 87:
```suggestion
        return np.array(torch.cat(predictions))
```

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,226 @@
...
+        # third term is log prob of prior terms in Q
+        theta_log = F.log_softmax(self.theta_logit)
+        theta_log_prior = Dirichlet(torch.tensor(self.dirichlet_concentration))
+        theta_log_prob = -theta_log_prior.log_prob(
+            torch.exp(theta_log) + THETA_LOWER_BOUND
+        )
+        prior_log_prob = theta_log_prob
+        if self.shrinkage:
+            delta_log_prior = Normal(self.delta_log_mean, self.delta_log_variance)
+            summed_delta_log = torch.sum(self.delta_log)
```
adamgayoso, line 209: remove

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,226 @@
...
+            delta_log_prob = -torch.sum(delta_log_prior.log_prob(summed_delta_log))
```
adamgayoso, line 210: mask the log_prob output by the rho matrix, and then sum

adamgayoso (changes requested):

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,226 @@
...
+        n_obs: int = 1.0,
+    ):
+        # generative_outputs is a dict of the return value from `generative(...)`
+        # assume that `n_obs` is the number of training data points
+        p_y_c = generative_outputs["p_y_c"]
+        gamma = generative_outputs["gamma"]
+
+        # compute Q
+        # take mean of number of cells and multiply by n_obs (instead of summing n)
+        q_per_cell = torch.sum(gamma * p_y_c, 1)
```
adamgayoso, line 198:
```suggestion
        q_per_cell = torch.sum(gamma * -p_y_c, 1)
```

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,229 @@
...
+        )
+        prior_log_prob = theta_log_prob
+        if self.shrinkage:
+            delta_log_prior = Normal(self.delta_log_mean, self.delta_log_variance)
+            delta_log_prob = torch.masked_select(
+                delta_log_prior.log_prob(self.delta_log), (self.rho > 0)
+            )
+            prior_log_prob += -torch.sum(delta_log_prob)
+
+        loss = torch.mean(q_per_cell) * n_obs + prior_log_prob
```
adamgayoso, line 216:
```suggestion
        loss = (torch.mean(q_per_cell) * n_obs + prior_log_prob) / n_obs
```

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,238 @@
...
+        # x has shape (n, g)
+        delta = torch.exp(self.delta_log)  # (g, c)
+        theta_log = F.log_softmax(self.theta_logit)  # (c)
+
+        # compute mean of NegBin - shape (n_cells, n_genes, n_labels)
+        s = size_factor
+        base_mean = torch.log(s)  # (n, 1)
+        base_mean_u = base_mean.unsqueeze(-1)  # (n, 1, 1)
+        base_mean_e = base_mean_u.expand(
+            s.shape[0], self.n_genes, self.n_labels
```
adamgayoso, line 137: Let's go for a round of cleanup:

`s.shape[0]` gets used a lot, you can instead define a variable near the top of the function, you'll have to check but I think doing like `n_cells = x.shape[0]` is the same

scvi/external/cellassign/_model.py
```diff
@@ -0,0 +1,102 @@
...
+        self._model_summary_string = (
+            "CellAssign Model with params: \nn_genes: {}, n_labels: {}"
+        ).format(
+            self.n_genes,
+            rho.shape[1],
+        )
+        self.init_params_ = self._get_init_params(locals())
+
+    @torch.no_grad()
+    def predict(self, adata: AnnData) -> np.ndarray:
```
adamgayoso, line 78:
```suggestion
    def predict(self) -> np.ndarray:
```

For now, predict will just use the adata attr of the class, we'll have to change this later. You can keep the validate anndata line below.

scvi/external/cellassign/_model.py
```diff
@@ -0,0 +1,102 @@
...
+    cell_type_markers
+        Binary marker gene matrix
+    use_gpu
+        Use the GPU or not.
+    **model_kwargs
+        Keyword args for :class:`~scvi.external.CellAssignModels`
+
+    Examples
+    --------
+    >>> # TODO
```
adamgayoso, line 35: add an example like in the notebook we were looking at today.

1. load data, marker matrix
2. initialize model
3. train
4. predict

see our other docstrings for examples

scvi/external/cellassign/_model.py
```diff
@@ -0,0 +1,102 @@
...
+        self,
+        adata: AnnData,
+        cell_type_markers: pd.DataFrame,
+        size_factor_key: str,
+        use_gpu: bool = True,
+        **model_kwargs,
+    ):
+        super().__init__(adata, use_gpu=use_gpu)
+        # check that genes are the same in cell_type_markers are present in the anndata
+        # anndata may have more
```
adamgayoso, line 48: I removed this code, but let's add something a bit different.

1. Check that adata and cell type markers contain the same genes
2. If they aren't in the same order, go ahead and sort the marker matrix inplace

after checking both contain same names,
`cell_type_markers = cell_type_markers.loc[adata.var_names]`

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,238 @@
...
+
+        input_dict = dict(x=x, size_factor=size_factor, design_matrix=design_matrix)
+        return input_dict
+
+    @auto_move_data
+    def inference(self):
+        return {}
+
+    @auto_move_data
+    def generative(self, x, size_factor, design_matrix=None):
```
adamgayoso, line 124: right now we aren't using the design matrix -- this would be equivalent to `X` in their code. Let's add the necessary beta coefficients for the design matrix **only in the case when it's not None**

adamgayoso (changes requested):

scvi/external/cellassign/_module.py
```diff
@@ -0,0 +1,255 @@
...
+        )
+        prior_log_prob = theta_log_prob
+        if self.shrinkage:
+            delta_log_prior = Normal(self.delta_log_mean, self.delta_log_variance)
+            delta_log_prob = torch.masked_select(
+                delta_log_prior.log_prob(self.delta_log), (self.rho > 0)
+            )
+            prior_log_prob += -torch.sum(delta_log_prob)
+
+        loss = torch.mean(q_per_cell) * n_obs + prior_log_prob / n_obs
```
adamgayoso, line 242:
```suggestion
        loss = (torch.mean(q_per_cell) * n_obs + prior_log_prob) / n_obs
```

PR #895, opened 2021-01-29
Pyro integration

### LINKED ISSUE #862: interoperability with pyro to simplify implementing complex hierarchical priors?
Hi scVI team!

I am wondering if you are planning any interoperability with pyro to simplify implementing complex hierarchical priors? Maybe you have other ideas of simplifying that? I would like to try to adding Variational Autoencoding posteriors to cell2location (https://github.com/BayraktarLab/cell2location, https://www.biorxiv.org/content/10.1101/2020.11.15.378125v1.full) - would be cool if this is possible by integrating with scVI.

Thanks

Vitalii

adamgayoso:
Initial implementation of a Pyro base module class, and a Pyro training plan (pytorch lightning class). See the example in the tests file I added.

- Closes #862


adamgayoso:
@vitkl, if you have a chance, it would be great if you could let me know if this code is general enough for your use case. I'm not super familiar with Pyro, but by having everything in a torch nn module (see test case), our save/load, etc of the high-level API will work automatically.

The thing to note about Pytorch lightning is that you need to write device agnostic code. The trainer handles moving things to the GPU for you. This code here works on my GPU.

vitkl:
Thanks @adamgayoso ! We are currently quite busy with paper revision. We (I and @yozhikoff) will try using cell2location model with you code next week (hopefully) or week thereafter. Given your comments - getting this to work will probably need some tweaking. 

Maybe I am missing something but where is `_CONSTANTS.X_KEY` specified? E.g. how do I tell scVI what are the arguments to the model? https://github.com/YosefLab/scvi-tools/pull/895/commits/a25ee2856422c0d7cc2408939fde73c7787a9540#diff-663073fe4ef3cfa8f3fad68a502f760ee33dbb94c3c803c50635f0d63f67ee30R28-R29

As far as I understand pyro plates, they need to be given index from the data loader and other args (`subsample_size=None, subsample=None, dim=None`, http://docs.pyro.ai/en/0.3.0-release/primitives.html#pyro.plate) to work as expected (normalised by minibatch size). Without providing index from the data loader as `subsample` it will sample random indices - in the model below it is important to get correct indices for local variable `cell_norm`.
https://github.com/YosefLab/scvi-tools/pull/895/commits/a25ee2856422c0d7cc2408939fde73c7787a9540#diff-663073fe4ef3cfa8f3fad68a502f760ee33dbb94c3c803c50635f0d63f67ee30R36
We currently do this by providing indices as argument to the model / guide functions:

```python
    def create_plates(self, x_data, idx, cell2sample, cell2covar):
        return [pyro.plate("obs_axis", self.n_obs, dim=-2, 
                           subsample_size=self.minibatch_size, 
                           subsample=idx),
                pyro.plate("var_axis", self.n_var, dim=-1),
                pyro.plate("factor_axis", self.n_fact, dim=-2),
                pyro.plate("experim_axis", self.n_experim, dim=-2)]
                
    def model(self, x_data, idx, cell2sample, cell2covar):
        
        obs_axis, var_axis, factor_axis, experim_axis = self.create_plates(x_data, idx, cell2sample, cell2covar)

        with var_axis, factor_axis:
           gene_loadings_fg = pyro.sample('gene_loadings_fg',
                                                      dist.Gamma(torch.ones([self.n_fact, self.n_var]),
                                                                           torch.ones([self.n_fact, self.n_var])))
        with var_axis:
           gene_alpha_g = pyro.sample('gene_alpha_g', dist.Exponential(torch.ones([1, self.n_var])))

        with experim_axis:
           detection_mean = pyro.sample('detection_mean', dist.Gamma(torch.ones([self.n_experim, 1]), torch.ones([self.n_experim, 1])))
        with obs_axis as ind:
           cell_norm = pyro.sample('cell_norm', dist.Gamma(200, 200 / torch.mm(cell2sample[ind], self.detection_mean)))
                                                                           
        with var_axis:
            with obs_axis as ind:
                self.mu_biol = torch.mm(cell2covar, self.gene_loadings_fg)
                self.total_count, self.logits = _convert_mean_disp_to_counts_logits(self.mu_biol, self.gene_alpha_g,
                                                                            eps=1e-8) # from scVI 
                self.data_target = pyro.sample('data_target',
                                               dist.NegativeBinomial(total_count=self.total_count, logits=self.logits),
                                               obs=x_data) # I did not manage to make pyro work with scVI NegativeBinomial distribution class
```

Do failed checks mean that this code does not work?

adamgayoso:
> Do failed checks mean that this code does not work?

Yes, but the issue is just that I need to tell it not to use the GPU, as I tested it locally with a GPU. I'll fix that.

> Maybe I am missing something but where is _CONSTANTS.X_KEY specified? 

This is definitely opaque. What happens is that [setup_anndata()](https://www.scvi-tools.org/en/stable/api/reference/scvi.data.setup_anndata.html#scvi.data.setup_anndata) adds a data registry to the anndata itself, that is then used by our data loader. So our [AnnDataLoader](https://www.scvi-tools.org/en/latest/api/reference/scvi.dataloaders.AnnDataLoader.html#scvi.dataloaders.AnnDataLoader) returns a dictionary with keys (accessible via `scvi._CONSTANTS` and values being the minibatched tensors that were registered originally. What Romain did for Stereoscope was to add an obs column of the index of each cell and register it, so that it could be used for stereoscope. You can see this in the [init of the spatial model](https://www.scvi-tools.org/en/latest/_modules/scvi/external/stereoscope/_model.html#SpatialStereoscope). 

And going back to your original comment about mixing autoguides with NN encoders -- that's not something that's really in our control as much as it is a question of whether you can do that in Pyro easily.

pyproject.toml
```diff
@@ -45,6 +45,7 @@ pandas = ">=1.0"
 pre-commit = {version = ">=2.7.1", optional = true}
 prospector = {version = "*", optional = true}
 pydata-sphinx-theme = {version = ">=0.4.0", optional = true}
+pyro-ppl = ">=1.1.0"
```
romain-lopez, line 48: We'd probably like to keep this optional

adamgayoso: I thought about this, but they have really minimal requirements, and it might confusing for some people why they need to install pyro. In other words, the end users don't care if it's pytorch or pyro. 

https://github.com/pyro-ppl/pyro/blob/51d230ff62aab731f93bab66993a5e3532b749d0/setup.py#L85

romain-lopez: sounds good to me

romain-lopez:
I added a simple version of scVI with Pyro. I think this is enough to show that our API makes it easy to implement models with Pyro as well! Once this is merged, we can add a skeleton in pyro to make this integration more clear!

romain-lopez:
what's with the codacy?

vitkl:
This looks very exciting. Thanks for putting this together! I will try
adding cell2location tomorrow or on Monday.

Is "ind_x" in this line an index in minibatch or is it a column in
`st_adata.obs`?
https://github.com/YosefLab/scvi-tools/blob/cda7bf82edd56d6580fb2e0f525e396f6350b124/scvi/external/stereoscope/_model.py#L165

vitkl:
With AnnotationDataLoader or otherwise, is it possible to load and use the whole dataset in each training iteration rather than mini-batches?
E.g. here: https://github.com/YosefLab/scvi-tools/blob/e90dacd7485e5342badb02c791d977fd9e73609d/tests/models/test_pyro.py#L152

romain-lopez:
it should be easy to change the pytorch data loader behavior to send the full dataset at each iteration (maximal batch size, and un-randomized sampling of indices)

adamgayoso:
> it should be easy to change the pytorch data loader behavior to send the full dataset at each iteration (maximal batch size, and un-randomized sampling of indices)

the only caveat here is it would put on cuda each time -- this is a new functionality we need to add -- full batch data loading on gpu.

vitkl:
This is exactly what I am asking about - is it possible to load full data
once rather than in every batch. Will try adding cell2location tomorrow -
full data would have been faster to test (we didn't fully sort out
minibatch training in pyro yet)

adamgayoso:
I'm not really sure how these autoguides work, but it's not happy on load, as when it's initialized it has no parameters. I think it's parameters get initialized after some data passes through?

adamgayoso:
@galenxing @romain-lopez I think I figured it out -- if using AutoGuides, which from my understanding is like ADVI? You have to run some data through the model so the guides get their torch params and THEN you can load from the state dict. Otherwise the guides have no torch params (I believe pyro needs to trace through the model to infer the guide, and this requires data)

In other words, I think we need to do the following

1. If the underlying module is Pyro and there are Autoguides, run a models train method for 1 training step just before loading the state dict (might need to clear param store before too?)

There are probably a few good ways we can detect the Pyro/autoguide prereq.

vitkl:
That's correct. AutoGuides provide an easy way to do ADVI by using
appropriately transformed normal distributions for VI approximation.

Yes, the AutoGuides need to be passed the data and all other arguments as
to the model function. I think think is used to guess dimensions. Also a
good idea to clear param store.

romain-lopez:
@fritzo @jamestwebber 
With our incoming major refactor, we are exploring possibilities for Pyro support. In particular, I have implemented a simple version of scVI here, based on your VAE tutorial, as well as your scanvi reimplementation. If this is something interesting to you, it would be great to have some feedback / suggestions.

martinjankowiak:
@romain-lopez exciting! in case it helps i implemented scanvi [here](https://github.com/pyro-ppl/pyro/tree/dev/examples/scanvi). also see [here](https://colab.research.google.com/drive/1rEOh3N988KkymwvScsb7BEYHXvEBopZ4)

scvi/lightning/_trainingplans.py
```diff
@@ -522,3 +523,55 @@ def validation_step(self, batch, batch_idx, optimizer_idx=0):
...
+        A model instance from class ``PyroBaseModuleClass``.
+    lr
+        Learning rate used for optimization.
+    """
+
+    def __init__(
+        self,
+        pyro_module: PyroBaseModuleClass,
+        lr: float = 1e-3,
+        loss_fn: Callable = pyro.infer.Trace_ELBO(),
```
fritzo, line 544: Warning: it's dangerous to set stateful default arguments. In this case the same default `Trace_ELBO` instance will be shared by all models and (details below) bad things may happen. I'd recommend defaulting to `None` and doing a standard `if loss_fn is None: loss_fn = pyro.infer.Trace_ELBO()`.

<details>

Each `Trace_ELBO` instance guesses and stores the number of plates in its model, and assumes it will be associated with only a single model. If you use it with a different model with a different number of plates, you might see tensor shape errors.

</details>

tests/models/test_pyro.py
```diff
@@ -0,0 +1,196 @@
...
+from scvi.data import synthetic_iid
+from scvi.dataloaders import AnnDataLoader
+from scvi.lightning import PyroTrainingPlan, Trainer
+
+
+class BayesianRegression(PyroModule, PyroBaseModuleClass):
+    def __init__(self, in_features, out_features):
+        super().__init__()
+
+        self._auto_guide = AutoDiagonalNormal(self)
```
fritzo, line 22: Technical detail: One of the rules of `PyroModule`s is that the model and guide must be separate `PyroModule` s, and cannot be contained in a single `PyroModule`; however they can be contained in a single `nn.Module`. You might consider refactoring to separate `PyroBaseModuleClass` from some sort of `PyroModelClass` like this (where `s <: t` means `issubclass(s, t)`):
```
PyroBaseModuleClass <: nn.Module
  .model : PyroBaseModelClass <: PyroModule
  .guide : AutoNormal <: PyroModule
```
Basically "a model can't have its guide as an attribute".

<details>
The reason is due to `PyroModule` naming schemes and caching of `pyro.sample` calls. If `model` and `guide` are contained in a single `PyroModule` there may be weird conflicts in both names and the `pyro.sample` cache.
</details>

tests/models/test_pyro.py
```diff
@@ -0,0 +1,196 @@
...
+            nb_logits = (px_rate + self.epsilon).log() - (
+                self.px_r.exp() + self.epsilon
+            ).log()
+            x_dist = dist.ZeroInflatedNegativeBinomial(
+                gate_logits=px_dropout, total_count=self.px_r.exp(), logits=nb_logits
+            )
+            # score against actual counts
+            pyro.sample("obs", x_dist.to_event(1), obs=x)
+
+    @pyro_method
```
fritzo, line 110: See note above: a `PyroModule` model cannot own its guide; you'll need an outer `nn.Module` to contain both.

fritzo (commented):
Looks great, glad to see you're trying out Pyro :smile:

I have only a couple warnings about sharp edges like `PyroModule` requirements. Let me know if you have any questions or want to chat (Pyro slack, PyTorch slack, zoom, ...).

romain-lopez:
Wonderful! Thanks @martinjankowiak and @fritzo! We will be in touch if we have more questions!

tests/models/test_pyro.py
adamgayoso: this is super helpful!

adamgayoso:
also @vitkl the API changed a bit, please take a look at the bayesian regression example. The wrapper NN Module class needs to have a `model` and `guide` attr.

tests/models/test_pyro.py
```diff
@@ -36,25 +34,30 @@ def __init__(self, in_features, out_features):
...
 
-    def guide(self, x, y):
-        return self._auto_guide(x, y)
+
+class BayesianRegressionModule(PyroBaseModuleClass):
+    def __init__(self, in_features, out_features):
+
+        super().__init__()
+        self.model = BayesianRegressionPyroModel(in_features, out_features)
+        self.guide = AutoDiagonalNormal(self.model)
```
fritzo, line 50: :+1:

adamgayoso: @fritzo -- is there a way to load in the `AutoGuide` parameters on a fresh initialization? It seems like some data needs to be first passed through.

fritzo: We usually simply `torch.save(model, ...)` instead of `torch.save(model.state_dict(), ...)` since the former also saves pythonic data other than tensors. Would that work for you?

adamgayoso: Isn't the former problematic if there are API changes/refactors? Maybe it's not an issue for the autoguide specifically?

fritzo: Hmm as I understand tensors and non-tensors are equally susceptible to API changes/refactors: tensors and non-tensor data structures will both be loaded from the saved version, but code will be loaded from the new version since pickle saves classes by name rather than by code.

In practice Pyro's autoguides are very stable: there has been only [one API change](https://github.com/pyro-ppl/pyro/pull/2127) since Pyro 1.0 release, and in fact that changed the tensors as well as code. I have never had an issue with model mismatch because my models tend to be short lived. Note Pyro's [release policy](https://github.com/pyro-ppl/pyro/blob/dev/RELEASE-MANAGEMENT.md) is to not guarantee compatibility of saved models across releases, but in practice the format is very stable.

adamgayoso: I see, thanks for the info. We will explore options, but do you see anything problematic with taking a training step before loading the state dict? Considering every model should be wrapped in a nn.Module.

fritzo: > taking a training step before loading the state dict

Hmm that should probably work, as long as you `pyro.clear_param_store()` after taking that step.

adamgayoso: Thanks, I suppose we'll try this option for now.

scvi/model/base/_base_model.py
```diff
@@ -434,7 +436,18 @@ def load(
         for attr, val in attr_dict.items():
             setattr(model, attr, val)
 
-        model.module.load_state_dict(model_state_dict)
+        # some Pyro modules with AutoGuides may need one training step
+        try:
+            model.module.load_state_dict(model_state_dict)
+        except RuntimeError as err:
+            if isinstance(model.module, PyroBaseModuleClass):
```
romain-lopez, line 443: nice!

tests/models/test_pyro.py
```diff
@@ -0,0 +1,207 @@
...
+
+        self.register_buffer("zero", torch.tensor(0.0))
+        self.register_buffer("one", torch.tensor(1.0))
+        self.register_buffer("ten", torch.tensor(10.0))
+
+        self.linear = PyroModule[nn.Linear](in_features, out_features)
+        # sets a prior over the weight vector
+        # self.linear.weight = PyroSample(
+        #     dist.Normal(self.zero, self.one)
+        #     .expand([out_features, in_features])
```
romain-lopez, line 30: should we keep those, uncomment, or delete?

romain-lopez (approved):
sounds great!

PR #913, opened 2021-02-06
Fix example for load()

saketkc:
Small fix for documentation of [load()](https://docs.scvi-tools.org/en/stable/api/reference/scvi.model.SCVI.load.html?highlight=load#scvi-model-scvi-load)

### DIFF
--- a/scvi/model/base/_base_model.py
+++ b/scvi/model/base/_base_model.py
@@ -423,7 +423,7 @@ class BaseModelClass(ABC):
 
         Examples
         --------
-        >>> vae = SCVI.load(adata, save_path)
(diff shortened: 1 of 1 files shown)


adamgayoso:
Thanks @saketkc! LGTM. For some reason post pytest success on CI fails for PRs from forks. I'll look into that soon.

adamgayoso (approved):

saketkc:
Thanks! I guess the build failure is because of Codacy - the default token probably isn't valid for forks?

adamgayoso:
Yeah it seems, that's kind of annoying...

PR #916, opened 2021-02-08
load pretrained scvi model

galenxing:


scvi/model/_scanvi.py
```diff
@@ -136,6 +136,25 @@ def __init__(
...
 
+    @classmethod
+    def load_pretrained_scvi(
+        cls,
+        scvi_model,
+        unlabeled_category,
+        adata=None,
+        use_gpu=None,
+        **scanvi_kwargs,
+    ):
```
adamgayoso, line 147: needs a docstring and typing

adamgayoso: also should check if scvi model has been trained, and provide a logger warning if it hasn't.

scvi/model/_scanvi.py
```diff
@@ -136,6 +136,25 @@ def __init__(
         )
         self.init_params_ = self._get_init_params(locals())
 
+    @classmethod
+    def load_pretrained_scvi(
```
adamgayoso, line 140: I like `from_scvi_model` better

adamgayoso (changes requested):
LGTM after these small changes.

adamgayoso (approved):

PR #927, opened 2021-02-15
fully bayesian pyro test

adamgayoso:


adamgayoso:
@fritzo -- do you think this is an adequate way to write this class in a device agnostic way? As I understand, using this lambda will ensure that the pyro sample statement does not occur until `forward()` is called? So the reason this works is that by that point the model is on the GPU.

tests/models/test_pyro.py
```diff
@@ -62,13 +99,9 @@ def test_pyro_bayesian_regression(save_path):
...
     # test save and load
-    post_dl = AnnDataLoader(adata, shuffle=False, batch_size=128)
-    mean1 = []
-    with torch.no_grad():
-        for tensors in post_dl:
-            args, kwargs = model._get_fn_args_from_batch(tensors)
-            mean1.append(model(*args, **kwargs).cpu().numpy())
-    mean1 = np.concatenate(mean1)
+    quants = model.guide.quantiles([0.5])
+    sigma_median = quants["sigma"][0].cpu().detach().numpy()
```
fritzo, line 103: nit: it's a little cheaper to detach as early as possible: `.detach().cpu().numpy()`

tests/models/test_pyro.py
```diff
@@ -15,15 +16,51 @@
...
+        self.in_features = in_features
+        self.out_features = out_features
 
         self.register_buffer("zero", torch.tensor(0.0))
         self.register_buffer("one", torch.tensor(1.0))
         self.register_buffer("ten", torch.tensor(10.0))
 
         self.linear = PyroModule[nn.Linear](in_features, out_features)
+        self.linear.weight = PyroSample(
+            lambda prior: dist.Normal(self.zero, self.one)
```
fritzo, line 55: This will work. Note it creates a reference cycle (self → linear → `._pyro_weight` → this lambda, whose closure → self), but you can probably ignore that.

Note the `prior` in `lambda prior` refers to `self.linear` (the owner of `.weight`). Our intended usage was to place any dynamic stuff on that owner, but I see how that would be cumbersome in your case, since the dynamic stuff lives in the root module.

adamgayoso: > Note the prior in lambda prior refers to self.linear (the owner of .weight). Our intended usage was to place any dynamic stuff on that owner, but I see how that would be cumbersome in your case, since the dynamic stuff lives in the root module.

I had noticed that... happy that this works though. Is it more desirable than the callback I had shown you earlier?

fritzo: Yes, your usage in this PR is pretty clean, I believe recommended over the callback you showed me.

adamgayoso: Thanks!

PR #930, opened 2021-02-16
[WIP] Pyro cell2location

vitkl:


scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+        sample_id = None,
+        use_gpu: bool = True,
+        batch_size: int = 1024,
+        **model_kwargs,
+    ):
+        
+        intersect = np.intersect1d(cell_state_df.index, adata.var_names)
+        cell_state_df = cell_state_df.loc[intersect,:]
+        adata = adata[:, intersect]
+        adata.varm['cell_state_varm'] = cell_state_df.values
```
adamgayoso, line 465: what is this?

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+        
+        super(Cell2locationModelClass, self).__init__(adata, 
+                                                      n_fact=cell_state_df.shape[1], 
+                                                      var_names_read=var_names_read,
+                                                      fact_names=cell_state_df.columns,
```
adamgayoso, line 470: just keep in mind during this review that I don't know what `fact_names` is, but there's probably a cleaner way to handle this.

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+            if log_y:
+                y = np.log10(y)
+            ax.plot(np.arange(iter_start, iter_end), y, label='train')
+            ax.set_xlabel('Training epochs')
+            ax.set_ylabel('Reconstruction accuracy (ELBO loss)')
+            ax.legend()
+            plt.tight_layout()
+            
+            
+    def export2adata(self, adata, slot_name='mod'):
```
adamgayoso, line 410: same here, just return the dict and let users do what they want with it (in tutorial just immediately add it back to anndata)

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+    def align_plot_stability(fac1, fac2, name1, name2,
+                             align=True, return_aligned=False):
+        r"""Align columns between two np.ndarrays using scipy.optimize.linear_sum_assignment,
+            then plot correlations between columns in fac1 and fac2, ordering fac2 according to alignment
+
+        :param fac1: np.ndarray 1, factors in columns
+        :param fac2: np.ndarray 2, factors in columns
+        :param name1: axis x name
+        :param name2: axis y name
+        :param align: boolean, match columns in fac1 and fac2 using linear_sum_assignment?
```
adamgayoso, line 328: we use numpydoc style docstrings. They are easier to read and write.

scvi/external/cell2location/_module.py
```diff
@@ -0,0 +1,336 @@
...
+    r"""
+    Function that converts mu/sigma Gamma distribution parametrisation into alpha/beta and
+    returns pyro Gamma distribution with an event of a given shape.
+    A thin wrapper over pyro.dist.Gamma.
+
+    :param mu: mean of Gamma distribution
+    :param sigma: variance of Gamma distribution. Note!!! this is :math:`\sigma^2` not :math:`\sigma`
+    :param alpha: shape parameter of Gamma distribution (mu ** 2 / sigma)
+    :param beta: rate parameter of Gamma distribution (mu / sigma)
+    :param shape: shape of the event / resulting variable. When None the shape is guessed based on input parameters
```
adamgayoso, line 82: nitpicky but it's probably better to write a custom Gamma distribution class that only adds a classmethod to start from the mu sigma parameterization.

scvi/external/cell2location/_module.py
```diff
@@ -0,0 +1,336 @@
...
+
+    if shape is None:
+        alpha = torch.tensor(alpha)
+        beta = torch.tensor(beta)
+    else:
+        alpha = torch.ones(shape) * torch.tensor(alpha)
+        beta = torch.ones(shape) * torch.tensor(beta)
+    return dist.Gamma(alpha, beta)
+
+class LocationModelLinearDependentWMultiExperiment(PyroBaseModuleClass):
```
adamgayoso, line 102: nitpicky but that's a very long class name

scvi/external/cell2location/_module.py
```diff
@@ -0,0 +1,336 @@
...
+                 cell_number_prior={'N_cells_per_location': 8,
+                               'A_factors_per_location': 7,
+                               'Y_combs_per_location': 2.5},
+                 cell_number_var_prior={'N_cells_mean_var_ratio': 1,
+                                   'A_factors_mean_var_ratio': 1,
+                                   'Y_combs_mean_var_ratio': 1},
+                 alpha_g_phi_hyp_prior={'mean': 3, 'sd': 1},
+                 gene_add_alpha_hyp_prior={'mean': 3, 'sd': 1},
+                 gene_add_mean_hyp_prior={'alpha': 1, 'beta': 100},
+                 w_sf_mean_var_ratio=5):
```
adamgayoso, line 117: IMO it's easier to deal with the code if you're explicit about all the hyperparams (each gets it's own parameters).

you should also set your IDE to use the [black formatter](https://black.readthedocs.io/en/stable/) on save. This is mandatory for code in our codebase and is really a painless thing to do.

scvi/external/cell2location/_module.py
```diff
@@ -0,0 +1,336 @@
...
+        self.cell_number_prior = cell_number_prior
+        
+        self.cell_state_mat = cell_state_mat
+        self.register_buffer("cell_state", torch.tensor(cell_state_mat.T))
+        
+        self.guide = AutoNormal(self.model, init_loc_fn=init_to_mean, 
+                                create_plates=self.create_plates)
+        # replace exp transform with softplus https://github.com/pyro-ppl/numpyro/issues/855
+        #self.guide = ExpToSoftplusHandler(self.guide)
+
```
adamgayoso, line 154: couple of comments here:

1. All prior params need to be device agnostic (see `register_buffer` in pytorch) there are also other ways to do this which I'll describe later
2. You should see our bayesian regression example in our tests. We split the logic of the model and the guide, and wrap into the PyroBaseModuleClass. Your model should be a `PyroModel` class like in that example.

scvi/external/cell2location/_module.py
```diff
@@ -0,0 +1,336 @@
...
+                           subsample_size=self.batch_size, 
+                           subsample=idx),
+                pyro.plate("var_axis", self.n_var, dim=-1),
+                pyro.plate("factor_axis", self.n_fact, dim=-1),
+                pyro.plate("combination_axis", self.n_comb, dim=-3),
+                pyro.plate("experim_axis", self.n_exper, dim=-2)]
+
+    def model(self, x_data, idx, obs2sample):
+        # register module with Pyro
+        pyro.module("cell2location", self)
```
adamgayoso, line 181: again, see our new way to structure Pyro code

scvi/external/cell2location/_module.py
```diff
@@ -0,0 +1,336 @@
...
+        
+        m_g_alpha_hyp = pyro.sample('m_g_alpha_hyp',
+                                                Gamma(mu=shape,
+                                                      sigma=shape_var,
+                                                      shape=(1, 1)))
+
+        m_g_beta_hyp = pyro.sample('m_g_beta_hyp',
+                                               Gamma(mu=rate,
+                                                     sigma=rate_var,
+                                                     shape=(1, 1)))
```
adamgayoso, line 201: I believe another way to make this device agnostic is to leverage the device of `x_data`. So you'd make a torch tensor and use the `dev = x_data.device`. You can also give `pyro.sample` a lambda function as input like `pyro.sample("m_g_beta_hyp", lambda self: Gamma(mu=self.rate, sigma=self.rate_var)` where rate and rate_var would be registered buffers.

adamgayoso (commented):
Seems like a lot of comments but I don't think the model is that far off from being ready to go.

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+
+    def __init__(
+        self,
+        adata: AnnData,
+        n_fact: int = 50,
+        var_names_read = None,
+        fact_names = None,
+        sample_id = None,
+        use_gpu: bool = True,
+        batch_size: int = 1024,
```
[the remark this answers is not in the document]
jamestwebber: Was curious to see how this is going, and hey I know this one!

This is short for `factor_names` (class names, cell type names, etc). Since `factor` is only two letters longer I think it makes sense to change it.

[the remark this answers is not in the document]
adamgayoso: I see, so like this is inferred from the scRNA data? If so, the model should just allow the user to input the factors by genes df and there's no need to have all this code to just handle the names.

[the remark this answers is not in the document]
jamestwebber: Vitalii is the source of truth here but I _believe_ yes, these could be inferred from the labels on the reference scRNA data.

vitkl:
> Seems like a lot of comments but I don't think the model is that far off from being ready to go.

Thanks a lot for the feedback! It will take a while to address (because I also want to keep things consistent with main cell2location pymc3 repo) and some of it I think needs discussion - but good to have feedback!

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+import torch
+
+from tqdm.auto import tqdm
+
+import matplotlib
+import matplotlib.pyplot as plt
+from scipy.optimize import linear_sum_assignment
+
+
+class Cell2locationBaseModelClass(BaseModelClass):
```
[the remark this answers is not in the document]
vitkl: I want to use sampling and plotting methods for several version of cell2location model as well as for other latent variable models I that am working on. For that purpose, having multiple classes separating that seems useful.

scvi/external/cell2location/_model.py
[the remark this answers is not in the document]
vitkl: `fact_names ` refers to cell type names for cell2location and to factor names for latent variable models. I agree about using `factor_names` instead (but I need to update that in all of our cell2location repo - so it will take a while). In cell2location, the names of the factors are inferred from a df with cell signatures: https://github.com/YosefLab/scvi-tools/pull/930/files/6aaf614765a89fa1b3da5689b83e354ab44aa05a#diff-f59806d7a4b2b195cb84ed4b7bbe89a982a0f3cea4819904e413f685627b593fR470. However, I also need to use this for other projects and latent variable models - where factor names won't be provided. How do you suggest to deal with this?

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+        if fact_names is None:
+            self.fact_names = pd.Series(['fact_' + str(i) for i in range(self.n_fact)])
+        else:
+            self.fact_names = pd.Series(fact_names)
+            
+        if sample_id is None:
+            self.sample_id = pd.Series(['sample' for i in range(self.n_obs)],
+                                       index=self.obs_names)
+        else:
+            self.sample_id = pd.Series(adata.obs[sample_id], index=self.obs_names)
```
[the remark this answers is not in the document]
vitkl: I see. So you would do this on the fly?
```python
obs2sample = torch.nn.functional.one_hot(batch_key_index)
```
The alternative to this is whether pytorch tensor can be indexed like this:
```python
s_g_gene_add[batch_key_index,:]
```
rather than using one-hot encoded batch like this:
```python
torch.matmul(obs2sample, s_g_gene_add)
```

Is there a way to fetch the total number of batches/experiments (needed to create pyro plates)? (I find it a bit confusing to call both training batch and experiment batch as 'batch')

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+            
+        self.fact_filt = None
```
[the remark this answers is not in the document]
vitkl: This is what you call batch index.

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+        
+        # generate one-hot encoded matrix telling which obs belong to whic samples
+        self.obs2sample_df = pd.get_dummies(self.sample_id)
+        # convert to np.ndarray and register
+        adata.obsm['obs2sample_obsm'] = self.obs2sample_df.values
+        self.n_exper = self.obs2sample_df.shape[1]
+        register_tensor_from_anndata(adata, "obs2sample", "obsm", "obs2sample_obsm")
```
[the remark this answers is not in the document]
vitkl: Replied here: https://github.com/YosefLab/scvi-tools/pull/930#discussion_r578301940

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+            torch.set_default_tensor_type(torch.FloatTensor)
+        
+        
+        if use_gpu:
+            self.module.cuda()
+        
+        train_dl = AnnDataLoader(self.adata, shuffle=True, 
+                                 batch_size=self.batch_size, 
+                                 drop_last=True)
+        pyro.clear_param_store()
```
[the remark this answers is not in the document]
vitkl: I do not know if it is possible (or a good idea) to change pyro plate size on the fly but otherwise would be good to have `drop_last` in the training plan.

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+        max_epochs: int = 30000,
+        lr: float = 0.001,
+        total_grad_norm_constraint: float = 200,
+        use_gpu: Optional[bool] = None,
+        train_size: float = 1,
+        validation_size: Optional[float] = None,
+        batch_size = None,
+        plan_kwargs: Optional[dict] = None,
+        **kwargs,
+    ):
```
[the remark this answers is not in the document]
vitkl: Thanks for explaining! I tried doing something similar to that in `train_pyro_v2` but that did not work because not all prior parameter tensors were registered.

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+        
+        #try:
+        #    self.history = self.trainer.logger.history
+        #except AttributeError:
+        #    self.history = None
+
+        self.is_trained_ = True
+
+        
+    def sample_node1(self, node, num_samples_batch: int = 10):
```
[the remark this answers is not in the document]
vitkl: 1. This function takes one batch of samples (e.g. 10 samples) from the posterior distribution. To generate 1000 samples, `sample_node1` with `num_samples_batch=10` is applied 100 times here: https://github.com/YosefLab/scvi-tools/pull/930/files/6aaf614765a89fa1b3da5689b83e354ab44aa05a#diff-f59806d7a4b2b195cb84ed4b7bbe89a982a0f3cea4819904e413f685627b593fR221-R235

2. On using scVI data loader as you suggest: 
A) Concatenating local parameters that come out of each data loader batch is non-trivial. It probably needs digging out which parameters belong to a batch plate (which means that batch plate should always have the same name, e.g.  `obs_plate`), then concatenating those parameters in the axis of that plate.

B) Global parameters will be sampled in every dataloader batch essentially meaning that while local parameters will have 1000 samples, global parameters will have `1000 * n_obs / batch_size` samples. I am not sure how to deal with this ... use only the last/first dataloader batch?

C) Doing a pass through data with dataloader (and repeating that 100 times to get 100 * 10 posterior samples) can be slow because it would move X to the GPU too many times. Asking for 1000 samples at the same time will likely hit GPU memory when there are many global parameters.

D) Unless guide is performing the autoencoding operation, X and other observed variables are actually not used by Predictive. 

E) Given D, it seems sensible to ignore dataloader batches and just give the model indices of all observations. We also do not want to move all observations in X to GPU so it seems sensible to replace it with arbitrary tensor.

Would be good if scVI provides general methods for doing Predictive for pyro models because this step is quite model agnosic. 

What do you think?

scvi/external/cell2location/_model.py
```diff
@@ -0,0 +1,572 @@
...
+        for it in tqdm(range(n_sampl_batches - 1)):
+            # sample remaining batches
+            post_samples = self.sample_all1(num_samples_batch=num_samples_batch)
+
+            # concatenate batches
+            self.adata.uns['mod']['post_samples'] = {k: np.concatenate((self.adata.uns['mod']['post_samples'][k],
+                                                               post_samples[k]), axis=0)
+                                            for k in post_samples.keys()}
+
+    def sample_posterior(self, node='all',
```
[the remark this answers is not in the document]
vitkl: I see. I will change that. So here we have 2 layers
{'mean, quantiles, sd, all samples': {'variable names': np.ndarray}

scvi/external/cell2location/_model.py
vitkl: This subsets adata and `cell_state_df` with signatures of cell types to common genes. Doing that in the notebook does not guarantee that users will always do it or do it correctly.
I can add a error instead

```python
all_matching = np.all(cell_state_df.index == adata.var_names)
if not all_matching:
    raise ValueError('cell_state_df and spatial adata should have the same genes')
```

scvi/external/cell2location/_model.py
vitkl: Where would be a good place to intermediately store posterior samples? (nowhere, just give them to users?)

scvi/external/cell2location/_model.py
vitkl: Yes, I need to change that.

scvi/external/cell2location/_module.py
vitkl: Do you mean a class that inherits from `dist.Gamma`?

scvi/external/cell2location/_module.py
vitkl: How do you name different versions of the model?
with abbreviations? E.g. LMLDWME for this?

scvi/external/cell2location/_module.py
vitkl: Do you mean a keyword argument for every prior value?

Which IDE do you use, VSCode? I used PyCharm but it breaks my RAM and stopped using it last month (it's just jupyter notebook at the moment).

scvi/external/cell2location/_module.py
vitkl: So you would do 2 classes/functions (the question is about where to add `create_plates`)?
1. `class LocationModelLinearDependentWMultiExperimentModel(PyroModel)` which implements model function as `forward` method
2. `class LocationModelLinearDependentWMultiExperiment(PyroBaseModuleClass)` which initialised model and guide:
```python
class LocationModelLinearDependentWMultiExperiment(PyroBaseModuleClass):

    def init(self)
        self.model = LocationModelLinearDependentWMultiExperimentModel(..., create_plates=self.create_plates)
        self.guide = AutoNormal(..., create_plates=self.create_plates)

    def create_plates(self, ...):
        ...
```

scvi/external/cell2location/_module.py
vitkl: Is the lambda function approach necessary if `self.rate, sigma=self.rate_var` are registered buffers?

scvi/external/cell2location/_module.py
adamgayoso: > Do you mean a keyword argument for every prior value?

Yes. And we use VSCode. We might soon also offer a VSCode docker dev container so everyone basically has the same vscode settings.

scvi/external/cell2location/_model.py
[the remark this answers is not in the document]
adamgayoso: I think using Mixin classes is cleaner here. Right now it's difficult to understand the inheritance structure, where Mixins provide isolated functionalities.

scvi/external/cell2location/autoguide.py
```diff
@@ -0,0 +1,384 @@
...
+        return result
+
+
+@biject_to.register(constraints.positive)
+@transform_to.register(constraints.positive)
+def _transform_to_positive(constraint):
+    return SoftplusTransform()
+
+
+class AutoNormalEncoder(AutoGuide):
```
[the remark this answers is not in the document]
romain-lopez: I guess then it's not an AutoGuide anymore :) That looks like pretty involved work to get this functional! Very interesting!

scvi/external/cell2location/tutorial/cell2location_analysis_2_estimating_signatures.ipynb
```diff
@@ -0,0 +1,1466 @@
+{
```
romain-lopez, line 1: <p>This is alright for now, but you want people to be able to fully run the analysis and not rely on external software</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/YosefLab/scvi-tools/pull/930/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

scvi/external/cell2location/_cell2location_v1.py
```diff
@@ -0,0 +1,91 @@
+import numpy as np
+import pandas as pd
+from anndata import AnnData
+
+import scvi
+from scvi.model.base import BaseModelClass
+
+from ._base import Cell2locationBaseModule, PltExportMixin, TrainSampleMixin
+from ._cell2location_v1_module import LocationModelLinearDependentWMultiExperimentModel
```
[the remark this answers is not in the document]
vitkl: With pyro there are actually 3 classes:
1. Model class which defines the model and inherits from `pyro.PyroModule`
2. Module class which initialises #1 model class and a guide, and inherits from nn.Module (crucially not PyroModule) and scvi `PyroBaseModuleClass`
3. User facing model class which initialises #2 Module class, does training, etc and inherits from scvi `BaseModelClass`

scvi/external/cell2location/_cell2location_v1.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+def intersect_var(adata, cell_state_df):
```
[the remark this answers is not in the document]
vitkl: Returned is a subset of genes from the original adata object and pd.DataFrame. `scvi.data.setup_anndata` does not accept view - hence copying

scvi/external/cell2location/_cell2location_v1_module.py
```diff
@@ -0,0 +1,351 @@
...
+        gene_add_alpha_hyp_prior={"alpha": 9.0, "beta": 3.0},
+        gene_add_mean_hyp_prior={
+            "alpha": 1.0,
+            "beta": 100.0,
+        },  # TODO initialise as average of empty locations
+        w_sf_mean_var_ratio=5.0,
+    ):
+
+        super().__init__()
+
```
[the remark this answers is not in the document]
vitkl: What do you think about this style of describing a model?
https://github.com/YosefLab/scvi-tools/pull/930/files#diff-fa36ab83dfef490cbb6dcc3d83dc7e43f81357d51b087173e30a3c4799537a36R83-R104

scvi/external/cell2location/_reference_module.py
```diff
@@ -0,0 +1,472 @@
...
+        aver = sparse_subset.mean(0)
+        averages_mat = np.concatenate((averages_mat, aver))
+    averages_mat = averages_mat[1:, :].T
+    averages_df = pd.DataFrame(data=averages_mat, index=var_names, columns=all_clusters)
+
+    return averages_df
+
+
+@biject_to.register(constraints.positive)
+@transform_to.register(constraints.positive)
```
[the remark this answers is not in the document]
vitkl: Not sure if this applies to only models defined in this file. Does this affect everything in scVI package?

scvi/external/cell2location/autoguide.py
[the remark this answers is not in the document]
vitkl: > different heads for different parts of the parameters

Can you please explain what you mean by head and by part?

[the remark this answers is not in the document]
vitkl: This class defines the following operations:
loc = NN(data) @ hidden2loc
scales = softplus(NN(data) @ hidden2scales - 2)
Where NN is encoder networks which could be general of specific to each variable, and hidden2loc/hidden2scales are tensors that convert hidden layer activations to locs/scales

scvi/external/cell2location/tutorial/cell2location_analysis_2_estimating_signatures.ipynb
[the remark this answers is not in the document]
vitkl: Hm, I actually do not understand this warning (never used lightning) and it is present in both notebooks / both models. Maybe you or @adamgayoso can explain?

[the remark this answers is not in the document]
vitkl: <p>Samples dict must exist at some point. In the previous iteration is was just added as an attribute to the model class without returning but Adam suggested to return it. I like to have posterior summary saved for all parameters somewhere</p>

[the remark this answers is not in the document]
vitkl: <p>I think users need to assess if this looks very non-diagonal and investigate if that is the case. This specific plot does not look great - see the plot from the other notebook for comparison</p>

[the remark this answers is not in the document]
adamgayoso: <p>yeah this is because pyro does the optimization, we should catch this warning.</p>

scvi/external/cell2location/_base.py
```diff
@@ -0,0 +1,1060 @@
...
+
+    - training models using minibatches (standard scVI interface) and using full data (copies data to GPU only once).
+    - computing median and quantiles of the posterior distribution using both direct and amortised inference
+    - generating samples from posterior distribution
+
+    """
+
+    @property
+    def _plan_class(self):
+        return PyroTrainingPlan
```
adamgayoso, line 168: this shouldn't be necessary if you rebase your branch on the head of master

scvi/external/cell2location/_cell2location_v1_module.py
```diff
@@ -0,0 +1,351 @@
...
+    def _get_fn_args_from_batch(tensor_dict):
+        x_data = tensor_dict[_CONSTANTS.X_KEY]
+        ind_x = tensor_dict["ind_x"].long().squeeze()
+        batch_index = tensor_dict[_CONSTANTS.BATCH_KEY]
+        return (x_data, ind_x, batch_index), {}
+
+    def _get_fn_args_full_data(self, adata):
+        x_data = get_from_registry(adata, _CONSTANTS.X_KEY)
+        if issparse(x_data):
+            x_data = np.array(x_data.toarray())
```
adamgayoso, line 152:
```suggestion
            x_data = np.asarray(x_data.toarray())
```
asarray should be cheaper

scvi/external/cell2location/_cell2location_v1.py
```diff
@@ -0,0 +1,91 @@
...
+    Reimplementation of cell2location [Kleshchevnikov20]_ model. User-end model class.
+
+    https://github.com/BayraktarLab/cell2location
+
+    Parameters
+    ----------
+    adata
+        spatial AnnData object that has been registered via :func:`~scvi.data.setup_anndata`.
+    cell_state_df
+        pd.DataFrame with reference expression signatures
```
adamgayoso, line 42: more description of the appropriate shape of this.

scvi/external/cell2location/_base.py
```diff
@@ -0,0 +1,1060 @@
+from datetime import date
+from typing import Optional
```
adamgayoso, line 2: run import sort `isort`

scvi/external/cell2location/_base.py
```diff
@@ -0,0 +1,1060 @@
...
+
+
+class Cell2locationBaseModule(PyroBaseModuleClass):
+    def __init__(
+        self,
+        model,
+        amortised: bool = False,
+        single_encoder: bool = True,
+        encoder_kwargs=None,
+        data_transform="log1p",
```
adamgayoso, line 30: we have a Literal type you can use for a few string options

scvi/external/cell2location/_base.py
```diff
@@ -0,0 +1,1060 @@
...
+        batch_size = self.module.model.batch_size
+        # initialise guide params (warmup)
+        train_dl = AnnDataLoader(self.adata, shuffle=True, batch_size=batch_size)
+        for tensor_dict in train_dl:
+            args, kwargs = self.module._get_fn_args_from_batch(tensor_dict)
+            args = [a.to(device) for a in args]
+            kwargs = {k: v.to(device) for k, v in kwargs.items()}
+            self.to_device(device)
+            self.module.guide(*args, **kwargs)
+            break
```
adamgayoso, line 302: for this, we can use the callback in the test_pyro file (I think we can move it to the `scvi.train` module in the callbacks file)

scvi/external/cell2location/_base.py
```diff
@@ -0,0 +1,1060 @@
...
+
+        args = [a.to(device) for a in args]
+        kwargs = {k: v.to(device) for k, v in kwargs.items()}
+        self.to_device(device)
+
+        if not continue_training or not self.is_trained_:
+            # models share param store, make sure it is cleared before training
+            pyro.clear_param_store()
+        # initialise guide params (warmup)
+        self.module.guide(*args, **kwargs)
```
adamgayoso, line 211: the `PyroJitGuideWarmup` can be used here too

scvi/external/cell2location/_cell2location_v1_module.py
```diff
@@ -0,0 +1,351 @@
...
+        self.register_buffer("eps", torch.tensor(1e-8))
+
+    @staticmethod
+    def _get_fn_args_from_batch(tensor_dict):
+        x_data = tensor_dict[_CONSTANTS.X_KEY]
+        ind_x = tensor_dict["ind_x"].long().squeeze()
+        batch_index = tensor_dict[_CONSTANTS.BATCH_KEY]
+        return (x_data, ind_x, batch_index), {}
+
+    def _get_fn_args_full_data(self, adata):
```
adamgayoso, line 149: I'm not actually sure why this function is necessary. You can make a data loader with `shuffle=False` `batch_size=adata.shape[0]`, `indices`

adamgayoso: and then use the batch one

scvi/external/cell2location/_cell2location_v1.py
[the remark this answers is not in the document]
adamgayoso: I don't see the value in this function. Your init can check if the names are valid and can inplace intersect them `adata._inplace_subset_var` (with a strong warning that you are doing this).

scvi/external/cell2location/_base.py
vitkl: @adamgayoso merged the most recent master branch into this branch, does this solve it?

scvi/external/cell2location/_base.py
```diff
@@ -0,0 +1,1060 @@
...
+            )
+            # sample global parameters
+            for tensor_dict in train_dl:
+                args, kwargs = self.module._get_fn_args_from_batch(tensor_dict)
+                args = [a.to(device) for a in args]
+                kwargs = {k: v.to(device) for k, v in kwargs.items()}
+                self.to_device(device)
+                break
+
+        means = self.module.guide.quantiles([q], *args, **kwargs)
```
adamgayoso, line 547: our `automovedata` decorator would work here

scvi/external/cell2location/_base.py
```diff
@@ -0,0 +1,1060 @@
...
+        # sample local parameters
+        i = 0
+        for tensor_dict in train_dl:
+            if i == 0:
+                args, kwargs = self.module._get_fn_args_from_batch(tensor_dict)
+                args = [a.to(device) for a in args]
+                kwargs = {k: v.to(device) for k, v in kwargs.items()}
+                self.to_device(device)
+
+                means = self.module.guide.quantiles([q], *args, **kwargs)
```
adamgayoso, line 448: automovedata (see other comment)

scvi/external/cell2location/_base.py
adamgayoso: though there's a chance it needs to be tweaked.

Overall though, it removes all this device specific code

scvi/external/cell2location/_cell2location_v1_module.py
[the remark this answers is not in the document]
vitkl: Added model description - is this what you had in mind?

scvi/external/cell2location/_base.py
vitkl: Can you please give me an example?

scvi/external/cell2location/_base.py
vitkl: what do you mean? is this a command-line app?

scvi/external/cell2location/_base.py
vitkl: This is also just a single line, and it will be not necessary when this is addressed https://github.com/YosefLab/scvi-tools/issues/1058

scvi/external/cell2location/_base.py
vitkl: Could I ask you to point me to usage example?

scvi/external/cell2location/_base.py
adamgayoso: yes, if you install `pre-commit` as described in our [contributing guide](https://docs.scvi-tools.org/en/stable/contributing/index.html), it won't let you push a commit without it passing black formatting, flake8, and isort (and will fix it for you)

scvi/external/cell2location/_cell2location_v1.py
```diff
@@ -0,0 +1,81 @@
...
+from scvi.model.base import BaseModelClass
+
+from ._base import PltExportMixin, TrainSampleMixin
+from ._cell2location_v1_module import (
+    Cell2locationBaseModule,
+    LocationModelLinearDependentWMultiExperimentModel,
+)
+
+
+class Cell2location(TrainSampleMixin, BaseModelClass, PltExportMixin):
```
adamgayoso, line 15:
```suggestion
class Cell2location(TrainSampleMixin, PltExportMixin, BaseModelClass):
```

adamgayoso (changes requested):

scvi/external/cell2location/_cell2location_v1_module.py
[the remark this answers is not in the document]
romain-lopez: sounds good!

scvi/external/cell2location/_reference_module.py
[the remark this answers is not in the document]
romain-lopez: I see, it probably not. Is there a way to inherit from a class that adds it? Unsure

scvi/external/cell2location/autoguide.py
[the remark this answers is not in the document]
romain-lopez: let's say you can group the parameters into groups [p_1, p_2, p_3] with p_1 abundance of cell types into spots, p_2 some other local parameters etc..

Then your encoder has the following architecture
x -> h1-> h2 (xx hidden layers) -> [p_1, p_2, p_3]

But you could try
x -> h1
h1 -> h2 -> p1
h1 -> h3 -> p2 (two differents "heads")

scvi/external/cell2location/tutorial/cell2location_analysis_2_estimating_signatures.ipynb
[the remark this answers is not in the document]
romain-lopez: hum, ok. I still think that the QC code should be a standalone function and not make appear those samples, for better user experience.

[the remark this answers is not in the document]
romain-lopez: I understand this, but you should just call this model.QC_plots() and it would show you the two plots you have to look at

scvi/external/cell2location/autoguide.py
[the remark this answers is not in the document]
vitkl: I see what you mean. In a way though, nodes in current h2 can be split depending on which p_1, p_2 or p_3 they affect. So current architecture could represent the function which you are proposing.

scvi/external/cell2location/_reference_module.py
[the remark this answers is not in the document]
vitkl: It will probably change the positive transform for all of scVI

scvi/external/cell2location/_cell2location_v1.py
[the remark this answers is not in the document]
vitkl: I renamed #1 pyro model class to *PyroModel. Not sure that the long model name is a problem. It names relevant features and I know how it relates to pymc3 implementation and to other versions of the model that I tried. Users interact with a class that has a reasonably short name class anyway.

vitkl:
Interestingly, the last test works but cell2location with full data and non-amortised inference gives an error which I do not quite understand. See HTML attached (can be reproduced by taking this notebook to colab).
[cell2location_analysis_3_scvi_ln_full_data.html.zip](https://github.com/YosefLab/scvi-tools/files/6490759/cell2location_analysis_3_scvi_ln_full_data.html.zip)

scvi/external/cell2location/autoguide.py
```diff
@@ -0,0 +1,384 @@
...
+
+from scvi.nn import FCLayers
+
+
+class FCLayersPyro(FCLayers, PyroModule):
+    pass
+
+
+class AutoGuideList(PyroAutoGuideList):
+    def quantiles(self, quantiles, *args, **kwargs):
```
fritzo, line 25: Looking forward to accepting this upstream.

scvi/external/cell2location/autoguide.py
fritzo, line 51: Hi @vitkl, @martinjankowiak and @eb8680 and I have been wanting something like this in Pyro for a long time, and we'd be happy to maintain part or all of this in the Pyro codebase (with tests, docs, etc.).

scvi/external/cell2location/autoguide.py
```diff
@@ -0,0 +1,384 @@
...
+
+            # Collect independence contexts.
+            self._cond_indep_stacks[name] = site["cond_indep_stack"]
+
+            # add linear layer for locs and scales
+            param_dim = (self.n_hidden, self.amortised_plate_sites["sites"][name])
+            # init_param = torch.normal(torch.full(size=param_dim, fill_value=0.),
+            #                          torch.full(param_dim, 1. / np.sqrt(self.n_hidden)),
+            #                          device=site["value"].device)
+            init_param = np.random.normal(
```
martinjankowiak, line 186: i think you generally want to avoid using numpy for randomness so setting the torch seed is sufficient for reproducibility

scvi/external/cell2location/autoguide.py
```diff
@@ -0,0 +1,384 @@
...
+        """
+
+        linear_locs = _deep_getattr(self.hidden2locs, name)
+        linear_scales = _deep_getattr(self.hidden2scales, name)
+
+        if not self.single_encoder:
+            # when using multiple encoders extract hidden layer for this parameter
+            encoded_hidden = encoded_hidden[name]
+
+        locs = encoded_hidden @ linear_locs
```
martinjankowiak, line 242: why are you doing this by hand? don't you want bias terms?

scvi/external/cell2location/autoguide.py
```diff
@@ -0,0 +1,384 @@
...
+            Pyro model args
+        kwargs
+            Pyro model kwargs
+        -------
+
+        """
+        in_names = self.amortised_plate_sites["in"]
+        x_in = [kwargs[i] if i in kwargs.keys() else args[i] for i in in_names]
+        # apply data_transform
+        x_in = [self.data_transform(x) for x in x_in]
```
martinjankowiak, line 262: support `data_transform==None` for identity?

scvi/external/cell2location/autoguide.py
```diff
@@ -0,0 +1,384 @@
...
+        ----------
+        args
+            Pyro model args
+        kwargs
+            Pyro model kwargs
+        -------
+
+        """
+        in_names = self.amortised_plate_sites["in"]
+        x_in = [kwargs[i] if i in kwargs.keys() else args[i] for i in in_names]
```
martinjankowiak, line 260: why is this either/or?

martinjankowiak (commented):
@vitkl looks great! i'm still not entirely sure about which assumptions are being made here, but like @fritzo said it'd be great to have something like this in pyro so if you want to consider making a PR upstream maybe that would be the right place for a more detailed discussion. in any case cool stuff

scvi/external/cell2location/autoguide.py
vitkl: Doing the last transformation separately allows using the same last hidden layer for all model variables.
Also, scVI encoder network applies activation after the last layer. This way AutoGuide class does not care if the activation was applied to the last layer. I also found that initial values need to be much smaller than default values in `torch.nn.Linear`. Do you suggest replacing this with `torch.nn.Linear`?

vitkl: @martinjankowiak Doing the last transformation separately allows using the same last hidden layer for all model variables.
Also, scVI encoder network applies activation after the last layer. This way AutoGuide class does not care if the activation was applied to the last layer. I also found that initial values need to be much smaller than default values in `torch.nn.Linear`. Do you suggest replacing this with `torch.nn.Linear`?

scvi/external/cell2location/autoguide.py
vitkl: Default value could also be `data_transform=lambda x: x`.
Log-transform is good default in the context of scVI.

scvi/external/cell2location/autoguide.py
vitkl: To enable choosing variables by name in `kwargs` or by index in `args`, where kwargs/args are the arguments to the model/guide/etc.

scvi/external/cell2location/autoguide.py
martinjankowiak: no that makes sense but why no bias?

scvi/external/cell2location/autoguide.py
vitkl: Just created a PR here: https://github.com/pyro-ppl/pyro/pull/2849

scvi/external/cell2location/autoguide.py
vitkl: No big reason. I did not think that was necessary - and it is more code to add it.

scvi/external/cell2location/_reference_module.py
```diff
@@ -0,0 +1,741 @@
...
+            obs2extra_categoricals = torch.cat(
+                [
+                    one_hot(
+                        extra_categoricals[:, i].view((extra_categoricals.shape[0], 1)),
+                        n_cat,
+                    )
+                    for i, n_cat in enumerate(self.n_extra_categoricals)
+                ],
+                dim=1,
+            )
```
vitkl, line 257: @adamgayoso is this a correct way to use categorical covariates?

adamgayoso: seems good to me. torch nn embedding might also be of interest, though I haven't looked carefully enough at what you're doing

vitkl:
Moved the model to https://github.com/BayraktarLab/cell2location/tree/master/cell2location/models

vitkl:
Moved the model to https://github.com/BayraktarLab/cell2location/tree/master/cell2location/models and closing this PR

PR #959, opened 2021-03-01
fix docstrings example stereoscope

giovp:
Hi,

noticed minor bugs in the docstrings examples of stereoscope and thought to push a fix real quick. Feel free to reject if not needed.

looks really great btw, looking forward to official release!

### DIFF
--- a/scvi/external/stereoscope/_model.py
+++ b/scvi/external/stereoscope/_model.py
@@ -26,8 +26,8 @@ class RNAStereoscope(UnsupervisedTrainingMixin, BaseModelClass):
     Examples
     --------
     >>> sc_adata = anndata.read_h5ad(path_to_sc_anndata)
-    >>> scvi.data.setup_anndata(sc_adata, label_key="labels")
-    >>> stereo = scvi.external.RNAStereoscope(sc_adata)
+    >>> scvi.data.setup_anndata(sc_adata, labels_key="labels")
+    >>> stereo = scvi.external.stereoscope.RNAStereoscope(sc_adata)
     >>> stereo.train()
     """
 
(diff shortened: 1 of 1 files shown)


scvi/external/stereoscope/_model.py
```diff
@@ -26,8 +26,8 @@ class RNAStereoscope(UnsupervisedTrainingMixin, BaseModelClass):
     Examples
     --------
     >>> sc_adata = anndata.read_h5ad(path_to_sc_anndata)
-    >>> scvi.data.setup_anndata(sc_adata, label_key="labels")
-    >>> stereo = scvi.external.RNAStereoscope(sc_adata)
+    >>> scvi.data.setup_anndata(sc_adata, labels_key="labels")
+    >>> stereo = scvi.external.stereoscope.RNAStereoscope(sc_adata)
```
adamgayoso, line 30: it should be accessible from just `scvi.external` right?

adamgayoso: both scRNA and spatial

adamgayoso (changes requested):
Thanks for the contribution! One quick comment

scvi/external/stereoscope/_model.py
giovp:
```python
scvi.external
AttributeError                            Traceback (most recent call last)
<ipython-input-2-f706a22e0ee9> in <module>
----> 1 scvi.external

AttributeError: module 'scvi' has no attribute 'external'
```

my feeling (but might be wrong) is that you should import it explicitly here
https://github.com/YosefLab/scvi-tools/blob/9b9370aa502b308f84e3129a7c940a9bea06426b/scvi/__init__.py#L10

If I import `external` there, it works.

another (very) minor pet peeve, but adding the `''` to `pip install -e .'[dev,tutorials]'` makes `zsh` users happy :)

adamgayoso: I just committed the latter change you suggested.

I suppose that pattern we can use in the examples here is `from scvi.external import RNAStereoscope`

I think we just aren't 100% sure we want to import into the top-level yet, as we are still figuring out what external is to some extent. This is also consistent with what scanpy does.

giovp: whatever you think it's best, I actually like the idea of not having explicit imports, and just implicit `import scvi` in the examples. Let me know what to do.

btw also really cool that cellassign is already there! Is this the port of the original implementation in tensorflow or the more recent one in torch https://github.com/camlab-bioml/astir/ ?

adamgayoso: ok we can leave it like you have it haha. 

This is a port of the original implementation, except that we minibatch during inference where they do "full-batch" EM. We'll have a notebook soon showing consistency with their implementation. We'd be happy for you to try it out!

We also talked to the Astir team about using scvi-tools... stay tuned!

PR #964, opened 2021-03-02
de silent mode

### LINKED ISSUE #863: Silent mode for `differential_expression`
I saw that both `poisson_gene_selection` and the `Trainer` class have an option `silent` that if True disables the progress bar.

Would it be possible to add this option to the `differential_expression` function? 

When looping through a custom combination of clusters sometimes I would prefer to disable it to prevent cluttering the output as in the picture below.

image

galenxing:
Fixes #863

### DIFF
--- a/scvi/_docs.py
+++ b/scvi/_docs.py
@@ -42,4 +42,6 @@ batchid2
     are decoded on each group in `group1` and `group2`, respectively.
 fdr_target
     Tag features as DE based on posterior expected false discovery rate.
+silent
+    If True, disables the progress bar. Default: False.
 """
--- a/scvi/model/_peakvi.py
+++ b/scvi/model/_peakvi.py
@@ -350,8 +350,8 @@ class PEAKVI(ArchesMixin, VAEMixin, UnsupervisedTrainingMixin, BaseModelClass):
         batchid1: Optional[Iterable[str]] = None,
         batchid2: Optional[Iterable[str]] = None,
         fdr_target: float = 0.05,
-        two_sided: bool = True,
         silent: bool = False,
+        two_sided: bool = True,
         **kwargs,
     ) -> pd.DataFrame:
         r"""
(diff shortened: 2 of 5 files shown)


adamgayoso (changes requested):
We need to add the docstring entry in _docs.py and when you do it you have to make sure the order is consistent everywhere (arguments with docstrings)

scvi/model/_peakvi.py
```diff
@@ -351,6 +346,7 @@ def differential_accessibility(
         batchid2: Optional[Iterable[str]] = None,
         fdr_target: float = 0.05,
         two_sided: bool = True,
+        silent: bool = False,
```
adamgayoso, line 349: move this underneath fdr target and then revert back the {doc_differential_expression}

scvi/model/_peakvi.py
```diff
@@ -360,9 +356,49 @@ def differential_accessibility(
...
+            Subset of categories from `batch_key` registered in :func:`~scvi.data.setup_anndata`,
+            e.g. [`'batch1'`, `'batch2'`, `'batch3'`], for `group1`. Only used if `batch_correction` is `True`, and
+            by default all categories are used.
+        batchid2
+            Same as `batchid1` for group2. `batchid2` must either have null intersection with `batchid1`,
+            or be exactly equal to `batchid1`. When the two sets are exactly equal, cells are compared by
+            decoding on the same batch. When sets have null intersection, cells from `group1` and `group2`
+            are decoded on each group in `group1` and `group2`, respectively.
+        fdr_target
+            Tag features as DE based on posterior expected false discovery rate.
```
adamgayoso, line 397: this needs to be reverted back

adamgayoso (changes requested):

scvi/_docs.py
```diff
@@ -42,4 +42,7 @@
     are decoded on each group in `group1` and `group2`, respectively.
 fdr_target
     Tag features as DE based on posterior expected false discovery rate.
+silent
+    If True, disables the progress bar. Default: False.
+
```
adamgayoso, line 47: remove this extra line

PR #966, opened 2021-03-02
Add logging of scanvi train and val classification loss

galenxing:


scvi/train/_trainingplans.py
```diff
@@ -516,13 +520,55 @@ def validation_step(self, batch, batch_idx, optimizer_idx=0):
...
             "n_obs": reconstruction_loss.shape[0],
         }
+        if hasattr(scvi_losses, "classification_loss"):
+            loss_dict["classification_loss"] = scvi_losses.classification_loss
+            loss_dict["n_labelled_tensors"] = scvi_losses.n_labelled_tensors
+        return loss_dict
+
+    def training_epoch_end(self, outputs):
+        super().training_epoch_end(outputs)
+        log_classification_loss = False
```
adamgayoso, line 538: what is this line doing?

scvi/train/_trainingplans.py
```diff
@@ -516,13 +520,55 @@ def validation_step(self, batch, batch_idx, optimizer_idx=0):
...
+                log_classification_loss = True
+
+        if log_classification_loss:
+            self.log(
+                "classification_loss_train", classifier_loss / total_labelled_tensors
+            )
+
+    def validation_epoch_end(self, outputs):
+        super().validation_epoch_end(outputs)
+        log_classification_loss = False
```
adamgayoso, line 556: and this line

adamgayoso (changes requested):

scvi/train/_trainingplans.py
galenxing: log as in calling self.log. If there is classification loss to log, this flag is set to true and the loss is logged.

PR #968, opened 2021-03-02
backwards compatibility loading

galenxing:
allows loading of models from scvi version 0.8.0

- still need to double check other versions
- need to add tests (after merging trained models into scvi-data)

### DIFF
--- a/scvi/model/base/_base_model.py
+++ b/scvi/model/base/_base_model.py
@@ -20,7 +20,12 @@ from scvi.dataloaders import AnnDataLoader
 from scvi.model._utils import parse_use_gpu_arg
 from scvi.module.base import PyroBaseModuleClass
 
-from ._utils import _initialize_model, _load_saved_files, _validate_var_names
+from ._utils import (
+    _attempt_backwards_compatible_load,
+    _initialize_model,
+    _load_saved_files,
+    _validate_var_names,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -348,6 +353,10 @@ class BaseModelClass(ABC):
                 pyro.clear_param_store()
                 model.module.load_state_dict(model_state_dict)
             else:
+                success = _attempt_backwards_compatible_load(
+                    model.module, model_state_dict
+                )
+            if success is not True:
                 raise err
 
         model.to_device(device)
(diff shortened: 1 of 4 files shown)


adamgayoso:
we should add a warning that we will deprecate this in version 1.0 and that users should resave their models?

tests/models/test_models.py
```diff
@@ -226,6 +228,29 @@ def test_save_load_model(cls, adata, save_path):
     assert model.is_trained is True
 
 
+@pytest.mark.internet
+def test_backwards_compatible_loading(save_path):
+    def download_080_models(save_path):
+        file_path = (
+            "https://github.com/galenxing/scVI-data/raw/master/testing_models.tar.gz"
```
adamgayoso, line 235:
```suggestion
            "https://github.com/yoseflab/scVI-data/raw/master/testing_models.tar.gz"
```

scvi/model/base/_base_model.py
```diff
@@ -348,6 +353,10 @@ def load(
                 pyro.clear_param_store()
                 model.module.load_state_dict(model_state_dict)
             else:
+                success = _attempt_backwards_compatible_load(
+                    model.module, model_state_dict
+                )
+            if success is not True:
```
adamgayoso, line 359: if it's Pyro, success will not be set

adamgayoso (changes requested):

scvi/model/base/_utils.py
```diff
@@ -84,6 +86,20 @@ def _validate_var_names(adata, source_var_names):
...
+        for key, value in model_state_dict.items():
+            new_key = key.replace(" ", "_")
+            new_model_state_dict.update({new_key: value})
+        module.load_state_dict(new_model_state_dict)
+    except RuntimeError:
+        return False
+
+    return True
+
+
```
adamgayoso, line 102: None of this will be necessary, there must've been a bad rebase somewhere and I just fixed FCLayers on master to be consistent with 0.8

scvi/model/base/_base_model.py
adamgayoso: none of this will be necessary

adamgayoso (approved):

PR #970, opened 2021-03-02
faster count check

adamgayoso:


adamgayoso:
We should be able to use numpy [reduce](https://numpy.org/doc/stable/reference/generated/numpy.ufunc.reduce.html) now to make this a robust numpy-like function

galenxing (approved):
lgtm

PR #996, opened 2021-03-09
LossRecorder defaults and literally log the loss

mjayasur:


scvi/module/base/_base_module.py
```diff
@@ -29,13 +29,17 @@ class LossRecorder:
         KL divergence associated with each observation in the minibatch.
     kl_global
         Global kl divergence term. Should be one dimensional with one value.
+    **kwargs
+        Set attributes for this LossRecorder object
```
adamgayoso, line 33:
```suggestion
        Additional metrics can be passed as keyword arguments and will
        be available as attributes of the object.
```

scvi/module/base/_base_module.py
```diff
@@ -51,6 +56,7 @@ def __init__(
         self._kl_global = (
             kl_global if isinstance(kl_global, dict) else dict(kl_global=kl_global)
         )
+        self.record_elbo = True
```
adamgayoso, line 59: I think this PR will work as is, but for now, let's just remove the record elbo stuff (so always record elbo, keep defaults as you set them)

adamgayoso (changes requested):

scvi/train/_trainingplans.py
```diff
@@ -104,12 +104,17 @@ def forward(self, *args, **kwargs):
         """Passthrough to `model.forward()`."""
         return self.module(*args, **kwargs)
 
+    def return_and_log_loss(self):
+        """Log loss and return dictionary of metrics."""
+        pass
+
```
adamgayoso, line 110: don't need

adamgayoso (changes requested):

PR #1016, opened 2021-04-05
changes numpy.array() on tensors to .numpy()

### LINKED ISSUE #1015: Use `.numpy()` to make torch into numpy instead of `np.array()` constructor
This can be found in many places.

mjayasur:
close #1015


wukathy (approved):
lgtm

PR #1024, opened 2021-04-12
Fix SaveBestState warning on first epoch

talashuach:
SaveBestState callback prints a warning on the first epoch since `current` is None at that point. Fixed to check only after the first epoch.


adamgayoso (approved):
LGTM

PR #1030, opened 2021-04-16
Fix bug missing items() call in loop for extending categorical covariates on transfer anndata setup

morris-frank:
Hey,

There was a missing items call for Looping over the extra categories dictionary.

### DIFF
--- a/tests/dataset/test_anndata.py
+++ b/tests/dataset/test_anndata.py
@@ -246,6 +246,38 @@ def test_extra_covariates():
     pd.testing.assert_frame_equal(df1, df2)
 
 
+def test_extra_covariates_transfer():
+    adata = synthetic_iid()
+    adata.obs["cont1"] = np.random.normal(size=(adata.shape[0],))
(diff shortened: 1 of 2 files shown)


adamgayoso:
Thank you for this! I suppose we aren't properly testing this. Would you be able to add a test? If not we'll get to it eventually.

tests/dataset/test_anndata.py
```diff
@@ -246,6 +246,38 @@ def test_extra_covariates():
...
+
+    transfer_anndata_setup(adata_source=adata, adata_target=bdata)
+
+    # give it a new category
+    del bdata.uns["_scvi"]
+    bdata.obs["cat1"] = 6
+    transfer_anndata_setup(
+        adata_source=adata, adata_target=bdata, extend_categories=True
+    )
+    assert bdata.uns["_scvi"]["extra_categoricals"]["mappings"][-1] == 6
```
adamgayoso, line 278:
```suggestion
    assert bdata.uns["_scvi"]["extra_categoricals"]["mappings"]["cat1"][-1] == 6
```

Issue #1032, opened 2021-04-20
Issue in marginal ll computation


PierreBoyeau:
The current marginal likelihood computes the marginal on the last minibatch (see [here](https://github.com/YosefLab/scvi-tools/blob/6dae6482efa2d235182bf4ad10dbd9483b7d57cd/scvi/model/base/_vaemixin.py#L82)) instead of all the data.

adamgayoso:
Closed by #1033

PR #1033, opened 2021-04-20
fix marginal computation

PierreBoyeau:
#1032


adamgayoso:
Thanks @PierreBoyeau. Can you add a note to the release notes? v0.10? I just checked and this bug was introduced in the 0.9.X releases, so you should mention that there.

PR #1045, opened 2021-05-01
Faster NB LL computation if data is known to be sparse

wukathy:
first try

### DIFF
--- a/scvi/distributions/_negative_binomial.py
+++ b/scvi/distributions/_negative_binomial.py
@@ -98,7 +98,7 @@ def log_nb_positive(
         nonzero_x = x[mask]
         nonzero_mu = mu[mask]
         nonzero_log_theta_mu_eps = log_theta_mu_eps[mask]
-        nonzero_theta, _ , _ = torch.broadcast_tensors(theta, x, x > 0)
+        nonzero_theta, _, _ = torch.broadcast_tensors(theta, x, x > 0)
         terms_for_nonzero = (
             nonzero_x * (torch.log(nonzero_mu + eps) - nonzero_log_theta_mu_eps)
             + torch.lgamma(nonzero_x + theta)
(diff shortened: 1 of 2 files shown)


scvi/distributions/_negative_binomial.py
```diff
@@ -91,14 +93,27 @@ def log_nb_positive(x: torch.Tensor, mu: torch.Tensor, theta: torch.Tensor, eps=
 
     log_theta_mu_eps = torch.log(theta + mu + eps)
 
+    if is_sparse:
+        nonzero_x = x[x > 0]
```
adamgayoso, line 97: compute the mask once, store in a variable

scvi/distributions/_negative_binomial.py
```diff
@@ -91,14 +93,27 @@ def log_nb_positive(x: torch.Tensor, mu: torch.Tensor, theta: torch.Tensor, eps=
...
+        terms_for_nonzero = (
+            theta * (torch.log(theta + eps) - log_theta_mu_eps)
```
adamgayoso, line 99:
```suggestion
```

adamgayoso: this is shared for zero and nonzero

scvi/distributions/_negative_binomial.py
```diff
@@ -91,14 +93,27 @@ def log_nb_positive(x: torch.Tensor, mu: torch.Tensor, theta: torch.Tensor, eps=
...
+            + nonzero_x * (torch.log(mu + eps) - log_theta_mu_eps)
+            + torch.lgamma(nonzero_x + theta)
+            - torch.lgamma(theta)
+            - torch.lgamma(nonzero_x + 1)
```
adamgayoso, line 103: need to mask `mu`, `log_theta_mu_eps` with the nonzero mask.

the term `torch.lgamma(nonzero_x + theta)` might be trickier because theta has not been previously broadcast. Right here its shape is [1, n_genes], and the shape of nonzero_x will be a 1d tensor with all nonzero terms. So maybe theta should be initially broadcast (see Pyro code)

adamgayoso (changes requested):

PR #1055, opened 2021-05-07
model/_scvi.py: Fix passing n_labels to VAE init - closes #1051

### LINKED ISSUE #1051: VAE does not recive n_labels when it is called from model.SCVI
Obvious bug in scvi.model._scvi.py
You should pass 
`n_labels=self.summary_stats["n_labels"],`
when calling init for VAE

cataclysmus:


adamgayoso (approved):
Thanks for the contribution!

PR #1059, opened 2021-05-12
Adding mixin classes for pyro training and posterior sampling

### LINKED ISSUE #1031: Add a Pyro training mixin

vitkl:
This PR adds mixin classes for

1. For pyro training
2. posterior sampling.

It also fixes an issue where pyro model history was erased on load

Fixes #1031


scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,593 @@
...
+        """
+
+        gpus, device = parse_use_gpu_arg(use_gpu)
+
+        self.module.eval()
+
+        train_dl = AnnDataLoader(self.adata, shuffle=False, batch_size=self.batch_size)
+        # sample local parameters
+        i = 0
+        with tqdm(train_dl, desc="Sampling local variables, batch: ") as tqdm_dl:
```
adamgayoso, line 432: we have a custom wrapper for tqdm. See here:

https://github.com/YosefLab/scvi-tools/blob/6dae6482efa2d235182bf4ad10dbd9483b7d57cd/scvi/data/_built_in_data/_download.py#L43-L45

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,593 @@
...
+
+
+class PyroSampleMixin:
+    """
+    This mixin class provides methods for:
+
+    - generating samples from posterior distribution using minibatches and full data
+    """
+
+    def _get_one_posterior_sample(
```
adamgayoso, line 282: torch no grad decorator so you can remove all detach calls.

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,593 @@
...
+
+        Returns
+        -------
+        dictionary {variable_name: [array with samples in 0 dimension]}
+
+        """
+
+        gpus, device = parse_use_gpu_arg(use_gpu)
+
+        self.module.eval()
```
adamgayoso, line 427: eval should be called in the training procedure. So I'd remove this

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,593 @@
...
+                                for i in range(len(samples[k]))
+                            ]
+                        )
+                        for k in samples.keys()
+                    }
+                i += 1
+
+        # sample global parameters
+        i = 0
+        for tensor_dict in train_dl:
```
adamgayoso, line 506: `next(iter(train_dl))`

tests/models/test_pyro.py
```diff
@@ -169,7 +230,7 @@ def test_pyro_bayesian_regression_jit():
     trainer.fit(plan, train_dl)
 
     # 100 features, 1 for sigma, 1 for bias
-    assert list(model.guide.parameters())[0].shape[0] == 102
+    # assert list(model.guide.parameters())[0].shape[0] == 102
```
adamgayoso, line 233: why comment this out?

tests/models/test_pyro.py
```diff
@@ -156,9 +207,19 @@ def test_pyro_bayesian_regression(save_path):
...
+        adata,
+        registry_key="ind_x",
+        adata_attr_name="obs",
+        adata_key_name="_indices",
+    )
     train_dl = AnnDataLoader(adata, shuffle=True, batch_size=128)
     pyro.clear_param_store()
-    model = BayesianRegressionModule(adata.shape[1], 1)
+    model = BayesianRegressionModule(
+        in_features=adata.shape[1], out_features=1, n_obs=adata.n_obs
```
adamgayoso, line 221: the training plan actually sets n_obs automatically (see https://github.com/YosefLab/scvi-tools/blob/6dae6482efa2d235182bf4ad10dbd9483b7d57cd/scvi/train/_trainingplans.py#L602-L606)

tests/models/test_pyro.py
```diff
@@ -156,9 +207,19 @@ def test_pyro_bayesian_regression(save_path):
 def test_pyro_bayesian_regression_jit():
     use_gpu = int(torch.cuda.is_available())
     adata = synthetic_iid()
+    # add index for each cell (provided to pyro plate for correct minibatching)
```
adamgayoso, line 210: we should comment that this isn't strictly necessary for this particular test case model as it only has global latent variables. In other words the old code worked too.

tests/models/test_pyro.py
```diff
@@ -62,20 +47,58 @@ def __init__(self, in_features, out_features):
...
+
+    @staticmethod
+    def _get_fn_args_full_data(adata):
+        x = get_from_registry(adata, _CONSTANTS.X_KEY)
+        if issparse(x):
+            x = np.asarray(x.toarray())
+        x = torch.tensor(x.astype("float32"))
+        ind_x = torch.tensor(get_from_registry(adata, "ind_x"))
+        y = torch.tensor(get_from_registry(adata, _CONSTANTS.LABELS_KEY))
+        return (x, y, ind_x), {}
```
adamgayoso, line 80: will no longer be necessary (see data splitter PR)

tests/models/test_pyro.py
```diff
@@ -62,20 +47,58 @@ def __init__(self, in_features, out_features):
...
+        and the number of dimensions in non-plate axis of each variable"""
+
+        return {
+            "name": "obs_plate",
+            "in": [0],  # index for expression data
+            "sites": {},
+        }
+
+    @staticmethod
+    def _get_fn_args_from_batch(tensor_dict):
```
adamgayoso, line 66: what's the purpose of this in the pyro model as opposed to one level up (as before)?

adamgayoso (changes requested):

scvi/model/base/_pyromixin.py
vitkl: @adamgayoso do you mean at the very end of training? e.i. after this line? https://github.com/YosefLab/scvi-tools/pull/1059/files#diff-cb30789d40b9630e34b2067718c7d61aba3238123aa1a7ae8852c80846031b6fR272

tests/models/test_pyro.py
vitkl: @adamgayoso AutoNormal guide does not have a single parameter - it has one locs and one scales parameter per variable (and I think it has to be used when the model has plates). What are you trying to test with this?

adamgayoso: right there should be 102 variables (for each of loc and scale), 100 for the weight vector, 1 for the sigma from the sample statement and one for the liner bias

tests/models/test_pyro.py
vitkl: I extended this model to (optionally) has cell-specific latent variables - that feature needs to be tested, right?

tests/models/test_pyro.py
vitkl: This test fails with AutoNormal guide probably because these are 3 separate parameters

scvi/model/base/_pyromixin.py
vitkl: what do you mean?

adamgayoso: I think you can use that to get the output of the data loader without having to have this for loop and then the break statement.

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,667 @@
...
+
+        plan_kwargs = plan_kwargs if isinstance(plan_kwargs, dict) else dict()
+
+        data_splitter = DataSplitter(
+            self.adata,
+            train_size=train_size,
+            validation_size=validation_size,
+            batch_size=batch_size,
+            use_gpu=use_gpu,
+        )
```
adamgayoso, line 99: if batch size is none, use the device backed splitter

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,667 @@
...
+        training_plan = PyroTrainingPlan(
+            pyro_module=self.module, n_obs=len(data_splitter.train_idx), **plan_kwargs
+        )
```
adamgayoso, line 102:
```suggestion
        training_plan = PyroTrainingPlan(
            pyro_module=self.module, **plan_kwargs
        )
```

The train runner now sets the property for this automatically

tests/models/test_pyro.py
adamgayoso: I think what this test was doing was to ensure the guide was properly initialized, otherwise you don't get latent var params for each latent var

scvi/model/base/_pyromixin.py
vitkl: `tensor_dict  = next(iter(train_dl))` like this?

tests/models/test_pyro.py
vitkl: Not sure I understand what you mean. Do you mean it tests that the number and shape of all latent variable corresponds to the number and shape of guide param? If yes, the code I added in replacement tests that for weight vector: https://github.com/YosefLab/scvi-tools/pull/1059/files#diff-663073fe4ef3cfa8f3fad68a502f760ee33dbb94c3c803c50635f0d63f67ee30R278-R280

tests/models/test_pyro.py
vitkl: You mean what is the difference between `list_obs_plate_vars(self)` and `create_plates(self, x, y, ind_x)`?

The latter creates a pyro plate, the former records the name of the minibatch plate such that posterior sampling knows which is the minibatch plate (`"name": "obs_plate"`). Posterior sampling also needs to know which variables belong to that plate (keys of `"sites"` dictionary). The dimensions in `"sites"` dictionary and `"in"` are needed for the automatic autoencoding guide - but not strictly necessary for this PR.

WDYT?

scvi/train/_trainingplans.py
```diff
@@ -600,13 +600,27 @@ def __init__(
...
         optim: Optional[pyro.optim.PyroOptim] = None,
+        n_obs: Optional[int] = None,
+        optim_kwargs: Optional[dict] = None,
     ):
         super().__init__()
         self.module = pyro_module
         self._n_obs_training = None
 
+        optim_kwargs = optim_kwargs if isinstance(optim_kwargs, dict) else dict()
+        optim_kwargs["lr"] = (
```
vitkl, line 611: @adamgayoso I added an option to change lr. It would be inconvenient to write a custom training method just to change lr. WDYT?

vitkl:
Interestingly, the last test works but cell2location with full data and non-amortised inference gives an error which I do not quite understand. See HTML attached (can be reproduced by taking this notebook to colab).
[cell2location_analysis_3_scvi_ln_full_data.html.zip](https://github.com/YosefLab/scvi-tools/files/6490759/cell2location_analysis_3_scvi_ln_full_data.html.zip)

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+        trainer_kwargs[es] = (
+            early_stopping if es not in trainer_kwargs.keys() else trainer_kwargs[es]
+        )
+        runner = TrainRunner(
+            self,
+            training_plan=training_plan,
+            data_splitter=data_splitter,
+            max_epochs=max_epochs,
+            use_gpu=use_gpu,
+            **trainer_kwargs,
```
adamgayoso, line 119: we should be able to detect if JIT trace elbo is being used here through the plan kwargs and add the callback accordingly.

I think if you look at the peakvi train method some callbacks are added as an example

https://github.com/YosefLab/scvi-tools/blob/a9da914764d311ff18b4b557b78d9550a3b940af/scvi/model/_peakvi.py#L216-L221

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+        use_gpu
+            Use default GPU if available (if None or True), or index of GPU to use (if int),
+            or name of GPU (if str, e.g., `'cuda:0'`), or use CPU (if False).
+        train_size
+            Size of training set in the range [0.0, 1.0].
+        validation_size
+            Size of the test set. If `None`, defaults to 1 - `train_size`. If
+            `train_size + validation_size < 1`, the remaining cells belong to a test set.
+        batch_size
+            Minibatch size to use during training.
```
adamgayoso, line 74:
```suggestion
            Minibatch size to use during training. If `None`, no minibatching occurs and all data is copied to device (e.g., GPU).
```

scvi/model/base/_pyromixin.py
adamgayoso, line 119: we should automatically add `early_stopping_monitor=train_loss_epoch` (or whatever it gets logged as) to the trainer kwargs, so the early stopping will work by default if user selects True.

Also, should we change the logged name to elbo?

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+
+        self.module.to(device)
+
+        return samples
+
+    def sample_posterior(
+        self,
+        num_samples: int = 1000,
+        return_sites: Optional[list] = None,
+        use_gpu: bool = False,
```
adamgayoso, line 356: why False? it could be None and then we automatically infer it

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+        batch_size: int = 128,
+        sample_kwargs=None,
+        return_samples: bool = False,
+    ):
+        """
+        Generate samples from posterior distribution for each parameter
```
adamgayoso, line 362: for these docstrings there should be one line description with period and then a line break and an empty line and then you can have a paragraph

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+        and compute mean, 5%/95% quantiles, standard deviation.
+
+        Parameters
+        ----------
+        num_samples
+            number of posterior samples to generate.
+        return_sites
+            get samples for pyro model variable, default is all variables, otherwise list variable names).
+        use_gpu
+            Use gpu?
```
adamgayoso, line 372: use same language as other functions.

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+        sample_kwargs
+            dictionary with arguments to _get_posterior_samples (see below):
+            return_observed
+                return observed sites/variables?
```
adamgayoso, line 376: not sure this is proper formatting (you can see numpydoc to check)

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+
+        Returns
+        -------
+        Posterior distribution samples, a dictionary with elements as follows,
+         containing dictionaries of numpy arrays for each variable:
+            post_sample_means - mean of the distribution for each variable;
+            post_sample_q05 - 5% quantile;
+            post_sample_q95 - 95% quantile;
+            post_sample_sds - standard deviation
+            posterior_samples - samples for each variable as numpy arrays of shape `(n_samples, ...)` (Optional)
```
adamgayoso, line 388: not sure this is proper formatting. You can see the docs build as one of the PR checks; though, you need to add these classes to the docs.

scvi/train/_trainingplans.py
```diff
@@ -600,13 +600,27 @@ def __init__(
...
+            optim_kwargs["lr"] if "lr" in list(optim_kwargs.keys()) else 1e-3
```
adamgayoso, line 612:
```suggestion
            optim_kwargs["lr"] if "lr" in optim_kwargs.keys() else 1e-3
```

do you need to make a list?

scvi/train/_trainingplans.py
adamgayoso: seems reasonable to me.

scvi/train/_trainingplans.py
adamgayoso: also I think you can just do `optim_kwargs.update({"lr": 1e-3})`

scvi/train/_trainingplans.py
```diff
@@ -600,13 +600,27 @@ def __init__(
...
+        )
+        optim_kwargs["clip_norm"] = (
+            optim_kwargs["clip_norm"]
+            if "clip_norm" in list(optim_kwargs.keys())
+            else 200
+        )
```
adamgayoso, line 618: same use of `.update()` here.

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+            results["posterior_samples"] = samples
+
+        results["post_sample_means"] = {v: samples[v].mean(axis=0) for v in param_names}
+        results["post_sample_q05"] = {
+            v: np.quantile(samples[v], 0.05, axis=0) for v in param_names
+        }
+        results["post_sample_q95"] = {
+            v: np.quantile(samples[v], 0.95, axis=0) for v in param_names
+        }
+        results["post_sample_sds"] = {v: samples[v].std(axis=0) for v in param_names}
```
adamgayoso, line 413: it seems like user can supply a dict in the form `{"means": np.mean, "std": np.std}` so users can get different summary stats? and if none is supplied you give this?

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+
+class PyroJitGuideWarmup(Callback):
+    def __init__(self, train_dl) -> None:
+        super().__init__()
+        self.dl = train_dl
+
+    def on_train_start(self, trainer, pl_module):
+        """
+        Way to warmup Pyro Guide in an automated way.
+        Also device agnostic.
```
adamgayoso, line 27: needs to be empty line above this one
```suggestion

        Also device agnostic.
```

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+            if site["type"] == "sample"
+            if any(
+                f.name == self.module.model.list_obs_plate_vars()["name"]
+                for f in site["cond_indep_stack"]
+            )
+        }
+        return obs_plate
+
+    def _posterior_samples_minibatch(
+        self, use_gpu: bool = True, batch_size: int = 128, **sample_kwargs
```
adamgayoso, line 262: use_gpu should be `None` everywhere

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+        Generate samples of the posterior distribution of each parameter, separating local (minibatch) variables
+        and global variables, which is necessary when performing minibatch inference.
+
+        Note for developers: requires model class method which lists observation/minibatch plate
+        variables (self.module.model.list_obs_plate_vars()).
+
+        Parameters
+        ----------
+        use_gpu
+            Bool, use gpu?
```
adamgayoso, line 274: use common use_gpu language

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+        self, use_gpu: bool = True, batch_size: int = 128, **sample_kwargs
+    ):
+        """
+        Generate samples of the posterior distribution of each parameter, separating local (minibatch) variables
+        and global variables, which is necessary when performing minibatch inference.
+
+        Note for developers: requires model class method which lists observation/minibatch plate
+        variables (self.module.model.list_obs_plate_vars()).
+
+        Parameters
```
adamgayoso, line 271: not fully documented, will be helpful to have

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+        return obs_plate
+
+    def _posterior_samples_minibatch(
+        self, use_gpu: bool = True, batch_size: int = 128, **sample_kwargs
+    ):
+        """
+        Generate samples of the posterior distribution of each parameter, separating local (minibatch) variables
+        and global variables, which is necessary when performing minibatch inference.
+
+        Note for developers: requires model class method which lists observation/minibatch plate
```
adamgayoso, line 268: numpydoc has a notes section that I believe goes after Returns

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,415 @@
...
+                                len(samples[k])
+                            )  # for each sample (in 0 dimension
+                        ]
+                    )
+                    for k in samples.keys()  # for each variable
+                }
+            i += 1
+
+        # sample global parameters
+        tensor_dict = next(iter(train_dl))
```
adamgayoso, line 332: for full batch, this will copy yet again to GPU, can't you `tensor_dict` without redfining it because it's defined in the previous loop?

adamgayoso: (and subsequently, the already defined args and kwargs in scope)

adamgayoso (changes requested):

scvi/model/base/_pyromixin.py
vitkl: @adamgayoso where is validation loss logged?

Yes, I think would be good to rename that to "train_elbo_epoch"/"validation_elbo_epoch"

adamgayoso: well we don't actually do any validation loop with Pyro, I wasn't sure how to exactly in a universal way. Is it just `evaulate_loss`? And how does that work for regular black box VI? if there are no variational parameters for the val set?

vitkl: I think you can just save ELBO loss like this:
```python
validation_elbo = svi.evaluate_loss(*validation_args, **validation_kwargs)
```

tests/models/test_pyro.py
adamgayoso: no I mean like why do you need this specific static method where the comment is?

tests/models/test_pyro.py
```diff
@@ -62,20 +48,64 @@ def __init__(self, in_features, out_features):
...
+    def __init__(self, **kwargs):
 
         super().__init__()
-        self._model = BayesianRegressionPyroModel(in_features, out_features)
-        self._guide = AutoDiagonalNormal(self.model)
+        self._model = BayesianRegressionPyroModel(**kwargs)
+        self._guide = AutoNormal(
+            self.model, init_loc_fn=init_to_mean, create_plates=self.model.create_plates
+        )
+        self._get_fn_args_from_batch = self._model._get_fn_args_from_batch
```
adamgayoso, line 108: going back to previous point... why is this necessary?, why not define the static method here instead?

tests/models/test_pyro.py
adamgayoso: on a different note, should `list_obs_plate_vars(self)` be one level up in the `PyroBaseModuleClass` and just do like

```
def list_obs_plate_vars(self):
    pass
```

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,416 @@
...
+        args, kwargs = self.module._get_fn_args_from_batch(tensor_dict)
+        args = [a.to(device) for a in args]
+        kwargs = {k: v.to(device) for k, v in kwargs.items()}
+        self.to_device(device)
+
+        global_samples = self._get_posterior_samples(args, kwargs, **sample_kwargs)
+        global_samples = {
+            k: v
+            for k, v in global_samples.items()
+            if k not in list(self.module.model.list_obs_plate_vars()["sites"].keys())
```
adamgayoso, line 343: wrote in a different comment, but this could very well be in self.module

vitkl: I think it is easier to keep track of plate-related things in pyro model class. `self.module` can often be general for many similar models.

scvi/model/base/_pyromixin.py
vitkl: @adamgayoso Would it hurt to always to this callback? (see what I added in https://github.com/YosefLab/scvi-tools/pull/1059/commits/cb3cae91fc6bab18b67b193520decbc8f526aeb6)

vitkl: Actually the callback does not work. I probably do not understand how to get dataloader from DataSplitter.

```python
========================================================================= short test summary info ==========================================================================
FAILED tests/models/test_pyro.py::test_pyro_bayesian_train_sample_mixin - AttributeError: 'DataSplitter' object has no attribute 'train_idx'
FAILED tests/models/test_pyro.py::test_pyro_bayesian_train_sample_mixin_full_data - AttributeError: 'DeviceBackedDataSplitter' object has no attribute 'train_tensor_dict'
FAILED tests/models/test_pyro.py::test_pyro_bayesian_train_sample_mixin_with_local - AttributeError: 'DataSplitter' object has no attribute 'train_idx'
FAILED tests/models/test_pyro.py::test_pyro_bayesian_train_sample_mixin_with_local_full_data - AttributeError: 'DeviceBackedDataSplitter' object has no attribute 'train_...
```

adamgayoso: yeah this is fine, just need to add `datasplitter.setup()` before trying to get the train data loadeer. Subsequent calls to setup are a no-op so this is fine.

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,423 @@
...
+        es = "early_stopping"
+        trainer_kwargs[es] = (
+            early_stopping if es not in trainer_kwargs.keys() else trainer_kwargs[es]
+        )
+
+        # if "callbacks" not in trainer_kwargs.keys():
+        #    trainer_kwargs["callbacks"] = []
+        # trainer_kwargs["callbacks"].append(
+        #    PyroJitGuideWarmup(data_splitter.train_dataloader())
+        # )
```
adamgayoso, line 119: uncomment, run `data_splitter.setup()` before this block of code

tests/models/test_pyro.py
vitkl: @adamgayoso what do you think about this solution?
https://github.com/YosefLab/scvi-tools/pull/1059/commits/08dbe1f8c27073838343c79e70635ac3048f3f38

vitkl:
Accidentally merged pyro-cell2location and reverted that.

tests/models/test_pyro.py
vitkl: This is necessary to add even in those tests because the model now has ind_x argument

tests/models/test_pyro.py
vitkl: We discussed this on slack - because it is easier to keep model-specific code in pyro model class - that way scVI module class can be reused for multiple similar models.

scvi/model/base/_pyromixin.py
vitkl: @adamgayoso could you please point where these classes should be added to docs? I don't seem to be able to find it.

scvi/train/_trainingplans.py
```diff
@@ -593,20 +593,31 @@ class PyroTrainingPlan(pl.LightningModule):
...
         super().__init__()
         self.module = pyro_module
         self._n_obs_training = None
 
+        optim_kwargs = optim_kwargs if isinstance(optim_kwargs, dict) else dict()
+        if "lr" not in optim_kwargs.keys():
+            optim_kwargs.update({"lr": 1e-3})
+        if "clip_norm" not in optim_kwargs.keys():
+            optim_kwargs.update({"clip_norm": 200})
+
```
adamgayoso, line 616: what if user passes in an optim that doesn't take clip norm? shouldn't this check first if lr and clip norm are in the signature of the optim init?

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,472 @@
...
+            Load model on default GPU if available (if None or True),
+            or index of GPU to use (if int), or name of GPU (if str), or use CPU (if False).
+
+        Returns
+        -------
+        dictionary {variable_name: [array with samples in 0 dimension]}
+
+        Notes
+        -----
+        Note for developers: requires scVI module property (a dictionary, self.module.list_obs_plate_vars)
```
adamgayoso, line 318:
```suggestion
        Note for developers: requires overwritten :func:`~scvi.module.base.PyroBaseModuleClass.list_obs_plate_vars` method.
```

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,472 @@
...
+        which lists observation/minibatch plate name and variables.
+        See PyroBaseModuleClass.list_obs_plate_vars for details of the variables it should contain.
+        This dictionary can be returned by model class method self.module.model.list_obs_plate_vars()
```
adamgayoso, line 321:
```suggestion
        This dictionary can be returned by model class method `self.module.model.list_obs_plate_vars()`
```

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,472 @@
...
+        to keep all model-specific variables in one place.
+
```
adamgayoso, line 323:
```suggestion
```

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,472 @@
...
+        self.module.to(device)
+
+        return samples
+
+    def sample_posterior(
+        self,
+        num_samples: int = 1000,
+        return_sites: Optional[list] = None,
+        use_gpu: bool = None,
+        batch_size: int = 128,
```
adamgayoso, line 404: let's use batch_size like here:

https://github.com/YosefLab/scvi-tools/blob/9fcda8f8c3bea9007e06403fa066607d41c6aad6/scvi/model/base/_vaemixin.py#L22

Default is None, and pulls from our global settings config.

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,472 @@
...
+        sample_kwargs=None,
+        return_samples: bool = False,
+        summary_fun: Optional[dict] = None,
+    ):
```
adamgayoso, line 408: let's add the return type here

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,472 @@
...
+        Parameters
+        ----------
+        num_samples
+            number of posterior samples to generate.
+        return_site
+            get samples for pyro model variable, default is all variables, otherwise list variable names).
+        use_gpu
+            Load model on default GPU if available (if None or True),
+            or index of GPU to use (if int), or name of GPU (if str), or use CPU (if False).
+        sample_kwargs
```
adamgayoso, line 424: instead of sample_kwargs, can we just enumerate the parameters here? Users won't go looking at a private method and it doesn't seem like many params?

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,472 @@
...
+
+    def sample_posterior(
+        self,
+        num_samples: int = 1000,
+        return_sites: Optional[list] = None,
+        use_gpu: bool = None,
+        batch_size: int = 128,
+        sample_kwargs=None,
+        return_samples: bool = False,
+        summary_fun: Optional[dict] = None,
```
adamgayoso, line 407:
```suggestion
        summary_fun: Optional[Dict[str, Callable]] = None,
```
you might need to import Dict and Callable from typing

scvi/module/base/_base_module.py
```diff
@@ -261,6 +261,21 @@ def model(self):
     def guide(self):
         pass
 
+    @property
+    def list_obs_plate_vars(self):
```
adamgayoso, line 265: I think this should be fine

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,472 @@
...
+        ----------
+        args
+            Arguments to the model.
+        kwargs
+            Keyword arguments to the model.
+
+        Returns
+        -------
+        Dictionary with keys corresponding to site names and values to plate dimension.
+
```
adamgayoso, line 281:
```suggestion
```
No empty space in all docstrings

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,472 @@
...
+        """
+
```
adamgayoso, line 283:
```suggestion
```
Same here, no empty lines

scvi/model/base/_pyromixin.py
```diff
@@ -0,0 +1,472 @@
...
+        Posterior distribution samples, a dictionary with elements as follows,
+         containing dictionaries of numpy arrays for each variable:
+            post_sample_means - mean of the distribution for each variable;
+            post_sample_q05 - 5% quantile;
+            post_sample_q95 - 95% quantile;
+            post_sample_sds - standard deviation
+            posterior_samples - samples for each variable as numpy arrays of shape `(n_samples, ...)` (Optional)
+
+        """
+
```
adamgayoso, line 445: no empty lines again

adamgayoso (changes requested):

vitkl:
Sorry for the accidental merge again, I don't get why this merge happened. I am trying to merge pyro-mixin into pyro-cell2location branch using GitHub pull request web interface. For some reason, that seems to create a merge commit from pyro-cell2location into pyro-mixin.

scvi/train/_trainingplans.py
vitkl: optim_kwargs are only used for default optimiser.

scvi/model/base/_pyromixin.py
vitkl: what are you referring to?

vitkl:
@adamgayoso after talking to @martinjankowiak I think the default optimiser should be Adam, not ClippedAdam. I can change that. Should I modify this branch/PR or create a new one?

PR #1061, opened 2021-05-12
use data modules for splitters

### LINKED ISSUE #1050: Use `LightningDataModule` instead of `DataSplitter`
We inadvertently implemented a replica of the [`LightningDataModule`](https://pytorch-lightning.readthedocs.io/en/stable/extensions/datamodules.html). We should consider refactoring by inheriting this class.

### LINKED ISSUE #1058: dataloader which loads all data to GPU once during training
For small datasets and for benchmarking purposes would be good to have the option to do training on full data rather than in minibatches. Current dataloader copies data to GPU in every epoch. Would be good to have a dataloader that can load data to GPU once during training.

adamgayoso:
Fixes #1050 
Fixes #1058 

This also adds a device-backed data splitter, which will be useful for training pyro models on a GPU using SVI.

However, it creates breaking changes for devs

1. No need to call the data splitter anymore (though I assume most have been using our mixin)
2. No need to give the training plans the number of training observations, now we can set it using the property.

### DIFF
--- a/scvi/model/_scanvi.py
+++ b/scvi/model/_scanvi.py
@@ -366,7 +366,6 @@ class SCANVI(RNASeqMixin, VAEMixin, ArchesMixin, BaseModelClass):
             batch_size=batch_size,
             use_gpu=use_gpu,
         )
-        data_splitter.setup()
         training_plan = SemiSupervisedTrainingPlan(self.module, **plan_kwargs)
         if "callbacks" in trainer_kwargs.keys():
             trainer_kwargs["callbacks"].concatenate(sampler_callback)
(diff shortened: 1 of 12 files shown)


scvi/model/_scanvi.py
```diff
@@ -366,6 +366,7 @@ def train(
             batch_size=batch_size,
             use_gpu=use_gpu,
         )
+        data_splitter.setup()
```
galenxing, line 369: Do we need to have these calls if setup() is already called in TrainRunner?

adamgayoso: It doesn't do anything after the first time it's called (is passed over), and it's more necessary to call here as it sets the `train_idx` attribute.

galenxing (approved):
lgtm!

PR #1064, opened 2021-05-15
ignore configure_optimizers warning

### LINKED ISSUE #1057: Weird message when training pyro model `this fit will run with no optimizer`
This warning is not expected. Would be good if you can catch this:
```
lib/python3.7/site-packages/pytorch_lightning/utilities/distributed.py:52: UserWarning: `LightningModule.configure_optimizers` returned `None`, this fit will run with no optimizer
  warnings.warn(*args, **kwargs)
```

mjayasur:
closes #1057

### DIFF
--- a/scvi/train/_trainer.py
+++ b/scvi/train/_trainer.py
@@ -7,9 +7,9 @@ import pytorch_lightning as pl
 from pytorch_lightning.callbacks.early_stopping import EarlyStopping
 from pytorch_lightning.loggers import LightningLoggerBase
 
-import scvi.train
 from scvi import settings
 from scvi._compat import Literal
+from scvi.train import PyroTrainingPlan
 
 from ._logger import SimpleLogger
 from ._progress import ProgressBar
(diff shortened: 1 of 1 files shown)


wukathy (approved):
lgtm

scvi/train/_trainer.py
```diff
@@ -156,4 +156,9 @@ def fit(self, *args, **kwargs):
                 category=UserWarning,
                 message="One of given dataloaders is None and it will be skipped",
             )
+            warnings.filterwarnings(
```
adamgayoso, line 159: let's only filter it if the model passed to fit is an instance of:

https://github.com/YosefLab/scvi-tools/blob/a9da914764d311ff18b4b557b78d9550a3b940af/scvi/train/_trainingplans.py#L581

See the pylit API here, model should be the first arg.

https://pytorch-lightning.readthedocs.io/en/latest/common/trainer.html#fit

adamgayoso (changes requested):

scvi/train/_trainer.py
```diff
@@ -7,6 +7,7 @@
 from pytorch_lightning.callbacks.early_stopping import EarlyStopping
 from pytorch_lightning.loggers import LightningLoggerBase
 
+import scvi.train
```
adamgayoso, line 10: import needs to be fixed :)

adamgayoso (changes requested):

PR #1066, opened 2021-05-18
Adding ability to change doublet ratios

njbernstein:
Also changing soft argument to have correct functionality

### DIFF
--- a/scvi/external/solo/_model.py
+++ b/scvi/external/solo/_model.py
@@ -141,8 +141,9 @@ class SOLO(BaseModelClass):
             batch_indices = None
 
         # anndata with only generated doublets
-        doublet_adata = cls.create_doublets(orig_adata, indices=batch_indices,
-                                            doublet_ratio=doublet_ratio)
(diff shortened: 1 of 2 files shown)


scvi/external/solo/_model.py
```diff
@@ -96,6 +96,7 @@ def from_scvi_model(
         scvi_model: SCVI,
         adata: Optional[AnnData] = None,
         restrict_to_batch: Optional[str] = None,
+        doublet_ratio: int = 2,
```
adamgayoso, line 99: need documentation for this parameter

adamgayoso (changes requested):
thanks! small change needed. Also could you add a note to the release notes?

https://github.com/YosefLab/scvi-tools/blob/master/docs/release_notes/v0.11.0.rst

scvi/external/solo/_model.py
njbernstein: Done

adamgayoso (approved):

PR #1067, opened 2021-05-19
[WIP] lvm-DE code

### LINKED ISSUE #1068: lvm-DE

PierreBoyeau:
Fixes #1068 
Here is the code we discussed for lvm-DE, an importance-sampling-based procedure for calibrated DE gene predictions.
A parallel PR will be opened asap for the associated notebook introducing the module.
Depending on your call on the place of this tool in `scvi-tools`, it could require some additional work (see **questions**).

# Presentation of the PR
## New files
- `external/wscvi`: In line with existing external packages, the `wscvi` folder contains the code to run an SCVI version that easily yields variational distributions, priors densities, as well as likelihoods.
Having access to these quantities is used for two reasons.
First, lvm-DE relies on importance-sampling, and hence directly requires these quantities.
A second reason is that this allows training scVI models with IWELBO which arguably improves the stability of lvm-DE.

- `model/base/_demixin.py` contains the `lvm-DE` core. 
I tried as much as possible to document it, but it requires additional documentation.
In particular, the `lvm_de` method applies the same mechanisms as `differential_expression` in `_rnamixin.py`. It also uses `DifferentialComputation` under the hood.

## Modified files
The main modification I made was to make `DifferentialComputation` usable with lvm-DE.
To get a given number of gene normalized expression samples for a selection of indices, `DifferentialComputation` 
randomly samples cell indices, [using](https://github.com/YosefLab/scvi-tools/blob/a9da914764d311ff18b4b557b78d9550a3b940af/scvi/utils/_differential.py#L424)
```
    idx = np.random.choice(np.arange(self.adata.shape[0])[selection], n_samples)
```
Next, we get one sample for each index using the variational distribution directly.
This reasoning does not work well for lvm-DE.

Consequently, I modified `DifferentialComputation` such that it lets `model_fn` take care of the sampling instead, through `n_samples_overall`.
For instance, when using the current differential tools,  `model_fn` now takes a `n_samples_overall` parameter, which takes care of calling `np.random.choice` the same way as before.
While this is a small change, it required me to change each so-mentioned function in the codebase (in `_peakvi.py`, `_scvi.py`, and `_totalvi.py`).
I also induced code duplication, which I am not fond of but will try to remove.

I also allowed the `get_marginal_ll` methods to return observation-specific marginal densities as an option, as lvm-DE requires these densities.

# Questions
Here are some uncertainties I have on what we should do with this code.

1. **Should we somehow "merge" WSCVI and SCVI?**
In other words, should we provide an option `n_particles` to SCVI to switch from a mode computing KL divergences to a one relying on log-ratios to train the model?
An easy first step in this direction consists in returning `torch.Distribution` in `inference` and `generative`.
From there, we have two options.
First, we could let the `loss` function
decide how to compute the loss (using `kl_divergence` when possible and densities otherwise).
In that scenario, we would have a somehow unified `VAE`/`WVAE` model.
This could however make the module heavier, and I don't know if this is a good idea.
The second option, more in phase with what `WVAE` does, would be to keep the log-ratio-friendly model on the side.
In any case, I think that a `WVAE` inspired structure can help developers in certain ways that go beyond DE.
Indeed, it can easily be changed to consider alternatives inference networks (e.g., using flows) and priors for `z` (hyperspherical, vamp priors, etc.), or maybe more easily consider hierarchical models.

2. **Where should we keep WSCVI?**
I think the way to go is to keep it as an external package, but what do you think?


scvi/model/base/_vaemixin.py
```diff
@@ -126,6 +149,7 @@ def get_latent_representation(
         give_mean: bool = True,
         mc_samples: int = 5000,
         batch_size: Optional[int] = None,
+        _adata=None,
```
adamgayoso, line 152: what is the purpose of this?

scvi/model/base/_vaemixin.py
```diff
@@ -50,8 +50,11 @@ def get_marginal_ll(
         adata: Optional[AnnData] = None,
         indices: Optional[Sequence[int]] = None,
         n_mc_samples: int = 1000,
+        n_samples_per_pass: int = 25,
         batch_size: Optional[int] = None,
-    ) -> float:
+        observation_specific: Optional[bool] = False,
+        _adata: Optional[AnnData] = None,
```
adamgayoso, line 56: what is the purpose of this?

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,453 @@
...
+    Mixin for using
+    importance-weighted DE content.
+    This however requires some additional structure on the
+    module's (e.g., VAE) methods and associate signatures
+    """
+
+    @_doc_params(
+        doc_differential_expression=doc_differential_expression,
+    )
+    def lvm_de(
```
adamgayoso, line 39: can we somehow call this `differential_expression` and add a mode to it for importance weighted estimator and then remove the DE function out of `RNAMixin`?

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,453 @@
...
+
+    @torch.no_grad()
+    def _inference_loop(
+        self,
+        adata: AnnData,
+        indices: Sequence,
+        n_samples: int,
+        marginal_n_samples_per_pass=500,
+        n_mc_samples_px=5000,
+        batch_size=64,
```
adamgayoso, line 241: typing

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,453 @@
...
+        return_numpy : Optional[bool], optional
+             Whether numpy should be returned, by default False
+        marginal_n_samples_per_pass : int, optional
+            Number of samples per pass to compute the marginal likelihood, by default 500
+        n_mc_samples_px : int, optional
+            Number of overall samples per cell used to compute the marginal likelihood, by default 5000
+        n_cells_per_chunk : Optional[int], optional
+            Number of cells to use in each minibatch, by default 500
+        max_chunks : Optional[int], optional
+            Maximum number of chunks to use, by default None
```
adamgayoso, line 160: no need to put the type in docstrings, sphinx does this for us.

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,453 @@
...
+            silent,
+            pseudocounts=pseudocounts,
+            use_permutation=True,
+            **kwargs,
+        )
+
+        return result
+
+    @torch.no_grad()
+    def get_population_expression(
```
adamgayoso, line 108: is this the best name for it? Is this something you expect users to call independently of DE?

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,453 @@
...
+        self,
+        adata: Optional[AnnData] = None,
+        indices: Optional[Sequence[int]] = None,
+        n_samples: int = 25,
+        n_samples_overall: int = None,
+        batch_size: Optional[int] = 64,
+        do_filter_cells: bool = True,
```
adamgayoso, line 115:
```suggestion
        filter_cells: bool = True,
```

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,453 @@
...
+        # Step 2.
+        # Determine number of cell chunks
+        # Because of the quadratic complexity we split cells in smaller chunks when
+        # the cell population is too big
+        # This ensures that the method remains scalable when looking at very large cell populations
+        n_cells = indices_.shape[0]
+        logger.debug("n cells {}".format(n_cells))
+        logger.debug(indices_)
+        n_cell_chunks = int(np.ceil(n_cells / n_cells_per_chunk))
+        np.random.shuffle(indices_)
```
adamgayoso, line 201: should there be a random state?

scvi/external/wscvi/_module.py
```diff
@@ -0,0 +1,261 @@
...
+            px_r=px_r,
+            px_rate=px_rate,
+            px_dropout=px_dropout,
+            log_pl=log_pl,
+            log_px_latents=log_px_latents,
+            log_pz=log_pz,
+            log_pjoint=log_pjoint,
+        )
+
+    def generative_evaluate(self, tensors, inference_outputs):
```
adamgayoso, line 230: a simple docstring to motivate this function

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,453 @@
...
+        hs = []
+        log_px_zs = []
+        for tensors in scdl:
+            inference_outputs, generative_outputs, = self.module.forward(
+                tensors,
+                inference_kwargs=dict(n_samples=n_samples, return_densities=True),
+                compute_loss=False,
+            )
+            z = inference_outputs["z"].reshape(-1, self.module.n_latent).cpu()
+            h = generative_outputs["px_scale"]
```
adamgayoso, line 293: should we make px_scale an argument to this function so people can pick different h?

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,453 @@
...
+        importance_weight = torch.logsumexp(
+            log_pz.unsqueeze(1)
+            + log_px_zs
+            - log_px
+            - torch.logsumexp(log_qz, 1, keepdims=True),
+            dim=1,
+        )
+        # importance_weight = log_pz + torch.logsumexp(
+        #     log_px_zs - log_px - torch.logsumexp(log_qz, 1, keepdims=True), -1
+        # )
```
adamgayoso, line 337: remove commented code

adamgayoso (changes requested):

scvi/model/base/_vaemixin.py
PierreBoyeau: The problem is that `_inference_loop`, in `_demixin.py` calls `get_marginal_ll` a lot. Unfortunately, this method needs to check the validity of the data at each call, by calling `_validate_anndata`.
`_validate_anndata` takes approximately 1sec to run, which is fine when it is executed once, but this is more of a problem when you call `get_marginal_ll` (or `get_latent_representation`) many times

PierreBoyeau: `_adata` bypasses this behavior by assuming that the anndata has already been checked.
One clarification: I added this to `get_latent_representation` because this function can still be called a number of times when you do 1vsAll tests for instance

scvi/model/base/_vaemixin.py
PierreBoyeau: see above for `_adata`
The option `observation_specific` allows to return observation-specific marginal estimates, i.e., estimates of p(x_i), instead of mean marginal estimates

scvi/model/base/_vaemixin.py
PierreBoyeau: I don't really know if I could do something cleaner, what do you think?

scvi/model/base/_demixin.py
PierreBoyeau: I see, the idea would be to move `get_normalized_expression` and `differential_expression` from `_rnamixin.py`, and compartmentalize every feature for differential expression (plugin and IS alike) in `_demixin.py`?
I am getting this right?

scvi/model/base/_vaemixin.py
adamgayoso: Ok so we should consider either a boolean argument (validate_anndata: bool = True) or maybe a private field in the anndata setup dictionary? what do you think @galenxing

adamgayoso: private field would be like `adata.uns["_scvi"]["_requires_validation"] = False`

scvi/model/base/_demixin.py
adamgayoso: Yes, but keep `get_normalized_expression` in RNA mixin, just move all DE specific functionality into this mixin. Though I see now it can get messy since `differential_expression` assumes `get_normalized_expression` exists...

adamgayoso: do you think there's a way around that?

scvi/model/base/_vaemixin.py
PierreBoyeau: Went for the latter option!

PierreBoyeau: See 1473719

scvi/model/base/_demixin.py
PierreBoyeau: @adamgayoso  I found a solution that I find pretty attractive 5f5ac98.
However, the new signature of `differential_expression` is even more intricate than it was before.
I am afraid that the `differential_expression` function is becoming a bit too heavy.
I think that now (or in a following up PR) is a good time to think about ways to facilitate and clarify DE queries within `scvi-tools`.
Do you agree?

Here are some random ideas that come to mind, I don't really know if they are good ones:
1. **Splitting up how DE is performed in two steps**. 
I think that performing a significant number of subpopulation comparisons within an experiment is a common use case. Unfortunately, the current codebase requires to specify which test is applied each time `differential_expression` is called, which is pretty cumbersome in my opinion.
For this reason, an option could be to split up `differential_expression` arguments into two categories: the ones that define which test to apply, and the ones that characterize which subpopulations are compared.


We would expect the arguments of the first category to be applied only once, before any DE query, but we would obviously let users change those parameters at any point.
In the first step, the user would explicit.

```python
    model = SCVI(adata=adata)
    model.train()

    # Step 1: specify which test is applied
    ## either 
    model.setup_differential_expression()  # default mode
    ## or
    model.setup_differential_expression(importance_sampling=True, delta=None, fdr_target=0.1)
    
    # Step 2: performing interactive DE queries
    results_1 = model.differential_expression(
        idx1="(labels == 'label_1') & (_scvi_batch == 1)",
        idx1="(labels == 'label_1') & (_scvi_batch == 0)",
        batch_correction=False,
    )
    results_2 = model.differential_expression(
        idx1="(labels == 'label_1')",
        idx1="(labels == 'label_2')",
        batch_correction=False,
    )
```

2. **Visualizing which test and which subpopulations are compared**
In a similar fashion than `setup_anndata` within a notebook, `setup_differential_expression` could display which test to perform, clarifying what is happening under the hood.
We could also do a similar thing when `differential_expression` is called, but I am not sure this is as important

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,473 @@
...
+            point_library = (
+                self.module.inference(**_inf_inputs, return_densities=True)[lib_key]
+                .squeeze(0)
+                .expand(_n_samples_loop, _n_cells, 1)
+            )
+
+            inference_outputs["z"] = _z_reshaped
+            inference_outputs["library"] = point_library
+            _log_px_zs.append(
+                self.module.generative_evaluate(
```
adamgayoso, line 426: just checking that we are testing importance sampling DE with scvi base model?

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,473 @@
...
+            Number of cells to use in each minibatch, by default 500
+        max_chunks
+            Maximum number of chunks to use, by default None
+        normalized_expression_key
+            Key associated to the normalized expression level in the model
+
+        Returns
+        -------
+        res
+            Importance weighted expression levels of shaoe (Number of samples, number of genes)
```
adamgayoso, line 183: shape?

scvi/model/base/_demixin.py
adamgayoso: `get_subpopulation_expression`

scvi/model/base/_demixin.py
```diff
@@ -0,0 +1,473 @@
...
+        transform_batch: Optional[Sequence[Union[Number, str]]] = None,
+        return_numpy: Optional[bool] = False,
+        marginal_n_samples_per_pass: int = 500,
+        n_mc_samples_px: int = 5000,
+        n_cells_per_chunk: Optional[int] = 500,
+        max_chunks: Optional[int] = None,
+        normalized_expression_key: str = "px_scale",
+    ) -> np.ndarray:
+        """
+        Computes importance-weighted expression levels within a subpopulation.
```
adamgayoso, line 140: you are sampling from an importance weighted version of the posterior predictive right? It should be better explained below how else this can be used, and why it's different than `get_normalized_expression` so people aren't confused.

adamgayoso (changes requested):

scvi/external/wscvi/_module.py
```diff
@@ -0,0 +1,270 @@
...
+                local_library_log_vars,
+            ) = self._compute_local_library_params(batch_index)
+            log_pl = (
+                Normal(local_library_log_means, torch.sqrt(local_library_log_vars))
+                .log_prob(library)
+                .sum(-1)
+            )
+
+        log_px_latents = -self.get_reconstruction_loss(x, px_rate, px_r, px_dropout)
+        log_pjoint = log_px_latents + log_pz + log_pl
```
jjhong922, line 203: these terms can all be computed in loss right? this feels out of place.

scvi/external/wscvi/_module.py
```diff
@@ -0,0 +1,270 @@
...
+from scvi.module import VAE
+from scvi.module.base import LossRecorder, auto_move_data
+from scvi.nn import one_hot
+
+
+def reparameterize_gaussian(mu, var):
+    return Normal(mu, var.sqrt()).rsample()
+
+
+class WVAE(VAE):
```
jjhong922, line 20: This is a lot of duplicate code just for multiple particles and a modified loss function. I think moving the n_particles portion (default to 1) to VAE makes sense then having WVAE just overwrite the loss function (after addressing below comment).

After that change, if we believe in this method enough, I think it would be worthwhile making this an option (switch out the module) in SCVI rather than as an external model, considering the DEMixin functions are shared with SCVI anyway.

jjhong922 (commented):
Left my thoughts on what should be done with WSCVI, but contingent on the context of this method. Seems like some conflict resolution has to be done if we want https://github.com/scverse/scvi-tools/pull/1360 to come before this. Generally the multiple particle feature seems generalizable to live in SCVI, but the loss function can be separate and served as an alternative child module.

PR #1072, opened 2021-05-24
fix transform batch for models that encode covs

adamgayoso:
The way it's implemented right now transforms the batch before it goes through the encoder, which affects totalVI by default.

The desired behavior is to change only before decoding. This bug started with the 0.9 release (pytorch lightning move).


galenxing (changes requested):
Just need to remove `transform_batch` as a param to inference in totalvae and vae, but otherwise lgtm!

PR #1076, opened 2021-05-26
adding ability to get scores for simulated doublets

njbernstein:


scvi/external/solo/_model.py
```diff
@@ -324,15 +324,19 @@ def train(
         return runner()
 
     @torch.no_grad()
-    def predict(self, soft: bool = True) -> pd.DataFrame:
+    def predict(
+        self, soft: bool = True, only_experimental_data: bool = True
```
adamgayoso, line 328:
```suggestion
        self, soft: bool = True, include_simulated_doublets: bool = False
```

scvi/external/solo/_model.py
```diff
@@ -357,7 +361,11 @@ def auto_forward(module, x):
 
         label = self.adata.obs["_solo_doub_sim"].values.ravel()
         singlet_mask = label == "singlet"
```
adamgayoso, line 363:
```suggestion
        mask = label == "singlet" if not include_simulated_doublets else slice(None)
```

scvi/external/solo/_model.py
```diff
@@ -357,7 +361,11 @@ def auto_forward(module, x):
...
-        preds = y_pred[singlet_mask]
+        if only_experimental_data:
+            preds = y_pred[singlet_mask]
+        else:
+            preds = y_pred
```
adamgayoso, line 367:
```suggestion
        preds = y_pred[mask]
```

adamgayoso: also need to use mask and not singlet mask in line 373 in pandas df creation

scvi/external/solo/_model.py
```diff
@@ -324,15 +324,19 @@ def train(
...
+    ) -> pd.DataFrame:
         """
         Return doublet predictions.
 
         Parameters
         ----------
         soft
             Return probabilities instead of class label
-
+        only_experimental_data
```
adamgayoso, line 337: rename as above

adamgayoso (changes requested):
Just a few chages, looks great though!

scvi/external/solo/_model.py
```diff
@@ -324,15 +324,18 @@ def train(
         return runner()
 
     @torch.no_grad()
-    def predict(self, soft: bool = True) -> pd.DataFrame:
+    def predict(
+        self, soft: bool = True, include_simulated_doublets: bool = True
```
adamgayoso, line 328:
```suggestion
        self, soft: bool = True, include_simulated_doublets: bool = False
```
Default should be false right? I don't think most users care to see the sim doubs?

njbernstein: yup!

adamgayoso (approved):
LGTM!

PR #1078, opened 2021-05-26
classifier kwargs and fix adata option for solo

adamgayoso:
Fixes #1077 

Fixes an issue where `get_latent_library_size` returned samples on the log scale but mean on the linear scale

PR #1085, opened 2021-06-01
logging simulated doublets lib size

njbernstein:


njbernstein:
My attempt to fix what i see because of https://github.com/YosefLab/scvi-tools/pull/1078

adamgayoso (approved):
Thanks for catching that!

PR #1091, opened 2021-06-21
extend history when train method run multiple times

### LINKED ISSUE #1065: continue training of already saved model (extending TrainRunner)
Would be great to have the option to continue the training of the already saved model. @adamgayoso said this needs to go into the TrainRunner. One thing you need to do is to combine old and new training history like this:

```python
if continue_training and self.is_trained_:
    # add ELBO listory
    index = range(
      len(self.module.history_),
      len(self.module.history_)
      + len(trainer.logger.history["train_loss_epoch"]),
    )
    trainer.logger.history["train_loss_epoch"].index = index
    self.module.history_ = pd.concat(
      [self.module.history_, trainer.logger.history["train_loss_epoch"]]
    )
else:
    self.module.history_ = trainer.logger.history["train_loss_epoch"]
    self.history_ = self.module.history_
```

adamgayoso:
Fixes #1065

### DIFF
--- a/scvi/train/_trainrunner.py
+++ b/scvi/train/_trainrunner.py
@@ -95,6 +95,9 @@ class TrainRunner:
             else:
                 new_history = self.trainer.logger.history
                 for key, val in self.model.history_.items():
+                    # e.g., no validation loss due to training params
+                    if key not in new_history:
+                        continue
                     prev_len = len(val)
                     new_len = len(new_history[key])
                     index = np.arange(prev_len, prev_len + new_len)
--- a/scvi/train/_trainrunner.py
+++ b/scvi/train/_trainrunner.py
@@ -1,6 +1,9 @@
 import logging
+import warnings
 from typing import Optional, Union
 
+import numpy as np
+import pandas as pd
 import pytorch_lightning as pl
 
 from scvi.dataloaders import DataSplitter, SemiSupervisedDataSplitter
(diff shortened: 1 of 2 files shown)


scvi/train/_trainrunner.py
```diff
@@ -71,12 +74,42 @@ def __call__(self):
...
+    def _update_history(self):
+        # model is being further trained
+        if self.model.history_ is not None:
+            # if not using the default logger (e.g., tensorboard)
+            if not isinstance(self.model.history_, dict):
+                warnings.warn(
+                    "Training history cannot be updated. Replacing old history with new history."
+                )
+                if hasattr(self.model, "history_"):
+                    self.model.history_ = self.trainer.logger.history
```
galenxing, line 93: Let's move the warning message into the if statement, so it only warns if it actually does replace the old history. Also, I'm pretty sure every model has a `history_` attribute. It's in the init of BaseModelClass.

galenxing, line 93: Let's move the warning message into the if statement, so it only warns if it actually does replace the old history. Also, I'm pretty sure every model has a history_ attribute. It's in the init of BaseModelClass.

galenxing (changes requested):

scvi/train/_trainrunner.py
```diff
@@ -71,12 +74,42 @@ def __call__(self):
...
+                        [
+                            val,
+                            new_history[key],
+                        ]
+                    )
+                    self.model.history_[key].index.name = val.index.name
+        else:
+            # set history_ attribute if it exists
+            if hasattr(self.model, "history_"):
+                self.model.history_ = self.trainer.logger.history
```
adamgayoso, line 115:
```suggestion
            self.model.history_ = self.trainer.logger.history
```

galenxing (approved):

PR #1100, opened 2021-07-14
fix scanvi predict typo

### LINKED ISSUE #1092: The scANVI prediction results in an array, not a data frame.
Thank developers for bringing such a powerful tool.

When using scANVI's model to predict results, if the parameter is set to soft, an array will be returned instead of a dataframe

```python
scanviModel.predict(soft=True)
```

This appears to have been caused by a typo.
https://github.com/YosefLab/scvi-tools/blob/bc415d8a48195cbe7d145abd240961d8b5a10b0d/scvi/model/_scanvi.py#L299
The return value here should be `pred` instead of `y_pred`

galenxing:
fixes #1092

### DIFF
--- a/docs/release_notes/v0.12.0.rst
+++ b/docs/release_notes/v0.12.0.rst
@@ -25,7 +25,7 @@ Bug fixes
 - Fix issue where anndata setup with a layer led to errors in :class:`~scvi.external.SOLO` (`#1098`_).
 - Fix `adata` parameter of :func:`scvi.external.SOLO.from_scvi_model`, which previously did nothing (`#1078`_).
 - Fix default `max_epochs` of :class:`~scvi.model.SCANVI` when initializing using pre-trained model of :class:`~scvi.model.SCVI` (`#1079`_).
-- Fix bug in `predict()` function of :class:`~scvi.model.SCANVI`, which only occurred for soft predictions (`#1100_`).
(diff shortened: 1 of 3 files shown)


docs/release_notes/v0.12.0.rst
```diff
@@ -25,6 +25,7 @@ Bug fixes
 - Fix issue where anndata setup with a layer led to errors in :class:`~scvi.external.SOLO` (`#1098`_).
 - Fix `adata` parameter of :func:`scvi.external.SOLO.from_scvi_model`, which previously did nothing (`#1078`_).
 - Fix default `max_epochs` of :class:`~scvi.model.SCANVI` when initializing using pre-trained model of :class:`~scvi.model.SCVI` (`#1079`_).
+- Fix bug in `predict()` function of :class:`~scvi.model.SCANVI` (`#1092_`).
```
adamgayoso, line 28:
```suggestion
- Fix bug in `predict()` function of :class:`~scvi.model.SCANVI` for soft predictions (`#1092_`).
```

adamgayoso (changes requested):

docs/release_notes/v0.12.0.rst
adamgayoso: Also need to add the link at the bottom of the file.

adamgayoso (commented):
Also we should add a test...

PR #1104, opened 2021-07-29
pin pylit to < 1.4

adamgayoso:
Pytorch Lighting 1.4 causes some major issues, we can pin it for now.

https://python-poetry.org/docs/dependency-specification/#tilde-requirements


galenxing (approved):
LGTM!

PR #1122, opened 2021-08-24
update r notebook

adamgayoso:


jjhong922 (approved):

watiss:
@MeeseeksDev backport to 0.14.0-test

watiss:
@MeeseeksDev backport to test

watiss:
@MeeseeksDev backport to test

watiss:
@MeeseeksDev backport to adam/neural_deconv

watiss:
@MeeseeksDev backport to testtest

watiss:
@MeeseeksDev backport to msdtest

watiss:
@MeeseeksDev backport to msdtest

Carreau:
So I'm not too sure, maybe the bot fell into an inconsistent state. I've rebooted it – hopefully this will work.

My guess is maybe one of the first attempt created a workbranch that did not get cleaned up. I try to not erase the branches too often as the drive is slow on heroku and cloning can takes minutes for large repo like matplotlib.

you can retry see if it work, if not i'll add more debug capabilities.

watiss:
@MeeseeksDev backport to msdtest2

watiss:
@Carreau Thanks, I think I got it to work. It might be that I was doing it wrong, because I was branching off of a commit that was part of this PR that I was trying to backport (we dont squash commits when we merge, we should probably fix that). That might have been the reason why the bot was failing. Or it might be that the bot was in a bad state. We might never know :)

Carreau:
> * Do you still need to whitelist repo's for using MeeseeksDev bot? If so, could you whitelist scvi-tools?

No, every repo is allowed, there is no block list, or allow list. 

> * Must we always have a milestone with an "on-merge" comment to get the bot to backport or is a comment on the merged PR enough

Either a milestone or a label with `on-merge:`, I haven't implemented comments though it would be doable. 
One of the issue is that scanning all PRs comments would consume a lot of requests and I'm limited to 5000/hours.

> It might be that I was doing it wrong,

Nah, likely a bug on my side and caching gone wrong. Cloning repo like matplotlib takes forever so I have optimization to always work on the same git repo and it's not the first time they fail. Also some cleanup happens only every 24h or so when the bot restart, so it might be it.

PR #1123, opened 2021-08-27
Move library size distribution computation to model init

### LINKED ISSUE #898: Add library size prior as parameter of the module?
>        local_l_mean = tensors[_CONSTANTS.LOCAL_L_MEAN_KEY]
>        local_l_var = tensors[_CONSTANTS.LOCAL_L_VAR_KEY]

Looking back at this, I am bugged that we need those two lines in the loss function everywhere. Also, these are not "local" parameters in scVI. It would be nicer to remove those and get them as global parameters in all of our models.

What do you think?

jjhong922:
Closes #898 

This PR additionally:
- For several models ensures that the library encoder is not run if use_observed_lib_size is set to True.
- Fixes some bugs with incorporation of ql_m and ql_v even if use_observed_lib_size is set to True.


scvi/_constants.py
```diff
@@ -1,9 +1,13 @@
...
+class _CONSTANTS_NT(NamedTuple):
+    X_KEY: str = "X"
+    BATCH_KEY: str = "batch_indices"
+    LABELS_KEY: str = "labels"
+    PROTEIN_EXP_KEY: str = "protein_expression"
+    CAT_COVS_KEY: str = "cat_covs"
+    CONT_COVS_KEY: str = "cont_covs"
+
+
+_CONSTANTS = _CONSTANTS_NT()
```
adamgayoso, line 13: I like this

scvi/model/_utils.py
```diff
@@ -229,3 +230,32 @@ def _get_batch_code_from_category(
...
+def _init_library_size(
+    adata: anndata.AnnData, n_batch: dict
+) -> Tuple[np.ndarray, np.ndarray]:
+    data = get_from_registry(adata, _CONSTANTS.X_KEY)
+    batch_indices = get_from_registry(adata, _CONSTANTS.BATCH_KEY)
+
+    library_log_means = np.zeros(n_batch)
+    library_log_vars = np.ones(n_batch)
+
+    for i_batch in np.unique(batch_indices):
```
adamgayoso, line 244: going to take a more extensive review later, but this needs to look at the anndata setup dictionary and iterate over all batch categories that are registered. There is a niche case with scarches that more categories are registered than are in anndata. The shapes of the `library_log_means`, var should be number of registered batches.

adamgayoso (changes requested):

scvi/model/_utils.py
jjhong922: I thought `n_batch` should be updated when the number of categories are extended in the archesmixin. The shapes are being set to `n_batch` in this case, and in the for loop we just update the indices present in the batch_indices and the leave the rest as defaults (0,1).

scvi/model/_autozi.py
```diff
@@ -190,63 +203,72 @@ def get_marginal_ll(
...
                 library = inf_outputs["library"]
 
                 # Reconstruction Loss
-                current_dev = px_rate.device
                 bernoulli_params_batch = self.module.reshape_bernoulli(
                     bernoulli_params,
-                    batch_index.to(current_dev),
-                    labels.to(current_dev),
+                    batch_index,
+                    labels,
```
adamgayoso, line 224: did you test this on a GPU?

scvi/model/_autozi.py
```diff
@@ -190,63 +203,72 @@ def get_marginal_ll(
...
                 )
-                p_x_zld = -reconst_loss.to(p_z.device)
+                p_x_zld = -reconst_loss
+                log_prob_sum += p_z + p_x_zld
+
                 q_z_x = Normal(qz_m, qz_v.sqrt()).log_prob(z).sum(dim=-1)
-                q_l_x = Normal(ql_m, ql_v.sqrt()).log_prob(library).sum(dim=-1)
+                log_prob_sum -= q_z_x
+
+                if not self.use_observed_lib_size:
```
adamgayoso, line 247: fixing bugs too <3

scvi/model/_autozi.py
```diff
@@ -98,10 +102,15 @@ def __init__(
...
     ):
         super(AUTOZI, self).__init__(adata)
 
+        self.use_observed_lib_size = use_observed_lib_size
+        n_batch = self.summary_stats["n_batch"]
+        library_log_means, library_log_vars = _init_library_size(adata, n_batch)
+
         self.module = AutoZIVAE(
             n_input=self.summary_stats["n_vars"],
             n_batch=self.summary_stats["n_batch"],
```
adamgayoso, line 116: reuse prev defined `n_batch`? This happens in a few places I won't flag all of them

scvi/model/_utils.py
adamgayoso: ah this is correct

scvi/model/_totalvi.py
```diff
@@ -299,9 +306,14 @@ def get_latent_library_size(
         for tensors in post:
             inference_inputs = self.module._get_inference_input(tensors)
             outputs = self.module.inference(**inference_inputs)
-            ql_m = outputs["ql_m"]
-            ql_v = outputs["ql_v"]
-            if give_mean is True:
+            if give_mean:
+                ql_m = outputs["ql_m"]
+                ql_v = outputs["ql_v"]
+                if ql_m is None or ql_v is None:
```
adamgayoso, line 312: doesn't look like totalvi will be none based on the way it was coded.

scvi/model/_utils.py
```diff
@@ -229,3 +230,32 @@ def _get_batch_code_from_category(
...
+            warnings.warn(
+                "This dataset has some empty cells, this might fail inference."
+                "Data should be filtered with `scanpy.pp.filter_cells()`"
+            )
+
+        log_counts = masked_log_sum.filled(0)
+        library_log_means[i_batch] = np.mean(log_counts).astype(np.float32)
+        library_log_vars[i_batch] = np.var(log_counts).astype(np.float32)
+
+    return library_log_means.reshape(1, -1), library_log_vars.reshape(1, -1)
```
adamgayoso, line 261: a comment of the shapes here would be helpful

adamgayoso (changes requested):

adamgayoso:
@jjhong922 I think we also need to update one of the developer guide tutorials on the tutorials repo

scvi/model/_autozi.py
jjhong922: no, I can try it out. Is it just running pytest w/ a GPU?/

scvi/model/_autozi.py
```diff
@@ -177,7 +188,7 @@ def get_marginal_ll(
         )
 
         log_lkl = 0
-        to_sum = torch.zeros((n_mc_samples,))
+        to_sum = torch.zeros((n_mc_samples,)).to(self.device)
```
jjhong922, line 191: @galenxing is there any reason this tensor was kept on cpu when you implemented it?

adamgayoso (changes requested):
Two things:

1. Are we properly testing the `use_observed_lib_size` option?
2. Can you run gimVI tutorial with this code (on colab) and see that result is similar?

jjhong922:
1. Added tests and caught a bug in my totalvi code. Ran on GPU.
2. GIMVI notebook outputs look good: https://colab.research.google.com/drive/1EferyWhZp44fe7Zp-Q5NCO-qPd2QVdUe?usp=sharing

adamgayoso (approved):
LGTM

PR #1129, opened 2021-09-02
Update Documentation to explain how we make releases and backport patches

watiss:


adamgayoso (changes requested):
LGTM except I'm not a fan of the extra link in the nav bar. I think all in one for now is fine as within a page you get the right side nav. In the future we can have multiple pages inside "contributing". Thoughts?

watiss:
Personally, I kind of like the separation. This is what Pandas does too: https://pandas.pydata.org/docs/development/maintaining.html. But I'm fine with leaving it as it is today and updating later if we see fit.
Update coming in shortly

adamgayoso (approved):
LGTM. We can make it like Pandas soon (multiple pages within a development tab)

adamgayoso:

PR #1132, opened 2021-09-02
LDA implementation

jjhong922:
Autoencoder Variational Bayes implementation of Latent Dirichlet Allocation as a PyroModule. Runs in the same magnitude of time as sklearn's implementation of LDA and achieves better perplexity on average. See benchmark here: https://colab.research.google.com/drive/1Iq_drlBTLadM8KJtZwIs96RdzG8MmEFA?authuser=1#scrollTo=lINSWBVshwvr.

Note: to have this work with reparametrization stably, I used a logistic normal to approximate the dirichlet distribution.

### DIFF
--- a/scvi/model/_amortizedlda.py
+++ b/scvi/model/_amortizedlda.py
@@ -93,13 +93,6 @@ class AmortizedLDA(PyroSviTrainMixin, BaseModelClass):
 
         self.init_params_ = self._get_init_params(locals())
 
-    def _check_if_not_trained(self):
-        """Check if the model is not trained. Raises a ValueError if not."""
-        if not self.is_trained_:
-            raise ValueError(
-                "Trying to query inferred values from an untrained model. Please train the model first."
-            )
-
     def get_gene_by_topic(self, give_mean=True) -> pd.DataFrame:
         """
         Gets the gene by topic matrix.
@@ -116,7 +109,7 @@ class AmortizedLDA(PyroSviTrainMixin, BaseModelClass):
         -------
         A `n_var x n_topics` Pandas DataFrame containing the gene by topic matrix.
         """
-        self._check_if_not_trained()
+        self._check_if_trained(warn=False)
 
         topic_by_gene = self.module.topic_by_gene(give_mean=give_mean)
 
@@ -153,7 +146,7 @@ class AmortizedLDA(PyroSviTrainMixin, BaseModelClass):
         A `n_obs x n_topics` Pandas DataFrame containing the normalized estimate
         of the topic distribution for each observation.
         """
-        self._check_if_not_trained()
+        self._check_if_trained(warn=False)
         adata = self._validate_anndata(adata)
 
         dl = self._make_data_loader(adata=adata, indices=indices, batch_size=batch_size)
@@ -198,7 +191,7 @@ class AmortizedLDA(PyroSviTrainMixin, BaseModelClass):
         -------
         The positive ELBO.
         """
-        self._check_if_not_trained()
+        self._check_if_trained(warn=False)
         adata = self._validate_anndata(adata)
 
         dl = self._make_data_loader(adata=adata, indices=indices, batch_size=batch_size)
@@ -235,7 +228,7 @@ class AmortizedLDA(PyroSviTrainMixin, BaseModelClass):
         -------
         Perplexity.
         """
-        self._check_if_not_trained()
+        self._check_if_trained(warn=False)
         adata = self._validate_anndata(adata)
 
         dl = self._make_data_loader(adata=adata, indices=indices, batch_size=batch_size)
--- a/scvi/model/_amortizedlda.py
+++ b/scvi/model/_amortizedlda.py
@@ -9,7 +9,7 @@ import torch
 from anndata import AnnData
 
 from scvi._constants import _CONSTANTS
-from scvi.module import LDAPyroModule
+from scvi.module import AmortizedLDAPyroModule
 
 from .base import BaseModelClass, PyroSviTrainMixin
 
@@ -83,7 +83,7 @@ class AmortizedLDA(PyroSviTrainMixin, BaseModelClass):
                 f"a float or a Sequence of length n_input."
             )
 
-        self.module = LDAPyroModule(
+        self.module = AmortizedLDAPyroModule(
             n_input=n_input,
             n_topics=n_topics,
             n_hidden=n_hidden,
--- a/scvi/model/_lda.py
+++ b/scvi/model/_lda.py
@@ -1,4 +1,4 @@
-import collections
+import collections.abc
 import logging
 from typing import Optional, Sequence, Union
 
@@ -54,7 +54,7 @@ class LDA(PyroSviTrainMixin, BaseModelClass):
             cell_topic_prior is not None
             and not isinstance(cell_topic_prior, float)
             and (
-                not isinstance(cell_topic_prior, collections.Sequence)
+                not isinstance(cell_topic_prior, collections.abc.Sequence)
                 or len(cell_topic_prior) != n_topics
             )
         ):
@@ -66,7 +66,7 @@ class LDA(PyroSviTrainMixin, BaseModelClass):
             topic_gene_prior is not None
             and not isinstance(topic_gene_prior, float)
             and (
-                not isinstance(topic_gene_prior, collections.Sequence)
+                not isinstance(topic_gene_prior, collections.abc.Sequence)
                 or len(topic_gene_prior) != n_input
             )
         ):
--- a/scvi/model/_lda.py
+++ b/scvi/model/_lda.py
@@ -92,7 +92,7 @@ class LDA(PyroSviTrainMixin, BaseModelClass):
                 "Trying to query inferred values from an untrained model. Please train the model first."
             )
 
-    def get_gene_by_topics(self, give_mean=True) -> pd.DataFrame:
+    def get_gene_by_topic(self, give_mean=True) -> pd.DataFrame:
         """
         Gets the gene by topic matrix.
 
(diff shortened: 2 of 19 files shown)


scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+    component_gene_prior
+        Prior of component gene distribution. If `None`, defaults to `1 / n_components`.
+    """
+
+    def __init__(
+        self,
+        adata: AnnData,
+        n_components: int = 20,
+        n_hidden: int = 128,
+        cell_component_prior: Optional[float] = None,
```
adamgayoso, line 41: should we allow this to be sequences of floats as well of len(n_components)?

adamgayoso: as well as the other prior...

scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+    def _check_if_not_trained(self):
+        """Check if the model is not trained. Raises a ValueError if not."""
+        if not self.is_trained_:
+            raise ValueError(
+                "Trying to query inferred values from an untrained model. Please train the model first."
+            )
+
+    def get_components(self) -> pd.DataFrame:
+        """
+        Gets the component to gene transition matrix.
```
adamgayoso, line 86:
```suggestion
        Gets the gene by topic matrix.
```

scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+            )
+
+    def _check_if_not_trained(self):
+        """Check if the model is not trained. Raises a ValueError if not."""
+        if not self.is_trained_:
+            raise ValueError(
+                "Trying to query inferred values from an untrained model. Please train the model first."
+            )
+
+    def get_components(self) -> pd.DataFrame:
```
adamgayoso, line 84:
```suggestion
    def get_genes_by_topics(self) -> pd.DataFrame:
```

scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+
+        Returns
+        -------
+        A `n_components x n_var` Pandas DataFrame containing the component to gene transition matrix.
+        """
+        self._check_if_not_trained()
+
+        return pd.DataFrame(
+            data=self.module.components.numpy(), columns=self.adata.var_names
+        )
```
adamgayoso, line 102: should we add a `give_mean=True` param and also allow sampling from the dirichlet?

scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def transform(self, adata: Optional[AnnData] = None) -> pd.DataFrame:
```
adamgayoso, line 104:
```suggestion
    def get_latent_representation(self, adata: Optional[AnnData] = None) -> pd.DataFrame:
```

This is just to be consistent with the vae-based models

adamgayoso: this also needs all the same params as the vaemixin get_latent... (e.g., batch size, etc)

scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+        user_adata = adata or self.adata
+        dl = self._make_data_loader(adata=user_adata)
+
+        transformed_xs = []
+        for tensors in dl:
+            x = tensors[_CONSTANTS.X_KEY]
+            transformed_xs.append(self.module.transform(x))
+
+        transformed_x = torch.cat(transformed_xs).numpy()
+        return pd.DataFrame(data=transformed_x, index=user_adata.obs_names)
```
adamgayoso, line 131: I would add columns here like "topic_{}"

scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def get_elbo(self, adata: Optional[AnnData] = None) -> float:
```
adamgayoso, line 133: this should mirror the vaemixin one in terms of functionality

scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+        ----------
+        adata
+            AnnData to compute ELBO for. If None, returns the ELBO for the source AnnData.
+
+        Returns
+        -------
+        The negative ELBO.
+        """
+        if adata is not None:
+            self._check_var_equality(adata)
```
adamgayoso, line 147: why not run `self._validate_anndata(adata)`?

adamgayoso: this occurs many times so only commenting here

scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+        dl = self._make_data_loader(adata=user_adata)
+
+        elbos = []
+        for tensors in dl:
+            x = tensors[_CONSTANTS.X_KEY]
+            library = x.sum(dim=1)
+            elbos.append(self.module.get_elbo(x, library, user_adata.n_obs))
+        return np.mean(elbos)
+
+    def perplexity(self, adata: Optional[AnnData] = None) -> float:
```
adamgayoso, line 160: should have same params as elbo, maybe also rename to get_perplexity? idk

scvi/module/_lda.py
```diff
@@ -0,0 +1,299 @@
...
+        self.register_buffer(
+            "cell_component_prior",
+            torch.full((self.n_components,), cell_component_prior),
+        )
+        self.register_buffer(
+            "component_gene_prior", torch.full((self.n_input,), component_gene_prior)
+        )
+
+        # Hack: to allow auto_move_data to infer device.
+        self._dummy = torch.nn.Parameter(torch.zeros(1), requires_grad=False)
```
adamgayoso, line 70: it's a little cleaner to register a buffer here. This only works because our optimizer filters params based on the requires grad flag

scvi/module/_lda.py
```diff
@@ -0,0 +1,299 @@
...
+    @auto_move_data
+    def forward(
+        self, x: torch.Tensor, library: torch.Tensor, n_obs: Optional[int] = None
+    ):
+        # Component gene distributions.
+        with pyro.plate("components", self.n_components):
+            component_gene_dist = pyro.sample(
+                "component_gene_dist",
+                dist.Dirichlet(self.component_gene_prior),
+            ).to(x.device)
```
adamgayoso, line 90:
```suggestion
            )
```

scvi/module/_lda.py
```diff
@@ -0,0 +1,299 @@
...
+        return self._model
+
+    @property
+    def guide(self):
+        return self._guide
+
+    @property
+    def components(self) -> torch.Tensor:
+        """
+        Gets the component to gene transition matrix from the Pyro parameter store.
```
adamgayoso, line 246: this is no longer true

scvi/module/_lda.py
```diff
@@ -0,0 +1,299 @@
...
+        -------
+        A `x.shape[0] x n_components` tensor containing the normalized component distribution.
+        """
+        cell_component_unnormalized_dist = self._unnormalized_transform(x)
+        return (
+            cell_component_unnormalized_dist
+            / cell_component_unnormalized_dist.sum(axis=1)[:, np.newaxis]
+        )
+
+    def get_elbo(self, x: torch.Tensor, library: torch.Tensor, n_obs: int) -> float:
```
adamgayoso, line 282: does this need the torch no grad decorator?

scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+        adata
+            AnnData to compare against.
+        """
+        source_var_names = self.adata.var_names.astype(str)
+        user_var_names = adata.var_names.astype(str)
+        if not np.array_equal(user_var_names, source_var_names):
+            raise ValueError(
+                "`adata` passed into `transform` does not have matching var_names "
+                "with the source adata the model was trained with."
+            )
```
adamgayoso, line 75: not sure why this is necessary when we have `self._validate_anndata`

scvi/model/_lda.py
```diff
@@ -0,0 +1,183 @@
...
+        super().__init__(adata)
+
+        self.module = LDAPyroModule(
+            n_input=self.summary_stats["n_vars"],
+            n_components=n_components,
+            n_hidden=n_hidden,
+            cell_component_prior=cell_component_prior,
+            component_gene_prior=component_gene_prior,
+        )
+
```
adamgayoso, line 56: believe you need to call the init_params = locals thing (see SCVI init)

tests/models/test_pyro.py
```diff
@@ -376,3 +377,31 @@ def test_pyro_bayesian_train_sample_mixin_with_local_full_data():
         adata.n_obs,
         1,
     )
+
+
+def test_lda_model():
```
adamgayoso, line 382: looks good, just need to test save/load

adamgayoso (changes requested):

scvi/module/_lda.py
jjhong922: but automove only iterates through model parameters. register_buffer doesn't add it to the parameters. Otherwise, this line wouldnt' be necessary since the previous lines already registered some stuff

scvi/train/_trainingplans.py
```diff
@@ -677,6 +695,19 @@ def optimizer_step(self, *args, **kwargs):
...
+        epoch_criterion = self.n_epochs_kl_warmup is not None
+        step_criterion = self.n_steps_kl_warmup is not None
+        if epoch_criterion:
+            kl_weight = min(1.0, self.current_epoch / self.n_epochs_kl_warmup)
+        elif step_criterion:
+            kl_weight = min(1.0, self.global_step / self.n_steps_kl_warmup)
+        else:
+            kl_weight = 1.0
+        return max(kl_weight, 1e-3)
+
```
adamgayoso, line 710: maybe we write a general util in the file and call that so we don't duplicate code?

adamgayoso (changes requested):

scvi/model/_amortizedlda.py
```diff
@@ -0,0 +1,247 @@
...
+        )
+
+        self.init_params_ = self._get_init_params(locals())
+
+    def _check_if_not_trained(self):
+        """Check if the model is not trained. Raises a ValueError if not."""
+        if not self.is_trained_:
+            raise ValueError(
+                "Trying to query inferred values from an untrained model. Please train the model first."
+            )
```
adamgayoso, line 101: should we add this to our base model class?

jjhong922: Sure

scvi/module/_amortizedlda.py
```diff
@@ -0,0 +1,347 @@
...
+    def get_topic_distribution(
+        self, x: torch.Tensor, give_mean: bool = True
+    ) -> torch.Tensor:
+        """
+        Converts `x` to its inferred topic distribution.
+
+        Parameters
+        ----------
+        give_mean
+            Give mean of distribution or sample from it.
```
adamgayoso, line 313: missing x here

scvi/module/_amortizedlda.py
```diff
@@ -0,0 +1,349 @@
...
+            self._validate_sample(value)
+        logits, value = dist.util.broadcast_all(self.logits, value)
+        logits = logits.clone(memory_format=torch.contiguous_format)
+        logits[(value == 0) & (logits == -math.inf)] = 0
+        log_powers = (logits * value).sum(-1)
+        return log_powers
+
+
+def logistic_normal_approximation(
+    alpha: torch.Tensor,
```
adamgayoso, line 31: can you add a comment/docstring about what's going on here?

adamgayoso (approved):
LGTM after the last small change

scvi/module/_amortizedlda.py
jjhong922: Also realized that the sigma here was the variance when I should be returning the std, don't think it actually changes a lot though

PR #1142, opened 2021-09-03
Add support for DDP.

tchaton:
This PR is a draft and not completed, but it shows a slightly different approach which is compatible with DDP.

Explanation: PyTorch Lightning injects a DistributedSampler within the DataLoader. To do so, it tries to access init arguments from dataloader attribues.

With the current custom AnnDataLoader, AnnDataset isn't saved, and therefore Lightning will fail.

To avoid further bug, I would recommend to avoid overriding the DataLoader unless there is a strong requirement or at least make sure it can re-instantianted from its attributes.

Best regards,
T.C

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,6 +1,6 @@
 [[package]]
 name = "absl-py"
-version = "0.12.0"
+version = "0.13.0"
 description = "Abseil Python Common Libraries, see https://github.com/abseil/abseil-py."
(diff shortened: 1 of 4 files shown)


adamgayoso:
Closing as this PR is stale and we have a solution we are working on involving disabling our custom sampler, which won't be necessary in the multi gpu regime

PR #1151, opened 2021-09-09
Use pytorch lightning's seed everything to set seed

### LINKED ISSUE #1111: Use pytorch lightning's seed everything in our config to set seed
We should refactor this function

https://github.com/YosefLab/scvi-tools/blob/7f06d616ddb6af8b3b0f5ec630f338e4a873e122/scvi/_settings.py#L141-L147

to use pytorch lightning's [`seed_everything`](https://pytorch-lightning.readthedocs.io/en/latest/api/pytorch_lightning.utilities.seed.html#pytorch_lightning.utilities.seed.seed_everything) which looks similar.

mjayasur:
Fix #1111

### DIFF
--- a/scvi/_settings.py
+++ b/scvi/_settings.py
@@ -2,7 +2,6 @@ import logging
 from pathlib import Path
 from typing import Union
 
-import numpy as np
 import pytorch_lightning as pl
 import torch
 from rich.console import Console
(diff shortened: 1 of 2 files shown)


watiss (approved):

adamgayoso (changes requested):
We should be able to properly test this right?

1. Set the seed and train a model 1 epoch
2. Set the seed again and do the same
3. NN weights should be exactly the same

adamgayoso (approved):

PR #1150, opened 2021-09-09
[AnnData setup refactor] Make setup_anndata a static method on model classes rather than one global function

watiss:
Closes #1135 
Closes #1107

PR #1152, opened 2021-09-10
Fix CondSCVI docstring

jjhong922:


adamgayoso (commented):
Isn't this going to be irrelevant with the setup anndata PR? @watiss

watiss:
> Isn't this going to be irrelevant with the setup anndata PR? @watiss

It's still going to be applicable but I'm changing this same line in my pr so I can also just incorporate this fix. @jjhong922  if you're cool with it, would you mind leaving a comment on my pr so it doesn't fall off the radar?

jjhong922:
Will fix in #1150 1150

PR #1158, opened 2021-09-14
Bug fix in pyromixin for posterior sampling

vitkl:
The method `._get_obs_plate_sites()` should discard observed variables (unless requested otherwise) and plates. Current code always performs posterior sampling even if unobserved local variables do not exist in the model, e.i. code never enters this statement:
```
if len(obs_plate_sites) == 0:
    # if no local variables - don't sample
    break
```

This PR addresses the issue by adding a check for observed and plates.

### DIFF
--- a/scvi/model/base/_pyromixin.py
+++ b/scvi/model/base/_pyromixin.py
@@ -236,7 +236,7 @@ class PyroSampleMixin:
         dictionary {variable_name: [array with samples in 0 dimension]}
         """
         samples = self._get_one_posterior_sample(
-            args, kwargs, return_sites=return_sites, sample_observed=return_observed
+            args, kwargs, return_sites=return_sites, return_observed=return_observed
         )
         samples = {k: [v] for k, v in samples.items()}
 
@@ -249,7 +249,7 @@ class PyroSampleMixin:
 
             # generate new sample
             samples_ = self._get_one_posterior_sample(
-                args, kwargs, return_sites=return_sites, sample_observed=return_observed
+                args, kwargs, return_sites=return_sites, return_observed=return_observed
             )
 
             # add new sample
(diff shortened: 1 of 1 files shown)


vitkl:
@adamgayoso please have a look at this

scvi/model/base/_pyromixin.py
```diff
@@ -262,7 +262,7 @@ def _get_obs_plate_return_sites(self, return_sites, obs_plate_sites):
         else:
             return obs_plate_sites
 
-    def _get_obs_plate_sites(self, args, kwargs):
+    def _get_obs_plate_sites(self, args, kwargs, sample_observed):
```
adamgayoso, line 265: can you document this function and add relevant typing?

adamgayoso (changes requested):
LGTM outside of the small request for docstring

vitkl:
I think this is done now.

scvi/model/base/_pyromixin.py
```diff
@@ -262,7 +270,12 @@ def _get_obs_plate_return_sites(self, return_sites, obs_plate_sites):
         else:
             return obs_plate_sites
 
-    def _get_obs_plate_sites(self, args, kwargs):
+    def _get_obs_plate_sites(
+        self,
+        args: Optional[list],
+        kwargs: Optional[dict],
```
adamgayoso, line 276: usually optional has a None default. Don't believe this is the case here?

adamgayoso (changes requested):

vitkl:
Yes, you are right actually - these arguments are not optional and don't
have default values. I will change this.

scvi/model/base/_pyromixin.py
adamgayoso:
```suggestion
        args: list,
        kwargs: dict,
```

adamgayoso (approved):

PR #1169, opened 2021-09-23
LDA logistic normal mean fix

### LINKED ISSUE #1168: LDA topic distribution and gene_by_topic should return an MC sample
The logistic normal distribution has no analytic solution for the mean, so we must take samples for an MC estimate for an accurate solution. This means these functions will be stochastic, but as is they are not correct. We will parametrize them with a num_samples parameter instead.

jjhong922:
Fixes #1168 

`get_gene_by_topic` and `get_latent_representation` now both are stochastic and take n_samples as an optional argument defaulting to 5000. We need to give a MC estimate of the mean of the logistic normal since there is no analytic solution. What was implemented before was just incorrect.

Tested in benchmarking colab notebook: https://colab.research.google.com/drive/1Iq_drlBTLadM8KJtZwIs96RdzG8MmEFA?authuser=1#scrollTo=R0LTgkF8gR_a

### DIFF
--- a/scvi/module/_amortizedlda.py
+++ b/scvi/module/_amortizedlda.py
@@ -279,16 +279,16 @@ class AmortizedLDAPyroModule(PyroBaseModuleClass):
     def guide(self):
         return self._guide
 
-    def topic_by_gene(self, give_mean: bool = True) -> torch.Tensor:
+    def topic_by_gene(self, n_samples: int) -> torch.Tensor:
         """
-        Gets the topic by gene matrix.
+        Gets a Monte-Carlo estimate of the expectation of the topic by gene matrix.
 
         Assumes the module has already been trained.
 
         Parameters
         ----------
-        give_mean
-            Give mean of distribution if True or sample from it.
+        n_samples
+            Number of samples to take for the Monte-Carlo estimate of the mean.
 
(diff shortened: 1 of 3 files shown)


scvi/module/_amortizedlda.py
```diff
@@ -298,40 +298,51 @@ def topic_by_gene(self, give_mean: bool = True) -> torch.Tensor:
...
-            return F.softmax(topic_gene_posterior_mu, dim=1)
-        return F.softmax(
-            torch.normal(topic_gene_posterior_mu, topic_gene_posterior_sigma),
-            dim=1,
+        return torch.mean(
+            F.softmax(
+                torch.normal(
+                    topic_gene_posterior_mu.unsqueeze(2).repeat(1, 1, n_samples),
+                    topic_gene_posterior_sigma.unsqueeze(2).repeat(1, 1, n_samples),
+                ),
```
adamgayoso, line 306: we typically use[ torch Normal distribution](https://pytorch.org/docs/stable/distributions.html#normal) object and its sample shape argument

adamgayoso (changes requested):

adamgayoso (approved):

PR #1181, opened 2021-09-28
Sphinx panels for docs index page

adamgayoso:


adamgayoso:
There's weird offset of the images inside the cards. Also we can do this to the installation page if we wish:

https://pandas.pydata.org/pandas-docs/stable/getting_started/index.html

(top part)

jjhong922:
What's the reason for this change? The way it looked before looked nicer imo

jjhong922:
Answered offline: makes code way cleaner to do the same thing and make more things clickable. Styles will be fixed

jjhong922 (approved):
clean

PR #1182, opened 2021-09-29
Cite-seq datasets

adamgayoso:
Adding some built in data so I can create a tutorial in a cleaner way.

### DIFF
--- a/scvi/data/_built_in_data/_cite_seq.py
+++ b/scvi/data/_built_in_data/_cite_seq.py
@@ -1,8 +1,10 @@
 import os
 
 import anndata
+import numpy as np
 import pandas as pd
 
+from scvi import settings
 from scvi.data._anndata import _setup_anndata
 from scvi.data._built_in_data._download import _download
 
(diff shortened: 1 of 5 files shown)


scvi/data/_built_in_data/_cite_seq.py
```diff
@@ -143,3 +145,80 @@ def _load_spleen_lymph_cite_seq(
...
+    aggregate_proteins: bool = True,
+    mask_protein_batches: int = 0,
+    run_setup_anndata: bool = True,
+):
+    url = "https://ndownloader.figshare.com/files/27458840"
+    save_fn = "pbmc_seurat_v4.h5ad"
+    _download(url, save_path, save_fn)
+    adata = anndata.read_h5ad(os.path.join(save_path, save_fn))
+
+    if aggregate_proteins is True:
```
watiss, line 162: nit: can remove the `is True`

scvi/data/_built_in_data/_cite_seq.py
```diff
@@ -143,3 +145,80 @@ def _load_spleen_lymph_cite_seq(
...
+        adata.obs["RNA log library size"] = np.log(adata.X.sum(1).A)
+
+        # actually filter
+        adata = adata[adata.obs["Protein log library size"] > 7.6]
+        adata = adata[adata.obs["Protein log library size"] < 10.3]
+        adata = adata[adata.obs["Number proteins detected"] > 150]
+        # filter doublet
+        adata = adata[adata.obs["celltype.l2"] != "Doublet"]
+        # MT
+        adata = adata[adata.obs["pct_counts_mt"] < 12].copy()
```
watiss, line 207: Why do we make a copy here?

watiss (approved):

scvi/data/_built_in_data/_cite_seq.py
adamgayoso: with anndata, without the copy that will be a view of the object in memory, which is not compatible with setup anndata

scvi/data/_built_in_data/_cite_seq.py
adamgayoso: fixed

scvi/data/_built_in_data/_cite_seq.py
watiss: ahhh.
Curiosity question: is this copy an expensive copy the same way as any `adata2 = adata.copy()` is (where adata is _not_ a view)? Or is it a call that doesn't involve any actual byte transfer, but instead just marks the view as a full adata?

adamgayoso: this is a good question.... and I don't have the exact answer

PR #1197, opened 2021-10-04
Add release notes for v0.14.0

watiss:


docs/release_notes/v0.14.0.rst
```diff
@@ -0,0 +1,54 @@
+New in 0.14.0 (2021-10-04)
+--------------------------
+
```
adamgayoso, line 3: let's add a paragraph here in plain english mentioning the user guide, the new lda model, and the setup anndata change with a code example (before -> after) in the case of one particular model

docs/release_notes/v0.14.0.rst
```diff
@@ -0,0 +1,54 @@
...
+Changes
+~~~~~~~
+- Added fixes to support PyTorch Lightning 1.4 (`#1103`_)
+- Simplified data handling in R tutorials with sceasy and addressed bugs in package installation (`#1122`_).
+- Moved library size distribution computation to model init (`#1123`_)
+- Updated Contribution docs to describe how we backport patches (`#1129`_)
+- Implemented Latent Dirichlet Allocation as a PyroModule (`#1132`_)
+- Restricted PyTorch Lightning version to <1.4 (`#1145`_)
```
adamgayoso, line 11: I believe this one happened in 0.13.1

adamgayoso: and then maybe I lifted it and put it back

adamgayoso (changes requested):

docs/release_notes/v0.14.0.rst
watiss: Ah ok, I can remove it then.

docs/release_notes/v0.14.0.rst
```diff
@@ -0,0 +1,71 @@
+New in 0.14.0 (2021-10-04)
```
adamgayoso, line 1: we will change this the day we merge it

docs/release_notes/v0.14.0.rst
```diff
@@ -0,0 +1,71 @@
...
+b. The relationship between the math and the functions associated with each model 
+c. The relationship between math variables and code variables 
+
+Our previous User Guide guide has been renamed to Tutorials and contains all of our existing tutorials (including tutorials for developers).
+
+Another noteworthy addition in this release is the implementation of the (amortized) Latent Dirichlet Allocation (aka LDA) model applied to single-cell gene expression data. We have also prepared a tutorial that demonstrates how to use this model, using a PBMC 10K dataset from 10x Genomics as an example application.
+
+Lastly, in this release we have made a change to reduce user and developer confusion by making the previously global ``setup_anndata`` method a static class-specific method instead. This provides more clarity on which parameters are applicable for this call, for each model class. Below is a before/after for the SOLO and TOTALVI model classes:
+
+.. figure:: figures/setup_anndata_before_after.svg
```
adamgayoso, line 16: for solo, most users will not call setup anndata because they use an alternative constructor that does it for them. I would pick another model. Also, the image is nice but you can just put code directly in this document

adamgayoso (changes requested):

docs/release_notes/v0.14.0.rst
watiss: Ah good call on Solo

docs/release_notes/v0.14.0.rst
```diff
@@ -0,0 +1,71 @@
...
+
+a. The math behind each method (in a succinct, online methods-like way) 
+b. The relationship between the math and the functions associated with each model 
+c. The relationship between math variables and code variables 
+
+Our previous User Guide guide has been renamed to Tutorials and contains all of our existing tutorials (including tutorials for developers).
+
+Another noteworthy addition in this release is the implementation of the (amortized) Latent Dirichlet Allocation (aka LDA) model applied to single-cell gene expression data. We have also prepared a tutorial that demonstrates how to use this model, using a PBMC 10K dataset from 10x Genomics as an example application.
+
+Lastly, in this release we have made a change to reduce user and developer confusion by making the previously global ``setup_anndata`` method a static class-specific method instead. This provides more clarity on which parameters are applicable for this call, for each model class. Below is a before/after for the SOLO and TOTALVI model classes:
```
adamgayoso, line 14:
```suggestion
Lastly, in this release we have made a change to reduce user and developer confusion by making the previously global ``setup_anndata`` method a static class-specific method instead. This provides more clarity on which parameters are applicable for this call, for each model class. Below is a before/after for the DESTVI and TOTALVI model classes:
```

adamgayoso (approved):

PR #1198, opened 2021-10-04
return dataframe instead of matrix

### LINKED ISSUE #1119: Denoised atac/rna from PeakVI/MultiVI should return pandas dataframe
We need to add the `return_numpy` argument as it is in this function:

https://github.com/YosefLab/scvi-tools/blob/e0632cecd019c15cfc73e6c39c75aac7906bde4a/scvi/model/base/_rnamixin.py#L41-L163

The tutorial can then be updated.

talashuach:
closes #1119


scvi/model/_multivi.py
```diff
@@ -406,13 +406,14 @@ def get_accessibility_estimates(
...
-        region_indices: Sequence[int] = None,
+        region_list: Optional[Sequence[str]] = None,
         transform_batch: Optional[Union[str, int]] = None,
         use_z_mean: bool = True,
         threshold: Optional[float] = None,
         normalize_cells: bool = False,
         normalize_regions: bool = False,
         batch_size: int = 128,
+        return_numpy: bool = False,
     ) -> Union[np.ndarray, csr_matrix]:
```
adamgayoso, line 417:
```suggestion
    ) -> Union[np.ndarray, csr_matrix, pd.DataFrame]:
```

adamgayoso (changes requested):

PR #1213, opened 2021-10-08
Update LDA tutorial image

justjhong:
Updates LDA image to show wordcloud after removing mitochondrial genes. Updates LDA tutorial from submodule as well.


watiss:
> This is good, but we should establish the backporting strategy with @watiss

Why don't you guys follow the guide I wrote up [here](https://docs.scvi-tools.org/en/stable/contributing/index.html#backporting) so that we can test how helpful it actually is :)

jjhong922:

PR #1228, opened 2021-10-14
Bug fix: Stereoscope sc model loss

adamgayoso:
To better reproduce the original implementation, we use the sum of a minibatch loss instead of the mean


adamgayoso:
@jjhong922 @watiss this is worthy of another patch release

romain-lopez:
LGTM

PR #1238, opened 2021-10-21
AnnData Manager refactor

jjhong922:
This PR introduces the AnnDataManager class as well as accompanying AnnDataField classes. These classes aim to bring structure to how data registration implementation code is organized.

The AnnDataManager serves the role of wrapping an AnnData object and specifying how that AnnData object is setup to interface properly with a specific model class. Each AnnDataField details how a single data field is referenced in the AnnData object, and any preprocessing steps required for the field.

This PR only has the new implementation complete for the scVI model, which still has bugs in accepting the new mapping for data loading. This will be addressed in future PRs and other models will follow with the introduction of other necessary AnnDataField implementations.


scvi/data/anndata/_compat.py
```diff
@@ -0,0 +1,46 @@
...
+    adata
+        AnnData object to be registered.
+    setup_dict
+        Setup dictionary created after registering an AnnData using an AnnDataManager object.
+    **kwargs
+        Keyword arguments to modify transfer behavior.
+    """
+    source_adata_manager = AnnDataManager()
+    data_registry = setup_dict[_constants._DATA_REGISTRY_KEY]
+    for scvi_key, adata_mapping in data_registry.items():
```
adamgayoso, line 29: I think it would be a good time to give `scvi_key` a new name. Perhaps `registry_key` or something else more informative. This is a global change, but I'll only comment here

scvi/data/anndata/_fields/_base_field.py
```diff
@@ -0,0 +1,88 @@
...
+    An AnnDataField class defines how scvi-tools will map a data field used by a model
+    to an attribute in an AnnData object.
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    @property
+    @abstractmethod
+    def scvi_key(self):
```
adamgayoso, line 23:
```suggestion
    def registry_key(self):
```

scvi/model/_scvi.py
```diff
@@ -163,3 +164,24 @@ def setup_anndata(
             continuous_covariate_keys=continuous_covariate_keys,
             copy=copy,
         )
+
+    @setup_anndata_dsp.dedent
+    def anndata_fields(batch_key=None, labels_key=None, layer=None):
```
adamgayoso, line 169: I think this is a bit much for devs. Let's try the decorator as we discussed on slack

tests/models/test_models.py
```diff
@@ -40,6 +37,19 @@
...
+def test_new_setup():
+    adata = synthetic_iid(run_setup_anndata=False)
+    adata2 = adata.copy()
+    SCVI.old_setup_anndata(
+        adata,
+        batch_key="batch",
+        labels_key="labels",
+    )
+
+    SCVI.setup_anndata(adata2, batch_key="batch", labels_key="labels")
```
adamgayoso, line 49: should the new transfer work for scvi?

adamgayoso (changes requested):
Initial feedback

tests/models/test_models.py
jjhong922: for batch and labels yes

scvi/model/base/_base_model.py
```diff
@@ -407,28 +419,40 @@ def __repr__(
...
     @abstractmethod
+    def anndata_fields(*args, **kwargs) -> Set[Type[BaseAnnDataField]]:
+        """Returns a set of AnnDataField instances used to register data for this model."""
+        pass
+
+    @classmethod
+    @setup_anndata_dsp.dedent
+    def _register_manager(cls, adata_manager: AnnDataManager):
+        """
+        $(adata_fields_summary)s.
```
adamgayoso, line 434: check if should be $ or %

scvi/data/anndata/_fields/_layer_field.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @property
+    def attr_key(self):
+        return self._attr_key
+
+    def validate_field(self, adata: AnnData) -> None:
+        super().validate_field(adata)
+        x = self.get_field(adata)
+
+        if _check_nonnegative_integers(x) is False:
```
adamgayoso, line 41: this should be an option, default is True. probably in the init of this class?

adamgayoso: i.e., not all models will require data to be count

PR #1247, opened 2021-11-12
Implement NonCategoricalJointObsField

jjhong922:
- Adds BaseObsmField and a child class JointObsField which allows for combination of obs field to form a obsm field to be fed into the model.
- Implements a NonCategoricalJointObsField which supports the continuous_covariate_keys in the SCVI model.


scvi/model/_scvi.py
```diff
@@ -5,7 +5,11 @@
 
 from scvi._compat import Literal
 from scvi._constants import _CONSTANTS
-from scvi.data.anndata._fields import CategoricalObsField, LayerField
+from scvi.data.anndata._fields import (
```
adamgayoso, line 8: we should make this import "public", yes? and then in the next PR we can start building the docs for the new classes

scvi/data/anndata/_fields/_obsm_field.py
```diff
@@ -0,0 +1,103 @@
...
+    @property
+    def attr_key(self):
+        return self._attr_key
+
+    @property
+    def is_empty(self) -> bool:
+        return self._is_empty
+
+
+class ContinuousObsmField(BaseObsmField):
```
adamgayoso, line 58: this should be able to cover e.g., protein expression for totalvi as well. Not sure this is the best name as counts are not continuous. Maybe `NonCategoricalObsmField` is better.

Also, in this PR, we could add the components to cover protein expression -- checking if counts, data reformatting to C contiguous memory, csr instead of coo etc.

adamgayoso (changes requested):

scvi/model/_scvi.py
jjhong922: sure

scvi/data/anndata/_fields/_obsm_field.py
jjhong922: I think they are sufficiently different to make it a separate class that inherits from BaseObsmField class. This is because this class takes obs_keys and combines them into a new obsm field whereas the protein expression field is expected to be in a obsm field already, then reformatted etc. I tinkered around w/ making this class do both but it gets crowded

scvi/data/anndata/_compat.py
```diff
@@ -1,7 +1,9 @@
 from anndata import AnnData
 
+from scvi._constants import _CONSTANTS
+
 from . import _constants
-from ._fields import CategoricalObsField, LayerField
+from ._fields import CategoricalObsField, LayerField, NonCategoricalJointObsField
```
adamgayoso, line 6: `NumericalJointObsField` might be a better name

scvi/data/anndata/fields/_obsm_field.py
```diff
@@ -0,0 +1,102 @@
...
+        return self._attr_key
+
+    @property
+    def is_empty(self) -> bool:
+        return self._is_empty
+
+
+class NonCategoricalJointObsField(JointObsField):
+    """
+    An AnnDataField for non-categorical .obs fields in the AnnData data structure.
```
adamgayoso, line 65: we're going to need to document all the classes you created, and when referencing scvi tools objects, you can use `:class:~path_to_class_from_import`

adamgayoso: I'm suggesting that we should just document everything now incrementally instead of all at once

scvi/data/anndata/fields/_obsm_field.py
```diff
@@ -0,0 +1,102 @@
...
+        for obs_key in self._obs_keys:
+            assert obs_key in adata.obs, f"{obs_key} not found in adata.obs."
+
+    def register_field(self, adata: AnnData) -> None:
+        super().register_field(adata)
+        self._combine_obs_fields(adata)
+        # TODO(jhong): remove if not necessary.
+        adata.uns[_constants._SETUP_DICT_KEY][self._columns_key] = self.get_field(
+            adata
+        ).columns.to_numpy()
```
adamgayoso, line 86: I would double check, this might have to be a list due to anndata save issues

adamgayoso: on that note, adding tests to check save load works will be important

scvi/data/anndata/fields/_base_field.py
```diff
@@ -26,16 +27,22 @@ def registry_key(self):
...
     @abstractmethod
-    def attr_key(self):
+    def attr_key(self) -> Optional[str]:
         """The key of the data field within the relevant AnnData attribute."""
         pass
 
+    @property
+    @abstractmethod
+    def is_empty(self) -> bool:
+        """Returns True if the field is empty as a function of its kwargs."""
```
adamgayoso, line 43: this could better describe why this could happen

adamgayoso: as in, I don't understand

scvi/data/anndata/fields/_obsm_field.py
```diff
@@ -0,0 +1,102 @@
...
+        return self._attr_name
+
+
+class JointObsField(BaseObsmField):
+    def __init__(self, registry_key: str, obs_keys: Optional[List[str]]) -> None:
+        super().__init__(registry_key)
+        self._attr_key = f"_scvi_{registry_key}"
+        self._columns_key = f"{self.registry_key}_keys"
+        self._obs_keys = obs_keys if obs_keys is not None else []
+        self._is_empty = len(self.obs_keys) == 0
```
adamgayoso, line 41: ah I see, this can have a default state of empty

adamgayoso: could still use a better description

adamgayoso (changes requested):

scvi/data/anndata/fields/_obsm_field.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def _combine_obs_fields(self, adata: AnnData) -> None:
+        series = []
+        for key in self.obs_keys:
+            s = adata.obs[key]
+            series.append(s)
+        adata.obsm[self.attr_key] = pd.concat(series, axis=1)
```
adamgayoso, line 48: this is probably inefficient, should be able to subset obs all at once

scvi/data/anndata/fields/_obsm_field.py
jjhong922: Agree just didn't think it would make sense to reference the BaseAnnDataField class here since I'm referring to the general collection of implemented fields. (i.e. Type[BaseAnnDataField])

scvi/data/anndata/fields/_obsm_field.py
jjhong922: This was copied directly from what it was before so im gonna assume it works

scvi/data/anndata/fields/_obsm_field.py
jjhong922: what do you mean by this? I made it into a generator expression just to make it concise, but don't think its any faster

adamgayoso: like if you do `adata.obsm[self.attr_key] = adata.obs[[self._obs_keys]].copy()`?

adamgayoso (approved):

PR #1237, opened 2021-10-21
setup_anndata and AnnDataManager Refactor

jjhong922:
Fixes #945
Fixes #1136

PR #1248, opened 2021-11-12
Implement CategoricalJointObsField

jjhong922:
- Adds the CategoricalJointObsField which supports the categorical_covariate_keys argument in the SCVI model.

### DIFF
--- a/scvi/data/anndata/fields/_obsm_field.py
+++ b/scvi/data/anndata/fields/_obsm_field.py
@@ -1,7 +1,9 @@
 import logging
-from typing import List, Optional
+from typing import Dict, List, Optional
 
+import numpy as np
 from anndata import AnnData
+from pandas.api.types import CategoricalDtype
 
 from scvi.data.anndata import _constants
 
@@ -48,9 +50,13 @@ class JointObsField(BaseObsmField):
     def __init__(self, registry_key: str, obs_keys: Optional[List[str]]) -> None:
         super().__init__(registry_key)
         self._attr_key = f"_scvi_{registry_key}"
-        self._columns_key = f"{self.registry_key}_keys"
         self._obs_keys = obs_keys if obs_keys is not None else []
         self._is_empty = len(self.obs_keys) == 0
+
+    def validate_field(self, adata: AnnData) -> None:
+        super().validate_field(adata)
+        for obs_key in self._obs_keys:
+            assert obs_key in adata.obs, f"{obs_key} not found in adata.obs."
 
     def _combine_obs_fields(self, adata: AnnData) -> None:
         adata.obsm[self.attr_key] = adata.obs[self.obs_keys].copy()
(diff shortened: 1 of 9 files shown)


scvi/data/anndata/fields/_obsm_field.py
```diff
@@ -92,11 +97,100 @@ def transfer_field(
...
+
+    def _initialize_mappings_dict(self, adata: AnnData) -> None:
+        adata.uns[_constants._SETUP_DICT_KEY][self._mappings_key] = dict(
+            mappings=dict(), keys=[]
+        )
+
+    def _make_obsm_categorical(
+        self, adata: AnnData, category_dict: Optional[Dict[str, List[str]]] = None
+    ) -> None:
+        assert self.obs_keys == self.get_field(adata).columns.tolist()
```
adamgayoso, line 133: codacy will be sad to see assert in the code, you should raise an error

scvi/data/anndata/fields/_obsm_field.py
```diff
@@ -92,11 +97,100 @@ def transfer_field(
...
+    def __init__(self, registry_key: str, obs_keys: Optional[List[str]]) -> None:
+        super().__init__(registry_key, obs_keys)
+        self._mappings_key = registry_key
+
+    def _initialize_mappings_dict(self, adata: AnnData) -> None:
+        adata.uns[_constants._SETUP_DICT_KEY][self._mappings_key] = dict(
+            mappings=dict(), keys=[]
+        )
+
+    def _make_obsm_categorical(
```
adamgayoso, line 130: does this make obsm categorical? It looks like it just records the categories per covariate

adamgayoso (changes requested):

scvi/data/anndata/fields/_obsm_field.py
```diff
@@ -92,11 +97,100 @@ def transfer_field(
...
+        setup_dict: dict,
+        adata_target: AnnData,
+        **kwargs,
+    ) -> None:
+        super().transfer_field(setup_dict, adata_target, **kwargs)
+
+        if self.is_empty:
+            return
+
+        extend_categories = getattr(kwargs, "extend_categories", False)
```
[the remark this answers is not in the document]
jjhong922: it's because this is called over a list of fields here: https://github.com/YosefLab/scvi-tools/pull/1237/files#diff-614a605196c42e8b5f08a458dc15b37502526431a01c1597c4dbbc44d67d4ab9R102.
As a result, we don't have field specific kwargs (unless you pass in a dictionary which is more cumbersome).

scvi/data/anndata/fields/_obsm_field.py
jjhong922: good point

adamgayoso: should we change the name?

scvi/data/anndata/fields/_obsm_field.py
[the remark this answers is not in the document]
adamgayoso: I'm still not sure I understand, it should be able to be an arg with default as false in the signature?

adamgayoso (changes requested):

scvi/data/anndata/fields/_obsm_field.py
jjhong922: i made it so it does make it categorical

PR #1249, opened 2021-11-12
Torch metrics for ELBO

### LINKED ISSUE #1113: Evaluate whether torchmetrics makes sense for ELBO
Perhaps we can use TorchMetrics (https://torchmetrics.readthedocs.io/en/latest/) to standardize the elbo calculation and reduce some code in the training plans.

PierreBoyeau:
Fixes #1113 

The current codebase computing training & validation metrics at the end of each epoch suffers from several problems. First, it has a lot of code duplication. Second, it requires to (i). compute batch-level metrics (ii). average them manually.
As an alternative, `pytorchmetrics` can solve this exact problem with minimal effort

This pull request aims at solving these two problems using `pytorchmetrics`

# TODO

- Verify that `weight` in `MeanMetric` has the proper behavior, i.e., that the metric is normalized by \sum_i w_i.
- Add tests to verify that all keys are conserved & have the proper names
- Add `pytorchmetric` dependency


scvi/train/_trainingplans.py
```diff
@@ -124,6 +125,19 @@ def __init__(
...
+            kl_local=MeanMetric(),
+            kl_global=MeanMetric(),
+        )
+        self.val_running_metrics = dict(
+            elbo=MeanMetric(),
+            reconstruction_loss=MeanMetric(),
+            kl_local=MeanMetric(),
+            kl_global=MeanMetric(),
+        )
+
```
adamgayoso, line 140: let's talk about this in person, I think we need to implement a custom metric object, which will even further reduce the custom methods you added.

adamgayoso (changes requested):

scvi/train/_trainingplans.py
adamgayoso: well actually, I think we just write `ElboMetric` as a subclass of their object, and it keeps track of the running metrics and the `compute` method would return a dictionary of elbo and it's components.

PierreBoyeau: Sounds good, this will also be easier if we want to include custom metrics natively

PierreBoyeau:
Refactored the code; metrics are now stored into a single class.
My assumption was that `the elbo, the reconstruction loss, and the KL metrics are always going to be logged. Other, model-specific metrics can be tracked, but may need to modify the `log_metrics`method (see the semisupervised training plan).

When I think about it, maybe we should instead have a model-specific Metric class instead, and let `log_metrics` to remain the same for all models.

If you agree, then my plan is to:
1. modify the current code to introduce a semisupervisedmetric
2. double-check for errors in the elbo computation

In this paradigm, I feel that the Metrics should belong in another file than the training plans, but I am not 100% sure about that

adamgayoso:
@PierreBoyeau I made some changes and force pushed. There's something about this torch metric that is making .grad of tensors None after training, and it's messing up some scarches stuff.

PierreBoyeau:
Thanks so much for your help @adamgayoso. I'll have a look at the code and see what to do from there. Can't we just clone the metrics or something?

adamgayoso:
@PierreBoyeau I found the embarrassing bug... I wasn't returning a loss -- weird how it was working in the first place

adamgayoso:
@PierreBoyeau can you review this PR?

scvi/train/_metrics.py
```diff
@@ -0,0 +1,53 @@
+import torch
+from torchmetrics import Metric
+
+
+class ElboMetric(Metric):
+    def __init__(self, n_obs_total: int, dist_sync_on_step: bool = False):
```
jjhong922, line 6: typing should be Optional[int] if you are accepting None

scvi/train/_trainingplans.py
```diff
@@ -141,68 +160,121 @@ def n_obs_training(self, n_obs: int):
...
+        self._n_obs_validation = n_obs
+        self.initialize_val_metrics()
 
     def forward(self, *args, **kwargs):
         """Passthrough to `model.forward()`."""
         return self.module(*args, **kwargs)
 
+    def compute_and_log_metrics(
+        self,
+        lossrecorder: LossRecorder,
```
jjhong922, line 191:
```suggestion
        loss_recorder: LossRecorder,
```

scvi/train/_trainingplans.py
```diff
@@ -141,68 +160,121 @@ def n_obs_training(self, n_obs: int):
...
+        Parameters
+        ----------
+        lossrecorder
+            LossRecorder object from scvi-tools module
+        metric_attr_name
+            The name of the torch metric object to use
+        mode
+            Train or validation, used for logging names
+        """
+        lr = lossrecorder
```
jjhong922, line 207: potentially unnecessary and confusing to abbrev as lr (learning rate)

scvi/train/_trainingplans.py
```diff
@@ -141,68 +160,121 @@ def n_obs_training(self, n_obs: int):
...
+        self.initialize_val_metrics()
 
     def forward(self, *args, **kwargs):
         """Passthrough to `model.forward()`."""
         return self.module(*args, **kwargs)
 
+    def compute_and_log_metrics(
+        self,
+        lossrecorder: LossRecorder,
+        metric_attr_name: str,
```
jjhong922, line 192: can't you just pass in the metric instead of inferring it with a string name?

scvi/train/_trainingplans.py
```diff
@@ -141,68 +160,121 @@ def n_obs_training(self, n_obs: int):
...
+        mode: Literal["train", "validation"],
```
jjhong922, line 193: can you bake this into the metric instance rather than having to specify it everytime? I can't think of a situation where a given metric would be used w/ multiple modes.

scvi/train/_trainingplans.py
```diff
@@ -141,68 +160,121 @@ def n_obs_training(self, n_obs: int):
...
-            kl_local += tensors["kl_local_sum"]
-            n_obs += tensors["n_obs"]
-        # kl global same for each minibatch
-        kl_global = outputs[0]["kl_global"]
-        elbo += kl_global
-        self.log("elbo_train", elbo / n_obs)
-        self.log("reconstruction_loss_train", rec_loss / n_obs)
-        self.log("kl_local_train", kl_local / n_obs)
-        self.log("kl_global_train", kl_global)
+        self.compute_and_log_metrics(
```
jjhong922, line 267: nice code cleanup :)

scvi/train/_metrics.py
jjhong922: my suggestion earlier is to add another kwarg here like `mode:str` or `train=True`.

jjhong922 (commented):
some code quality comments. Otherwise the code factored out looks great

scvi/train/_trainingplans.py
adamgayoso: reconstruction loss and kl's need this to be uniquely named

scvi/train/_trainingplans.py
adamgayoso: done

scvi/train/_trainingplans.py
adamgayoso: done

scvi/train/_trainingplans.py
jjhong922: I get that but can you pass it into the ELBOMetric() class as an arg. Then later reference it with metric.mode

adamgayoso:
There's something going on with the first epoch:

Latest release, 10 epochs of scvi on synthetic iid:

```
      elbo_train
epoch           
0        338.495
1        331.845
2        327.447
3         323.57
4        320.989
5        318.708
6        317.046
7        315.781
8        314.374
9        313.127
```

New code:
```
      elbo_train
epoch           
0          2.005
1        331.845
2        327.447
3         323.57
4        320.989
5        318.708
6        317.046
7        315.781
8        314.374
9        313.127
```

PierreBoyeau:
@adamgayoso : I checked for reproducibility of the metrics in [this notebook](https://colab.research.google.com/drive/1m_tFmCwQk5indfRzuyDvSYKOT8VLQtAr?usp=sharing) for the autoZI model, and it looks like all metrics make sense.

adamgayoso:
There are some glitches going on with this PR and python 3.8, as well as Codacy. I'm going to merge now

PR #1253, opened 2021-11-15
fix load example

### LINKED ISSUE #1087: Loading saved DestVI.from_rna_model dosn't consider two anndatas used to create model
When saving the DestVI model, it requires the anndata as input to scvi.model.SCVI.load. 
There are two anndatas used to create the model; the reference scRNA-seq data, or the spatial transcriptomics data. Trying either of these, I get errors when attempting to reload the model. 

It appears reloading the data uses some information from both anndata,
but the scvi.model.SCVI.load function was only built to consider one. 

I am following the vignette: https://docs.scvi-tools.org/en/stable/user_guide/notebooks/DestVI_tutorial.html
So the creation of the model is shown as below:
```
scvi.data.setup_anndata(vdata_, labels_key='cell_labels')
sc_model = CondSCVI(vdata_, weight_obs=True)
sc_model.train(max_epochs=100)

###### Now performing the deconvolution #########
scvi.data.setup_anndata(data_, batch_key='batch')

st_model = DestVI.from_rna_model(data_, sc_model)

st_model.train(max_epochs=250)

# The proportions ! #
data_.obsm["proportions"] = st_model.get_proportions()
```

Then I attempt to save using either the scRNA-seq adata (vdata_) or the spatial transcriptomics adata (data_):
In the first case:
```
st_model.save(f"{h5_dir}my_model/")
model = scvi.model.SCVI.load(f"{h5_dir}my_model/", vdata_)
```

In the second case, I run:

```
model = scvi.model.SCVI.load(f"{h5_dir}my_model/", data_)
```

In the first case (scRNA-seq as the adata), i get this error:
```pytb
scvi/data/_anndata.py in _transfer_batch_and_labels(adata_target, categorical_mappings, extend_categories)
    510             raise KeyError(
    511                 '.obs["{}"] was used to setup source, but not found in target.'.format(
--> 512                     original_key
    513                 )
    514             )

KeyError: '.obs["batch"] was used to setup source, but not found in target.'
```

In the second case (spatial as adata), i get this error:
```pytb
scvi/model/_scvi.py in __init__(self, adata, n_hidden, n_latent, n_layers, dropout_rate, dispersion, gene_likelihood, latent_distribution, **model_kwargs)
    102             gene_likelihood=gene_likelihood,
    103             latent_distribution=latent_distribution,
--> 104             **model_kwargs,
    105         )
    106         self._model_summary_string = (

TypeError: __init__() got an unexpected keyword argument 'cell_type_mapping'
```

#### Versions:
scvi.__version__
'0.11.0'
> VERSION

Thanks alot for developing a fantastic tool!

### LINKED ISSUE #1088: Load docstring examples show only SCVI globally
We should figure out how to have unique examples in the docstring for each child class.

See #1087

adamgayoso:
Fixes #1088
Fixes #1087

I went for a generic and conservative solution here.

### DIFF
--- a/scvi/model/base/_base_model.py
+++ b/scvi/model/base/_base_model.py
@@ -346,7 +346,7 @@ class BaseModelClass(ABC):
 
         Examples
         --------
-        >>> model = ModelClass.load(save_path, adata)
+        >>> model = ModelClass.load(save_path, adata) # use the name of the model class used to save
         >>> model.get_....
         """
         load_adata = adata is None
--- a/scvi/model/base/_base_model.py
+++ b/scvi/model/base/_base_model.py
@@ -346,8 +346,8 @@ class BaseModelClass(ABC):
 
         Examples
         --------
-        >>> vae = SCVI.load(save_path, adata)
-        >>> vae.get_latent_representation()
+        >>> model = ModelClass.load(save_path, adata)
+        >>> model.get_....
         """
         load_adata = adata is None
         use_gpu, device = parse_use_gpu_arg(use_gpu)


scvi/model/base/_base_model.py
```diff
@@ -346,8 +346,8 @@ def load(
 
         Examples
         --------
-        >>> vae = SCVI.load(save_path, adata)
-        >>> vae.get_latent_representation()
+        >>> model = ModelClass.load(save_path, adata)
+        >>> model.get_....
```
watiss, line 350: Personally I would either remove the Examples section altogether or call out this example is specifically for the SCVI class.
This is also fine, it just does not add too much value than if we had no Examples IMO.

watiss (approved):

scvi/model/base/_base_model.py
adamgayoso: I added a comment

adamgayoso:

PR #1280, opened 2021-11-24
Pass manager to data loader

jjhong922:
Passes AnnDataManager object to data loaders and splitters. This will enable the following in coming PRs:
- Model-data specific setups (less clobbering)
- setup dict saved on the manager object, not the AnnData object.

### DIFF
--- a/scvi/model/base/_base_model.py
+++ b/scvi/model/base/_base_model.py
@@ -13,11 +13,11 @@ from anndata import AnnData
 
 from scvi import _CONSTANTS, settings
 from scvi.data._utils import _check_nonnegative_integers
-from scvi.data.anndata import get_from_registry, transfer_anndata_setup
+from scvi.data.anndata import get_from_registry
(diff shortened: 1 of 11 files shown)


scvi/model/base/_base_model.py
```diff
@@ -119,6 +140,16 @@ def _make_data_loader(
         data_loader_kwargs
             Kwargs to the class-specific data loader class
         """
+        if adata is None:
+            adata_manager = self.adata_manager
+        else:
+            adata_manager = self.get_anndata_manager(adata)
+            if adata_manager is None:
```
adamgayoso, line 147: should we proceed here with transfer? This function is typically run after validation. See get latent rep for example...

adamgayoso: perhaps we raise an error to go run validate?

jjhong922: i see, that sounds good

PR #1284, opened 2021-12-02
Copy the scvi state dict during `SCANVI` initialization to prevent it from being overwritten within `scvi_model`

### LINKED ISSUE #1283: Scanvi should copy scvi state dict on initialization
https://github.com/YosefLab/scvi-tools/blob/7d9e9f74754bb361fd4f25e7db0a13967c1ef739/scvi/model/_scanvi.py#L203

This would prevent an issue where the original scvi model weights get updated

watiss:
Fixes #1283

### DIFF
--- a/scvi/model/_scanvi.py
+++ b/scvi/model/_scanvi.py
@@ -200,7 +200,7 @@ class SCANVI(RNASeqMixin, VAEMixin, ArchesMixin, BaseModelClass):
         scanvi_model = cls(
             adata, unlabeled_category, **non_kwargs, **kwargs, **scanvi_kwargs
         )
-        scvi_state_dict = scvi_model.module.state_dict()
+        scvi_state_dict = scvi_model.module.state_dict().copy()
         scanvi_model.module.load_state_dict(scvi_state_dict, strict=False)
         scanvi_model.was_pretrained = True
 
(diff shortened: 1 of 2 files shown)


scvi/model/_scanvi.py
```diff
@@ -200,7 +200,7 @@ def from_scvi_model(
         scanvi_model = cls(
             adata, unlabeled_category, **non_kwargs, **kwargs, **scanvi_kwargs
         )
-        scvi_state_dict = scvi_model.module.state_dict()
+        scvi_state_dict = scvi_model.module.state_dict().copy()
```
adamgayoso, line 203: I think we might need deepcopy here. We should add a test to check that it works -- pick any tensor inside the state dict and see if it's not equal

adamgayoso (changes requested):

scvi/model/_scanvi.py
watiss: So, within Pytorch's `load_state_dict` function (the call we make right below this one), they do a simple copy:
```
       # copy state_dict so _load_from_state_dict can modify it
        metadata = getattr(state_dict, '_metadata', None)
        state_dict = state_dict.copy()
```
Could you explain why we would need a deepcopy in our case?

Sure, I will add a test.

watiss: Talked offline. We don't think we need a `copy()` or `deepcopy()` as PyTorch does this internally.
I added a test though.

adamgayoso (approved):

adamgayoso:

PR #1285, opened 2021-12-03
Move and reorganize setup dict out of AnnData into Manager

jjhong922:
This PR includes:
- New setup dict organization more amenable to generalizable fields (now called a registry)
  - The organization is as such: registry -> some metadata as well as field_registries
  - field_registries: keys are registry_keys (the ones you feed into setup_anndata - guaranteed to be unique), values are field_registry dictionaries
  - field_registry has a data_registry (same as before), summary_stats (same as before), and a new state_registry which holds any field specific information (e.g. categorical_mappings, extra_continuous_keys)
- Dataloading and various utils have been modified to take in an anndata manager object rather than an anndata since this is now necessary to access setup metadata.

### DIFF
--- a/scvi/data/anndata/_utils.py
+++ b/scvi/data/anndata/_utils.py
@@ -21,7 +21,7 @@ from . import _constants
 logger = logging.getLogger(__name__)
 
 
-def _get_field(
+def get_anndata_attribute(
     adata: anndata.AnnData, attr_name: str, attr_key: Optional[str]
 ) -> np.ndarray:
     adata_attr = getattr(adata, attr_name)
@@ -82,7 +82,7 @@ def get_from_registry(adata: anndata.AnnData, key: str) -> np.ndarray:
         data_loc[_constants._DR_ATTR_KEY],
     )
 
-    return _get_field(adata, attr_name, attr_key)
+    return get_anndata_attribute(adata, attr_name, attr_key)
 
 
 @deprecated(
@@ -326,7 +326,7 @@ def _verify_and_correct_data_format(adata, data_registry):
             mapping[_constants._DR_ATTR_NAME],
             mapping[_constants._DR_ATTR_KEY],
         )
-        data = _get_field(adata, attr_name, attr_key)
+        data = get_anndata_attribute(adata, attr_name, attr_key)
         if isspmatrix(data) and (data.getformat() != "csr"):
             warnings.warn(
                 "Training will be faster when sparse matrix is formatted as CSR. It is safe to cast before model initialization."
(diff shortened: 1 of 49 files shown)


scvi/model/_totalvi.py
```diff
@@ -1045,6 +1046,79 @@ def _validate_anndata(
...
+        import warnings
+
+        from sklearn.exceptions import ConvergenceWarning
+        from sklearn.mixture import GaussianMixture
+
+        warnings.filterwarnings("error")
+
+        adata = self._validate_anndata(adata)
+        batch = self.get_from_registry(adata, _CONSTANTS.BATCH_KEY).ravel()
+        cats = adata.uns["_scvi"]["categorical_mappings"]["_scvi_batch"]["mapping"]
```
adamgayoso, line 1060: is this safe now?

scvi/data/anndata/_utils.py
```diff
@@ -765,7 +777,8 @@ def _make_obs_column_categorical(
         warnings.warn(
             "Is adata.obs['{}'] continuous? SCVI doesn't support continuous obs yet."
         )
```
adamgayoso, line 779: this could probably be changed as well

scvi/model/base/_base_model.py
```diff
@@ -44,8 +41,8 @@ def __init__(self, adata: Optional[AnnData] = None):
         if adata is not None:
             self.adata_manager = self.get_anndata_manager(adata, required=True)
             self.adata = self.adata_manager.adata
-            self.scvi_setup_dict_ = self.adata_manager.get_setup_dict()
-            self.summary_stats = self.scvi_setup_dict_[_SUMMARY_STATS_KEY]
+            self.registry = self.adata_manager.registry
```
adamgayoso, line 44: should this be registry or data_registry? And if it's the latter should registry be a private attribute? I'm not sure I understand the distinction between the two in the manager

scvi/model/base/_base_model.py
```diff
@@ -112,9 +109,39 @@ def get_anndata_manager(
 
         return cls.manager_store[adata_uuid]
 
+    def get_from_registry(
+        self,
+        adata: AnnData,
+        registry_key: str,
+    ) -> np.ndarray:
```
adamgayoso, line 116: this actually also can return pandas dataframe, like if protein expression is a df

adamgayoso (changes requested):

scvi/model/_totalvi.py
jjhong922: no I haven't moved all these references yet, I'm planning on doing that in chunks

scvi/data/anndata/_utils.py
jjhong922: should it be removed?

PR #1294, opened 2021-12-17
Save/Load with saved setup inputs

jjhong922:
This PR enables save load with the new registry in addition to backwards compatibility in the old setup dict.

This is done by adding a **kwargs argument to setup_anndata for the purpose of transferring setup from a loaded model. Developers will be instructed to follow the boilerplate of the original function to do the same thing (moving it into it's own function is avoided to keep the developer experience consistent).

### DIFF
--- a/scvi/data/anndata/_compat.py
+++ b/scvi/data/anndata/_compat.py
@@ -77,7 +77,10 @@ def registry_from_setup_dict(setup_dict: dict) -> dict:
 
 
 @deprecated(
-    extra="The save format of models has been updated. Please update your saved model files accordingly."
+    extra=(
+        "The save format of models has been updated. Please update your saved model files accordingly. "
+        "Backwards compatibility with be removed in v1.0."
+    )
 )
 def manager_from_setup_dict(
     cls, adata: AnnData, setup_dict: dict, **transfer_kwargs
--- a/scvi/data/anndata/_compat.py
+++ b/scvi/data/anndata/_compat.py
@@ -4,13 +4,13 @@ from anndata import AnnData
 from sklearn.utils import deprecated
 
 from . import _constants
+from ._manager import AnnDataManager
 from .fields import (
     CategoricalJointObsField,
     CategoricalObsField,
     LayerField,
     NumericalJointObsField,
 )
-from .manager import AnnDataManager
 
 
 def registry_from_setup_dict(setup_dict: dict) -> dict:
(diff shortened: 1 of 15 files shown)


scvi/data/anndata/_compat.py
```diff
@@ -75,8 +76,11 @@ def registry_from_setup_dict(setup_dict: dict) -> dict:
     return registry
 
 
+@deprecated(
+    extra="The save format of models has been updated. Please update your saved model files accordingly."
+)
```
adamgayoso, line 81: I think we should say this will be removed in v1.0?

scvi/data/anndata/manager.py
```diff
@@ -99,8 +101,8 @@ def register_fields(
         ----------
         adata
             AnnData object to be registered.
-        source_setup_dict
-            Setup dictionary created after registering an AnnData using an AnnDataManager object.
+        source_registry
+            Registry created after registering an AnnData using an AnnDataManager object.
```
adamgayoso, line 105:
```suggestion
            Registry created after registering an AnnData using an :class:`scvi.data.AnnDataManager` object.
```
not 100% sure this is correct, but syntax is

scvi/data/anndata/manager.py
```diff
@@ -145,9 +147,7 @@ def register_fields(
         self._assign_uuid()
         self._assign_source_uuid(source_registry)
 
-    def transfer_setup(
-        self, adata_target: AnnData, source_registry: Optional[dict] = None, **kwargs
-    ) -> AnnDataManager:
+    def transfer_setup(self, adata_target: AnnData, **kwargs) -> AnnDataManager:
```
adamgayoso, line 150: need to document **kwargs

scvi/data/anndata/manager.py
```diff
@@ -212,6 +206,15 @@ def summary_stats(self) -> dict:
 
         return summary_stats
 
+    @property
+    def setup_inputs(self) -> dict:
+        """Returns the setup inputs, including the model name, that were used to initialize this AnnDataManager instance."""
```
adamgayoso, line 211:
```suggestion
        """Returns the setup inputs, including the model name, that were used to initialize this :class:`scvi.data.AnnDataManager` instance."""
```

adamgayoso: inputs -> arguments?

scvi/model/base/_base_model.py
```diff
@@ -41,7 +48,7 @@ def __init__(self, adata: Optional[AnnData] = None):
         if adata is not None:
             self.adata_manager = self.get_anndata_manager(adata, required=True)
             self.adata = self.adata_manager.adata
-            self.registry = self.adata_manager.registry
+            self.registry_ = self.adata_manager.registry
```
adamgayoso, line 51: I would just add a comment above why there's a _

scvi/model/_scvi.py
```diff
@@ -203,6 +209,6 @@ def setup_anndata(
             ),
             NumericalJointObsField(_CONSTANTS.CONT_COVS_KEY, continuous_covariate_keys),
         ]
-        adata_manager = AnnDataManager(fields=anndata_fields)
-        adata_manager.register_fields(adata)
+        adata_manager = AnnDataManager(fields=anndata_fields, setup_inputs=setup_inputs)
```
adamgayoso, line 212:
```suggestion
        adata_manager = AnnDataManager(fields=anndata_fields, setup_meth_kwargs=setup_inputs)
```
thoughts?

adamgayoso (changes requested):

adamgayoso (approved):

PR #1295, opened 2021-12-18
Model specific scvi-tools generated adata fields

jjhong922:
Fields can optionally specify a model_name kwarg which will prefix any columns created by that field's register_field function with the model_name. This will prevent cloberring between models.

I decided this should be a field init kwarg since not all fields create columns and fields must be initialized before the Manager and the convention is that field's carry no state other than their initialization params. Thus, it doesn't make sense for the Manager to inject the model name during the registration process.


jjhong922:
Will rethink this and come up with a better solution which leverages the fact that we already have model name in AnnDataManager on initialization.

PR #1301, opened 2022-01-12
Adapt all internal models to new setup

jjhong922:
To adapt some models, this PR also adds an LabelsWithUnlabeledObsFIeld for PeakVI.

### DIFF
--- a/scvi/data/anndata/fields/_scanvi.py
+++ b/scvi/data/anndata/fields/_scanvi.py
@@ -1,0 +1,89 @@
+from typing import Optional, Union
+
+import numpy as np
+from anndata import AnnData
+from pandas.api.types import CategoricalDtype
+
+from scvi.data.anndata._utils import _make_obs_column_categorical
+
+from ._obs_field import CategoricalObsField
+
+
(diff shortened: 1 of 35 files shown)


scvi/data/anndata/fields/_scanvi.py
```diff
@@ -0,0 +1,89 @@
...
+from scvi.data.anndata._utils import _make_obs_column_categorical
+
+from ._obs_field import CategoricalObsField
+
+
+class LabelsWithUnlabeledObsField(CategoricalObsField):
+    """
+    An AnnDataField for labels which include explicitly unlabeled cells.
+
+    Remaps unlabeled category to the final index if present in labels.
```
adamgayoso, line 16: I would just better explain that unlabelled cells are labelled with a special category name that is user defined.

jjhong922: > Labeled and labelled are both correct spellings. Labeled is the preferred spelling in American English. Labelled is the preferred spelling in British English.
Already integrating yourself into British culture I see

PR #1309, opened 2022-01-23
temp pin setuptools so package can install

adamgayoso:
upstream pytorch issue


adamgayoso:

PR #1315, opened 2022-01-27
Field specific `view_anndata_setup`

jjhong922:
Adds a new abstract method to BaseAnnDataField for a field specific state registry summary rich Table. Adds optional argument `hide_state_registries` to the `view_anndata_setup` method which lets you hide the state registries detail if desired (can be verbose with all the state registries).

image

### DIFF
--- a/scvi/data/anndata/fields/_base_field.py
+++ b/scvi/data/anndata/fields/_base_field.py
@@ -109,7 +109,7 @@ class BaseAnnDataField(ABC):
     @abstractmethod
     def view_state_registry(self, state_registry: dict) -> Optional[rich.table.Table]:
         """
-        Returns a rich Table summarizing a state registry produced by this field.
+        Returns a :class:`rich.table.Table` summarizing a state registry produced by this field.
 
         Parameters
         ----------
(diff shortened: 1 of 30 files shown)


scvi/data/anndata/fields/_base_field.py
```diff
@@ -104,7 +105,23 @@ def get_summary_stats(self, state_registry: dict) -> dict:
             This mapping is then combined with the mappings of other fields to make up
             the summary stats mapping.
         """
-        return dict()
+
+    @abstractmethod
+    def view_state_registry(self, state_registry: dict) -> Optional[rich.table.Table]:
+        """
+        Returns a rich Table summarizing a state registry produced by this field.
```
adamgayoso, line 112: reference like `:class:rich.table.Table`?

scvi/data/anndata/fields/_base_field.py
```diff
@@ -104,7 +105,23 @@ def get_summary_stats(self, state_registry: dict) -> dict:
...
+        Parameters
+        ----------
+        state_registry
+            Dictionary returned by :meth:`~scvi.data.anndata.fields.BaseAnnDataField.register_field`.
+            Printed summary should always be a function of information stored in this dictionary.
+
+        Returns
+        -------
+        state_registry_summary
+            Optional rich Table summarizing the ``state_registry``.
```
adamgayoso, line 123: same here

adamgayoso (changes requested):
LGTM other than small comment

PR #1319, opened 2022-01-31
Fix n_obs being set to val size in loss kwargs

adamgayoso:


cane11:
I'm open to this fix. However, the loss will be wrong for validation cells. I prefer defining self.loss_kwargs_val = loss_kwargs and update this with n_obs_validation.

PierreBoyeau (approved):
As I understand it, the issue is that most models do not take `n_obs` as input for the loss. Consequently, the loss kwargs update may not have effects and should be removed. Indeed,the `ElboMetric` class already rescales metrics, through its `n_obs_total`, used [here](https://github.com/YosefLab/scvi-tools/blob/0ffb1959ffcdf65225e2231b041f480c175231d6/scvi/train/_trainingplans.py#L135), and [here](https://github.com/YosefLab/scvi-tools/blob/0ffb1959ffcdf65225e2231b041f480c175231d6/scvi/train/_trainrunner.py#L71).

As a side question for future reference
@adamgayoso you mentioned that the validation loss may not be scaled according to `n_obs_validation`, but to `n_obs_training`. I could not understand that from the code. Could you elaborate?

adamgayoso:
@PierreBoyeau lmk if the comment makes sense

cane11:
I don't see an issue with staying with n_obs_total here. Currently n_obs_validation is 0 for destVI (right?) as it needs to train on every cell for the non-amortized celltype proportions. So we can be consistent with the ELBO here.

adamgayoso:
> I don't see an issue with staying with n_obs_total here. Currently n_obs_validation is 0 for destVI (right?) as it needs to train on every cell for the non-amortized celltype proportions. So we can be consistent with the ELBO here.

There would be a weird issue where validation loss and validation don't agree for an amortized model with global latent variables (n_val > 0), but loss on scale of full data. It's a weird caveat and I think fine for now (tell people to use elbo metric if they have a bayesian model)

PR #1334, opened 2022-02-05
Size Factor Field for SCVI, MULTIVI, SCANVI, TOTALVI

### LINKED ISSUE #1321: Add `size_factor_key` to relevant models
This would be in `setup_anndata` so that users can put a custom size factor in the likelihood. Setting this should make the output activation of `px_rho` to be a softplus, not a softmax

- [x] scVI
- [x] totalVI
- [x] MultiVI
- [x] scANVI

jjhong922:
Fixes #1321


scvi/nn/_base_components.py
```diff
@@ -399,7 +412,7 @@ def forward(
         px_scale = self.px_scale_decoder(px)
         px_dropout = self.px_dropout_decoder(px)
         # Clamp to high value: exp(12) ~ 160000 to avoid nans (computational stability)
-        px_rate = torch.exp(library) * px_scale  # torch.clamp( , max=12)
+        px_rate = size_factor * px_scale  # torch.clamp( , max=12)
```
adamgayoso, line 415: I know it seems weird, but this is a breaking change now because this class assumed library was on the log scale...

scvi/module/_multivae.py
```diff
@@ -412,8 +424,10 @@ def generative(
         p = self.z_decoder_accessibility(decoder_input, batch_index, *categorical_input)
 
         # Expression Decoder
+        if not self.use_size_factor_key:
+            size_factor = torch.exp(libsize_expr)
```
adamgayoso, line 428: I would log the size factor instead of exp the lib size (same in scvi vae)

scvi/module/_vae.py
```diff
@@ -322,8 +336,16 @@ def generative(
         if transform_batch is not None:
             batch_index = torch.ones_like(batch_index) * transform_batch
 
+        if not self.use_size_factor_key:
+            size_factor = torch.exp(library)
```
adamgayoso, line 340: log the size factor instead of exp the library (breaking change)

scvi/utils/_docstrings.py
```diff
@@ -75,6 +75,11 @@ class SetupAnnDataDocstringProcessor(DocstringProcessor):
         key in `adata.obs` for label information. Categories will automatically be converted into integer
         categories and saved to `adata.obs['_scvi_labels']`. If `None`, assigns the same label to all the data."""
 
+    param_size_factor_key = """\
+    size_factor_key
+        key in `adata.obs` for size factor information. Instead of using library size as a size factor, the provided
+        size factor column will be used as a coefficient for a softplussed `px_scale` term."""
```
adamgayoso, line 81: "as offset in the mean of the likelihood. Assumed to be on linear scale."

tests/models/test_models.py
```diff
@@ -471,17 +479,22 @@ def test_new_setup_compat():
     model = SCVI(adata)
     model.view_anndata_setup(hide_state_registries=True)
 
+    registry = adata_manager.registry[_constants._FIELD_REGISTRIES_KEY]
```
adamgayoso, line 482: what is going on here?

adamgayoso (changes requested):

tests/models/test_models.py
jjhong922: oops poor naming. But basically keeping this test working even if new fields are added to SCVI (e.g. size factor) - not supported by backwards compatibility.

adamgayoso (approved):

PR #1340, opened 2022-02-07
Size Factor implementation for SCANVI and TOTALVI

jjhong922:
Addresses remaining two parts of https://github.com/YosefLab/scvi-tools/issues/1321


adamgayoso (approved):
LGTM

PR #1342, opened 2022-02-08
Avoid clobber in setup_anndata using model instance and manager ID back references

jjhong922:
This PR introduces model instance and AnnDataManager ids (uuid4) to avoid clobbering between models referencing the same AnnData object with different parameters.

This is accomplished by adding a model instance ID which is used to create a model instance specific mapping to AnnDataManagers used to initialize given models. The same goes with AnnDataManagers created via `transfer_setup` on different AnnData objects (e.g. via `get_latent_representation()`). This allows model instances to reliably fetch AnnDataManagers created for/by that model instance, rather than the last AnnDataManager created for this AnnData object in this class.

Secondly, it is still possible for an AnnData object to be mutated by another AnnDataManager's `register_fields()` function call. This can result in a situation where a model instance retrieves the correct AnnDataManager object, but the underlying AnnData has changed. To avoid this situation, we replace `source_scvi_uuid` with `scvi_last_manager_uuid` which reports which AnnDataManager instance was used last to setup the AnnData object. Now, one can call `adata_manager.validate()` to check whether the AnnData has been tampered with, and if so rerun `register_fields()` with the original arguments to recreate the desired setup.


scvi/model/base/_base_model.py
```diff
@@ -44,17 +62,19 @@ class BaseModelMetaClass(ABCMeta):
...
 class BaseModelClass(metaclass=BaseModelMetaClass):
     """Abstract class for scvi-tools models."""
 
     def __init__(self, adata: Optional[AnnData] = None):
+        self.id = str(uuid4())  # Used for cls._manager_store keys.
         if adata is not None:
-            self.adata_manager = self.get_anndata_manager(adata, required=True)
-            self.adata = self.adata_manager.adata
+            self.adata = adata
+            self.adata_manager = self.get_latest_anndata_manager(adata, required=True)
```
adamgayoso, line 76: `get_most_recent`?

scvi/model/base/_base_model.py
```diff
@@ -92,13 +112,55 @@ def to_device(self, device: Union[str, int]):
...
+        Registers an :class:`~scvi.data.anndata.AnnDataManager` instance with this model instance.
+
+        Creates a model-instance specific mapping in ``cls._manager_store`` for this
+        :class:`~scvi.data.anndata.AnnDataManager` instance.
+        """
+        adata_id = adata_manager.get_adata_uuid()
+        manager_store_key = _ManagerStoreKey(
+            adata_id=adata_id, model_instance_id=self.id
+        )
+        self._manager_store[manager_store_key] = adata_manager
```
adamgayoso, line 152: here you're using self, and this key will be different than in the `register_manager` call on 139. Is that intended? Shouldn't we overwrite?

scvi/model/base/_base_model.py
```diff
@@ -113,16 +175,70 @@ def get_anndata_manager(
                 )
             return None
 
-        adata_uuid = adata.uns[_SCVI_UUID_KEY]
-        if adata_uuid not in cls.manager_store:
+        adata_id = adata.uns[_SCVI_UUID_KEY]
+        manager_store_key = _ManagerStoreKey(adata_id=adata_id, model_instance_id=None)
```
adamgayoso, line 179: why is model instance id none here?

adamgayoso: oh wait I get it now... this is why you don't overwrite the key earlier?

scvi/model/base/_base_model.py
adamgayoso: I got it as I continued the review, there has to be a better way we can communicate what's going on. Maybe we have two stores so the code is more readable?

scvi/model/base/_base_model.py
```diff
@@ -229,16 +345,11 @@ def _validate_anndata(
...
-        ):
-            logger.info(
-                "Input AnnData requires setup with AnnData the model was initialized with. "
-                "Attempting to transfer setup with initial AnnData."
+            self._register_manager_for_instance(
+                self.adata_manager.transfer_setup(adata)
             )
-            self.register_manager(self.adata_manager.transfer_setup(adata))
+        else:
+            adata_manager.validate()
```
adamgayoso, line 352: the manager here is the correct one for the instance and this line will replay registration? If so let's comment that

scvi/model/base/_base_model.py
jjhong922: yes they have two different meanings. I tried to explain it here but agree its not straightforward https://github.com/YosefLab/scvi-tools/pull/1342/files#diff-41fbf1b79b096021b95dc02658ce6c30c499f58e43904cbb5a986b99ec02f747R41-R51

scvi/model/base/_base_model.py
```diff
@@ -40,21 +40,33 @@ class BaseModelMetaClass(ABCMeta):
...
+    ``cls._setup_adata_manager_store`` maps from AnnData object UUIDs to :class:`~scvi.data.anndata.AnnDataManager` instances.
+    This mapping is populated everytime ``cls.setup_anndata()`` is called.
+    ``cls._per_isntance_manager_store`` maps from model instance UUIDs to AnnData UUID::class:`~scvi.data.anndata.AnnDataManager` mappings.
+    These :class:`~scvi.data.anndata.AnnDataManager` instances are tied to a single model instance and populated either
+    during model initialization or after running ``self._validate_anndata()``.
     """
 
     def __init__(cls, name, bases, dct):
-        cls.manager_store = dict()
+        cls._setup_adata_manager_store: Dict[
```
adamgayoso, line 52: correct me if I'm wrong, but this manager store will only ever receive one anndata now, the one used to initialize the model?

adamgayoso: if this is the case, should we make it not a dict?

adamgayoso: actually this is general enough...

scvi/data/anndata/_manager.py
```diff
@@ -97,23 +100,13 @@ def _assign_uuid(self):
         scvi_uuid = self.adata.uns[_constants._SCVI_UUID_KEY]
         self._registry[_constants._SCVI_UUID_KEY] = scvi_uuid
 
-    def _assign_source_uuid(self, source_registry: Optional[dict]):
+    def _assign_most_recent_manager_uuid(self):
```
adamgayoso, line 103: should we be more consistent with most recent vs latest?

scvi/model/base/_base_model.py
```diff
@@ -92,13 +104,55 @@ def to_device(self, device: Union[str, int]):
...
+        """
+        if self.id not in self._per_instance_manager_store:
+            self._per_instance_manager_store[self.id] = dict()
+
+        adata_id = adata_manager.get_adata_uuid()
+        instance_manager_store = self._per_instance_manager_store[self.id]
+        instance_manager_store[adata_id] = adata_manager
+
+    @classmethod
+    def get_latest_anndata_manager(
```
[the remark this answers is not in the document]
jjhong922: yea i'll make it private, it's used in gimVI but I think once we do mudata integration we should be able to remove it altogether

scvi/model/base/_base_model.py
jjhong922: it stores it for the whole class. this allows for situations where an AnnData is setup once and used to create two separate models.

adamgayoso (approved):

PR #1350, opened 2022-02-13
Always add unlabeled category for scanvi

adamgayoso:
Even when the labels don't contain the unlabeled category. This fixes an edge case with scarches (see added test case).

I also cleared some unused code.

### DIFF
--- a/scvi/data/anndata/_utils.py
+++ b/scvi/data/anndata/_utils.py
@@ -121,7 +121,6 @@ def _make_obs_column_categorical(
     column_key,
     alternate_column_key,
     categorical_dtype=None,
-    return_mapping=False,
 ):
     """
     Makes the data in column_key in obs all categorical.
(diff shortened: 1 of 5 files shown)


scvi/data/anndata/_utils.py
```diff
@@ -146,16 +145,6 @@ def _make_obs_column_categorical(
         )
     adata.obs[alternate_column_key] = codes
 
-    if not return_mapping:
```
jjhong922, line 149: thanks

jjhong922 (approved):
looks great, really simplifies a lot of code

scvi/data/anndata/fields/_scanvi.py
```diff
@@ -48,24 +48,25 @@ def _remap_unlabeled_to_final_category(
...
-                self.attr_key,
-                categorical_dtype=cat_dtype,
-                return_mapping=True,
-            )
-            remapped = True
-        else:
-            remapped = False
+        # could be in mapping in transfer case
+        elif self._unlabeled_category not in mapping:
+            # no unlabeled category, so no need to remap
```
adamgayoso, line 53:
```suggestion
```

PR #1351, opened 2022-02-13
Adata and manager as properties

adamgayoso:
Prevents arbitrary setting of sensitive model attributes.

Should we make registry_ and summary_stats as properties that pull from the current adata_manager? I don't think it matters?

### DIFF
--- a/scvi/model/base/_base_model.py
+++ b/scvi/model/base/_base_model.py
@@ -96,7 +96,7 @@ class BaseModelClass(metaclass=BaseModelMetaClass):
 
     @property
     def adata_manager(self) -> AnnDataManager:
-        """AnnDataManager instance associated with self.adata."""
+        """Manager instance associated with self.adata."""
         return self._adata_manager
 
     def to_device(self, device: Union[str, int]):
--- a/scvi/model/base/_base_model.py
+++ b/scvi/model/base/_base_model.py
@@ -64,8 +64,8 @@ class BaseModelClass(metaclass=BaseModelMetaClass):
     def __init__(self, adata: Optional[AnnData] = None):
         self.id = str(uuid4())  # Used for cls._manager_store keys.
         if adata is not None:
-            self.adata = adata
-            self.adata_manager = self._get_most_recent_anndata_manager(
+            self._adata = adata
+            self._adata_manager = self._get_most_recent_anndata_manager(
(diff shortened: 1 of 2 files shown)


scvi/model/base/_base_model.py
```diff
@@ -81,6 +81,24 @@ def __init__(self, adata: Optional[AnnData] = None):
...
+        """Data attached to model instance."""
+        return self._adata
+
+    @adata.setter
+    def adata(self, adata: AnnData):
+        if adata is None:
+            raise ValueError("adata cannot be None.")
+        self._validate_anndata(adata)
+        self._adata = adata
+        self._adata_manager = self.get_anndata_manager(adata)
```
jjhong922, line 95: This needs to reset self.registry_ and self.summary_stats_. Should have a test corresponding to this as well.

jjhong922: Would add `required=True` here since we never want None returned here (even if validate shouldnt let that happen anyways)

jjhong922 (approved):
Fine with this addition, although not sure how much we want to encourage this mode of model manipulation

scvi/model/base/_base_model.py
adamgayoso: I will add this

adamgayoso:
We shouldn't encourage it, but we are making it harder to do.

We might also consider adding more properties (self.module, etc.) so people can't easily overwrite them

scvi/model/base/_base_model.py
jjhong922: still need required flag

tests/models/test_models.py
```diff
@@ -280,6 +280,25 @@ def test_scvi_sparse(save_path):
...
+def test_setting_adata_attr():
+    n_latent = 5
+    adata = synthetic_iid()
+    SCVI.setup_anndata(adata, batch_key="batch")
+    model = SCVI(adata, n_latent=n_latent)
+    model.train(1, train_size=0.5)
+
+    adata2 = synthetic_iid()
+    model.adata = adata2
+    model.get_latent_representation()
```
jjhong922, line 292: check this is equivalent to doing `model.get_latent_representation(adata2)` and not `model.get_latent_representation(adata)`.

PR #1355, opened 2022-02-15
`view_setup_args` for saved model paths

jjhong922:
image

Necessary to replace former `view_anndata_setup` functionality for saved model paths. Handy when working with transfer learning.


adamgayoso (changes requested):
lgtm after one small clarification

adamgayoso (approved):

PR #1356, opened 2022-02-15
Distributions outb

PierreBoyeau:


scvi/model/_autozi.py
```diff
@@ -235,19 +233,18 @@ def get_marginal_ll(
                 )
 
                 # Log-probabilities
-                log_prob_sum = torch.zeros(qz_m.shape[0]).to(self.device)
```
jjhong922, line 238: has this been tested on a gpu? I remember making this change bc on a GPU the components would live one different devices

scvi/module/_vae.py
```diff
@@ -360,8 +353,20 @@ def generative(
 
         px_r = torch.exp(px_r)
 
+        if self.gene_likelihood == "zinb":
```
jjhong922, line 356: are these related changes?

scvi/module/_totalvae.py
```diff
@@ -701,7 +692,7 @@ def marginal_ll(self, tensors, n_mc_samples):
             reconst_loss_protein = reconst_loss["reconst_loss_protein"]
 
             # Log-probabilities
-            log_prob_sum = torch.zeros(qz_m.shape[0]).to(self.device)
+            log_prob_sum = torch.zeros(ql.loc.shape[0]).to(self.device)
```
jjhong922, line 695: qz.loc?

scvi/module/_amortizedlda.py
```diff
@@ -200,7 +200,9 @@ def forward(
         with pyro.plate(
             "cells", size=n_obs or self.n_obs, subsample_size=x.shape[0]
         ), poutine.scale(None, kl_weight):
-            cell_topic_posterior_mu, cell_topic_posterior_sigma, _ = self.encoder(x)
+            cell_topic_posterior, _ = self.encoder(x)
+            cell_topic_posterior_mu = cell_topic_posterior.loc
+            cell_topic_posterior_sigma = 2.0 * cell_topic_posterior.scale.log()
```
jjhong922, line 205: What's the reason for this? For equivalence to what it was before wouldn't it just be `cell_topic_posterior.scale**2`?

scvi/model/_autozi.py
PierreBoyeau: let me double check, but since the outputs of `forward` should be on the GPU, then all quantities should be on the GPU

scvi/module/_amortizedlda.py
PierreBoyeau: You're right. I thought that `cell_topic_posterior_sigma` was the log-variance here. 
However, on an unrelated note to this PR, this means that in this model the posterior variance is lower bounded by `Softplus(0.0) ~ 0.6931` though. Why is that?

scvi/module/_totalvae.py
PierreBoyeau: Updated, resolved

scvi/module/_vae.py
PierreBoyeau: Yes! I feel that returning the likelihood distribution in `generative` simplifies `loss`, and can be used elsewhere (e.g., in `sample`), without having to reconstruct the distribution.
However, this may not be a good idea for all models, as some have custom likelihoods that are not coded as `torch.distributions.Distribution` (e.g., MultiVI).
Do you think I should remove this?

scvi/module/_amortizedlda.py
jjhong922: Yea that's a good point. I think was using  FC Layers at one point then switched it out so I had assumed the output var value could be negative. Even with it being constrained to be positive, I think I had to do this because Pyro would try stepping into a negative value and plugging it into the Normal distribution, so softplussing it prevented that. I can make an issue for this

scvi/model/_autozi.py
PierreBoyeau: Hi again @jjhong922 , I confirm that this code runs fine on a GPU (through `test_autozi`)

scvi/module/_vae.py
jjhong922: ah I see, no I think it makes sense since its parameterized as such w/ the gene_likelihood arg.

scvi/nn/_base_components.py
```diff
@@ -293,8 +293,9 @@ def forward(self, x: torch.Tensor, *cat_list: int):
         q = self.encoder(x, *cat_list)
         q_m = self.mean_encoder(q)
         q_v = self.var_activation(self.var_encoder(q)) + self.var_eps
-        latent = self.z_transformation(reparameterize_gaussian(q_m, q_v))
-        return q_m, q_v, latent
+        dist = Normal(q_m, q_v.sqrt())
+        latent = self.z_transformation(dist.rsample())
+        return dist, latent
```
adamgayoso, line 298: this is a breaking change, some people might rely on this code. I would add a `return_dist` option, set it False and then make it true in each module class

scvi/nn/_base_components.py
```diff
@@ -594,9 +595,9 @@ def forward(self, x: torch.Tensor, head_id: int, *cat_list: int):
 
         q_m = self.mean_encoder(q)
         q_v = torch.exp(self.var_encoder(q))
-        latent = reparameterize_gaussian(q_m, q_v)
-
-        return q_m, q_v, latent
+        dist = Normal(q_m, q_v.sqrt())
+        latent = dist.rsample()
+        return dist, latent
```
adamgayoso, line 600: I'm fine with this breaking change, but maybe want to be consistent?

scvi/nn/_base_components.py
```diff
@@ -1026,4 +1031,4 @@ def forward(self, data: torch.Tensor, *cat_list: int):
         untran_latent["z"] = untran_z
         untran_latent["l"] = log_library_gene
 
-        return qz_m, qz_v, ql_m, ql_v, latent, untran_latent
+        return q_z, q_l, latent, untran_latent
```
adamgayoso, line 1034: this breaking change is fine because this class is really only used by totalvi

scvi/module/_vae.py
```diff
@@ -360,8 +353,20 @@ def generative(
...
+            px_latents = ZeroInflatedNegativeBinomial(
+                mu=px_rate, theta=px_r, zi_logits=px_dropout
+            )
+        elif self.gene_likelihood == "nb":
+            px_latents = NegativeBinomial(mu=px_rate, theta=px_r)
+        elif self.gene_likelihood == "poisson":
+            px_latents = Poisson(px_rate)
         return dict(
-            px_scale=px_scale, px_r=px_r, px_rate=px_rate, px_dropout=px_dropout
+            px_latents=px_latents,
```
adamgayoso, line 365: I think `px` is better here to be consistent that it's a distribution and not a parameter thereof

scvi/module/_vae.py
adamgayoso: but aren't there more changes to be done in this file? E.g., can we remove `get_reconstruction_loss`?

adamgayoso: and if this is the case, should we do it globally? And finally, with our negative binomial, not each parameter is identifiable from the distribution object, namely, `px_scale`

adamgayoso (changes requested):

scvi/nn/_base_components.py
PierreBoyeau: gotcha

scvi/nn/_base_components.py
PierreBoyeau: what do you mean exactly?

scvi/module/_vae.py
PierreBoyeau: we can remove the `get_reconstruction_loss`. A solution could be to feed `px_scale` and `library` in addition to `px_rate` to some `Distribution` object, that could then be used later on.
This mostly make sense for models which rely on `Distribution` to compute the reconstruction loss though, and I am not sure what we should do for other models.

scvi/module/_vae.py
PierreBoyeau: i agree, i am on it

scvi/nn/_base_components.py
PierreBoyeau: i understand, on it

scvi/module/_totalvae.py
```diff
@@ -480,23 +480,21 @@ def inference(
...
         )
+
         z = latent["z"]
         untran_z = untran_latent["z"]
         untran_l = untran_latent["l"]
         if not self.use_observed_lib_size:
             library_gene = latent["l"]
 
         if n_samples > 1:
-            qz_m = qz_m.unsqueeze(0).expand((n_samples, qz_m.size(0), qz_m.size(1)))
```
jjhong922, line 493: who wrote this lol

jjhong922 (commented):
looks good to me, needs to be added to release note

scvi/module/_vae.py
PierreBoyeau: Done.

scvi/module/_totalvae.py
PierreBoyeau: I don't know, it's very old code :D

scvi/distributions/_negative_binomial.py
```diff
@@ -236,6 +238,25 @@ def _gamma(theta, mu):
...
+    Poisson distribution.
+
+    Parameters
+    ----------
+    rate
+        rate of the Poisson distribution.
+    validate_args
+        whether to validate input.
+    scale
+        Normalized mean expression of the distribution.
```
adamgayoso, line 252: it should be noted that this is not used in any computations, same in the other distributions

adamgayoso: like it's only used for caching the value basically for later

adamgayoso (changes requested):

scvi/model/base/_rnamixin.py
```diff
@@ -483,9 +483,9 @@ def get_likelihood_parameters(
                 inference_kwargs=inference_kwargs,
                 compute_loss=False,
             )
-            px_r = generative_outputs["px_r"]
-            px_rate = generative_outputs["px_rate"]
-            px_dropout = generative_outputs["px_dropout"]
+            px_r = generative_outputs["px"].theta
+            px_rate = generative_outputs["px"].mu
+            px_dropout = generative_outputs["px"].zi_probs
```
adamgayoso, line 488: put px in a variable so this code is not repeated 3 times?

scvi/module/_autozivae.py
```diff
@@ -365,32 +371,28 @@ def loss(
...
         # Parameters for z latent distribution
-        qz_m = inference_outputs["qz_m"]
-        qz_v = inference_outputs["qz_v"]
-        px_rate = generative_outputs["px_rate"]
-        px_r = generative_outputs["px_r"]
-        px_dropout = generative_outputs["px_dropout"]
+        qz = inference_outputs["qz"]
+        px_rate = generative_outputs["px"].mu
+        px_r = generative_outputs["px"].theta
+        px_dropout = generative_outputs["px"].zi_logits
```
adamgayoso, line 377: repated code

scvi/model/_autozi.py
```diff
@@ -212,13 +212,11 @@ def get_marginal_ll(
...
                 inf_outputs, gen_outputs, _ = self.module.forward(tensors)
 
-                px_r = gen_outputs["px_r"]
-                px_rate = gen_outputs["px_rate"]
-                px_dropout = gen_outputs["px_dropout"]
-                qz_m = inf_outputs["qz_m"]
-                qz_v = inf_outputs["qz_v"]
+                px_r = gen_outputs["px"].theta
+                px_rate = gen_outputs["px"].mu
+                px_dropout = gen_outputs["px"].zi_logits
```
adamgayoso, line 217: repeated code

adamgayoso (changes requested):

adamgayoso:
@PierreBoyeau can you also add a new v0.17.0 release note and get it started in this PR?

PierreBoyeau:
Sure!

scvi/distributions/_negative_binomial.py
PierreBoyeau: Done!

scvi/distributions/_negative_binomial.py
```diff
@@ -236,6 +238,21 @@ def _gamma(theta, mu):
...
 
 
+class Poisson(PoissonTorch):
+    r"""
+    Poisson distribution.
+
+    Parameters
+    ----------
+    rate
+        rate of the Poisson distribution.
```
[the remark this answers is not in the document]
PierreBoyeau: Done!

PierreBoyeau:
i addressed the points you raised @jjhong922 @adamgayoso

docs/release_notes/v0.17.0.md
```diff
@@ -3,6 +3,7 @@
 ## Changes
 
 - Experimental MuData support for {class}`~scvi.model.TOTALVI` via the method {meth}`~scvi.model.TOTALVI.setup_mudata`. For several of the existing `AnnDataField` classes, there is now a MuData counterpart with an additional `mod_key` argument used to indicate the modality where the data lives (e.g. {class}`~scvi.data.fields.LayerField` to {class}`~scvi.data.fields.MuDataLayerField`). These modified classes are simply wrapped versions of the original `AnnDataField` code via  the new {method}`scvi.data.fields.MuDataWrapper` method [#1474].
+- Modification of the `generative` method's outputs to return prior and likelihood properties as `torch.Distribution` objects. Concerned modules are `_amortizedlda.py`, `_autozivae.py`, `multivae.py`, `_peakvae.py`, `_scanvae.py`, `_vae.py`, and `_vaec.py`. This allows facilitating the manipulation of these distributions for model training and inference [#1356].
```
jjhong922, line 6: Use the api doc notation e.g. {class}`~scvi.module.VAE` instead of `_vae.py`

scvi/distributions/_negative_binomial.py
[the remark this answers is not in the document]
jjhong922: sorry wasnt clear, just meant to expand on the definitions which looks good now. You can remove ": optional" because sphinx takes care of that for us

jjhong922 (approved):
final minor comments, but i think its good to go

adamgayoso (approved):

scvi/distributions/_negative_binomial.py
```diff
@@ -236,6 +238,33 @@ def _gamma(theta, mu):
...
+    validate_args : optional
+        whether to validate input.
+    scale : optional
+        Normalized mean expression of the distribution.
+        This optional parameter is not used in any computations, but allows to store
+        normalization expression levels.
```
adamgayoso, line 254: @PierreBoyeau @jjhong922 can we fix this?

jjhong922: are you talking about the docs or the optional parameter here? I can remove the :optional in the docs now

PR #1367, opened 2022-02-22
JaxSCVI

adamgayoso:
- Implements a MVP Jax-based scVI model.
- We can continue to add functionality to this during 0.15.x patches and so on.
- Supports GPU and CPU, `get_latent_representation`
- No save/load
- Removes Numba, reimplements nonnegative integer check with jax

### DIFF
--- a/scvi/_settings.py
+++ b/scvi/_settings.py
@@ -216,7 +216,7 @@ class ScviConfig:
         if value is False:
             os.environ["XLA_PYTHON_CLIENT_PREALLOCATE"] = "false"
         elif isinstance(value, float):
-            if value > 0.99 or value < 0:
+            if value >= 1 or value <= 0:
                 raise ValueError("Need to use a value between 0 and 1")
             # format is ".XX"
             os.environ["XLA_PYTHON_CLIENT_MEM_FRACTION"] = str(value)[1:4]
--- a/scvi/_settings.py
+++ b/scvi/_settings.py
@@ -57,7 +57,6 @@ class ScviConfig:
         jax_preallocate_gpu_memory: bool = False,
     ):
 
-        self.verbosity = verbosity
         self.seed = seed
         self.batch_size = batch_size
         if progress_bar_style not in ["rich", "tqdm"]:
@@ -68,6 +67,7 @@ class ScviConfig:
         self.dl_pin_memory_gpu_training = dl_pin_memory_gpu_training
         self._num_threads = None
         self.jax_preallocate_gpu_memory = jax_preallocate_gpu_memory
+        self.verbosity = verbosity
 
     @property
     def batch_size(self) -> int:
@@ -178,7 +178,9 @@ class ScviConfig:
             console = Console(force_terminal=True)
             if console.is_jupyter is True:
                 console.is_jupyter = False
-            ch = RichHandler(show_path=False, console=console, show_time=False)
+            ch = RichHandler(
+                level=level, show_path=False, console=console, show_time=False
+            )
             formatter = logging.Formatter("%(message)s")
             ch.setFormatter(formatter)
             scvi_logger.addHandler(ch)
@@ -192,7 +194,7 @@ class ScviConfig:
         This is useful if piping outputs to a file.
         """
         scvi_logger.removeHandler(scvi_logger.handlers[0])
-        ch = RichHandler(show_path=False, show_time=False)
+        ch = RichHandler(level=self._verbosity, show_path=False, show_time=False)
         formatter = logging.Formatter("%(message)s")
         ch.setFormatter(formatter)
         scvi_logger.addHandler(ch)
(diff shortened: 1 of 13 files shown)


scvi/data/_utils.py
```diff
@@ -194,24 +196,18 @@ def _check_nonnegative_integers(
...
-        boolean(float64),
-    ],
-    target="parallel",
-    cache=True,
-)
-def _is_not_count(d):
-    return d < 0 or d % 1 != 0
+    inds = np.random.choice(len(data), size=(n_to_check,))
+    check = jax.device_put(data.flat[inds], device=jax.devices("cpu")[0])
+    negative, non_integer = _is_count(check)
```
adamgayoso, line 201:
```suggestion
    negative, non_integer = _is_not_count_val(check)
```

scvi/data/_utils.py
```diff
@@ -194,24 +196,18 @@ def _check_nonnegative_integers(
...
+    return not (negative or non_integer)
+
+
+@jax.jit
+def _is_count(data: jnp.ndarray):
```
adamgayoso, line 206:
```suggestion
def _is_not_count_val(data: jnp.ndarray):
```

scvi/module/_jaxvae.py
```diff
@@ -0,0 +1,144 @@
...
+    n_input: int
+    n_latent: int
+    n_hidden: int
+    dropout_rate: int
+    is_training: Optional[bool] = None
+
+    @nn.compact
+    def __call__(self, inputs, is_training):
+
+        is_training = nn.merge_param("is_training", self.is_training, is_training)
```
jjhong922, line 32: from what i understand this is meant to avoid the dataclass `is_training` to not conflict w the arg in `__call__`. However, by making it an arg it seems like you want it to be always passed on `__call__`. Is there a point to having the dataclass attr?

scvi/module/_jaxvae.py
```diff
@@ -0,0 +1,144 @@
...
+
+class FlaxEncoder(nn.Module):
+    n_input: int
+    n_latent: int
+    n_hidden: int
+    dropout_rate: int
+    is_training: Optional[bool] = None
+
+    @nn.compact
+    def __call__(self, inputs, is_training):
```
jjhong922, line 30: are typings ok in these call fcns?

jjhong922 (approved):
overall looks good, and helped me understand what needs to be solved wrt the API. Some minor comments/questions

scvi/_settings.py
```diff
@@ -185,10 +194,35 @@ def reset_logging_handler(self):
...
+        """
+        return self._jax_gpu
+
+    @jax_preallocate_gpu_memory.setter
+    def jax_preallocate_gpu_memory(self, value: Union[float, bool]):
+        # see https://jax.readthedocs.io/en/latest/gpu_memory_allocation.html#gpu-memory-allocation
+        if value is False:
+            os.environ["XLA_PYTHON_CLIENT_PREALLOCATE"] = "false"
+        elif isinstance(value, float):
+            if value > 0.99 or value < 0:
```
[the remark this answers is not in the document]
jjhong922: \>=1

scvi/module/_jaxvae.py
adamgayoso: this is a good question -- let's address this in future PRs as I don't know!

scvi/model/_jaxscvi.py
```diff
@@ -0,0 +1,342 @@
...
+            params=params,
+            tx=optax.adamw(lr, eps=0.01, weight_decay=1e-6),
+            batch_stats=batch_stats,
+        )
+
+        @jax.jit
+        def train_step(state, array_dict, rngs, kl_weight=1) -> jnp.ndarray:
+            rngs = {k: random.split(v)[1] for k, v in rngs.items()}
+
+            # batch stats can't be passed here
```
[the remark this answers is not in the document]
adamgayoso: yes

scvi/model/_jaxscvi.py
```diff
@@ -0,0 +1,342 @@
...
+
+    def load(self):
+        raise NotImplementedError
+
+    def to_device(self):
+        raise NotImplementedError
+
+    @property
+    def device(self):
+        pass
```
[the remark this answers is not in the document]
adamgayoso: well, we can actually implement this in the future, just move the pytree of jax params to the device

scvi/model/_jaxscvi.py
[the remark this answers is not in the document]
jjhong922: should be careful bc if the inputs are the same but batch_stats is changed (probably wont happen but still), it might return the cached output. is this the flax recommended way?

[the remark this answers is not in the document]
adamgayoso: this is what I found in flax's examples

scvi/module/_jaxvae.py
adamgayoso: fixed

PR #1373, opened 2022-02-25
fix scanvi from_scvi_model

adamgayoso:
user shouldn't have to specify labels key in the scvi model upfront


jjhong922 (approved):

jjhong922 (commented):
Add this change to the 0.15.0 release note w/ a breaking change note.

PR #1379, opened 2022-02-26
indices handling for non-amortized models

adamgayoso:
Strange that it was even working before, but we were using the count data to select the index!

Let's rerun the destvi tutorials again after this gets in.

### DIFF
--- a/scvi/module/_mrdeconv.py
+++ b/scvi/module/_mrdeconv.py
@@ -344,7 +344,7 @@ class MRDeconv(BaseModuleClass):
         gamma_select = gamma_ind[
             :, y_torch, torch.arange(ind_x.shape[0])
         ].T  # minibatch_size, n_latent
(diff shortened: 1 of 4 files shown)


scvi/module/_mrdeconv.py
```diff
@@ -140,7 +140,7 @@ def _get_inference_input(self, tensors):
 
     def _get_generative_input(self, tensors, inference_outputs):
         x = tensors[REGISTRY_KEYS.X_KEY]
-        ind_x = tensors[REGISTRY_KEYS.X_KEY].long()
+        ind_x = tensors[REGISTRY_KEYS.INDICES_KEY].long().ravel()
```
adamgayoso, line 143: FYI @jjhong922 this was the big change, all the other changes are to prevent this from happening again

jjhong922 (approved):
oops thanks for finding this

PR #1385, opened 2022-03-01
multi particles

PierreBoyeau:


scvi/model/_jaxscvi.py
```diff
@@ -301,14 +303,18 @@ def get_latent_representation(
 
         @jax.jit
         def _get_val(array_dict):
-            out = self.bound_module(array_dict)
-            return out.qz.mean
+            out = self.bound_module(array_dict, n_samples=mc_samples)
+            if give_mean:
```
jjhong922, line 307: We should be careful when using variables within a jit compiled function like this. I think perhaps a better practice here would to be to include `give_mean` in the function signature then addition as a static arg if necessary to get it jit compiled. The reason is because if `give_mean` is ever changed and `_get_val` is called with the same `array_dict` it will return a cached value instead of modifying the output as appropriate.

This is just a recommendation for overall code style going forward with the jax code. Nothing forces you to do this in jax, so it's important we're aware of how jit affects the behavior. (it's not so important here, but the we should still stick to the principle!)

scvi/module/_jaxvae.py
```diff
@@ -73,6 +73,9 @@ def __call__(self, z: jnp.ndarray, batch: jnp.ndarray, is_training: bool):
         )
         h = nn.relu(h)
         h = nn.Dropout(self.dropout_rate)(h, deterministic=not is_training)
+        if h.ndim == 3:
```
jjhong922, line 76: this is an interesting choice haha.. wonder if there is a more explicit way to bake this into the decoder

adamgayoso: @PierreBoyeau  let's instead do the same thing I do at line 69, as this is something that can be done with jax but not pytorch efficiently

scvi/model/_jaxscvi.py
adamgayoso: agree

adamgayoso (changes requested):

adamgayoso (approved):

PR #1386, opened 2022-03-01
scvi-tools NB parameterization for numpyro

adamgayoso:
- Adds `scvi.distributions.JaxNegativeBinomialMeanDisp`

Here I mirror the log prob of our pytorch NB dist.

This restores the performance on scIB metrics for JaxSCVI:

PyTorch Metrics:

Screen Shot 2022-03-01 at 8 43 52 AM

Jax Metrics:

Screen Shot 2022-03-01 at 8 44 25 AM

Without this dist, integration metrics were worse.

The remaining differences are either stochasticity or the way our Dense layer is initialized.


adamgayoso:
@jjhong922 I figure we should work in only the two new files until we figure out where things should go

jjhong922 (approved):

PR #1390, opened 2022-03-01
stage release notes for patch

adamgayoso:
Once we get this in, let's do the same for v0.16.0. This should also be backported (the label I added should do it automatically)


adamgayoso:
@meeseeksdev backport to 0.15.x

PR #1406, opened 2022-03-07
Implement and use `JaxBaseModuleClass`

adamgayoso:


scvi/model/_jaxscvi.py
```diff
@@ -132,6 +132,7 @@ def module(self):
     def train(
         self,
         max_epochs: Optional[int] = None,
+        check_val_every_n_epoch: Optional[int] = None,
```
jjhong922, line 135: should we always append new kwargs to the end of the signature unless it really doesnt make sense? Just to avoid situations when they are used as args.

scvi/model/_jaxscvi.py
```diff
@@ -191,32 +198,33 @@ def key(i: int):
...
-            apply_fn=module.apply,
+            apply_fn=train_module.apply,
             params=params,
             tx=optax.adamw(lr, eps=0.01, weight_decay=1e-6),
             batch_stats=batch_stats,
         )
 
         @jax.jit
-        def train_step(state, array_dict, rngs, kl_weight=1) -> jnp.ndarray:
+        def train_step(state, array_dict, rngs, kwargs):
```
jjhong922, line 213: is **kwargs not allowed in jit?

scvi/module/base/_base_module.py
```diff
@@ -36,14 +38,35 @@ class LossRecorder:
 
     def __init__(
         self,
-        loss: Union[Dict[str, torch.Tensor], torch.Tensor],
+        loss: Union[
+            Dict[str, Union[torch.Tensor, jnp.ndarray]],
```
jjhong922, line 42: I should really address https://github.com/scverse/scvi-tools/issues/1131

scvi/module/base/_base_module.py
```diff
@@ -69,19 +92,19 @@ def _get_dict_sum(dictionary):
         return total
 
     @property
-    def loss(self) -> torch.Tensor:
```
jjhong922, line 72: Might be overkill, but at the loss recorder level it would be nice if the values were just stored agnostic to the storage type (tensor vs jnp array).

scvi/module/base/_base_module.py
```diff
@@ -343,3 +366,133 @@ def create_predictive(
     def forward(self, *args, **kwargs):
         """Passthrough to Pyro model."""
         return self.model(*args, **kwargs)
+
+
+class JaxBaseModuleClass(linen.Module):
+    """Abstract class for Jax-based scvi-tools modules."""
+
+    def __init__(*args, **kwargs):
+        super().__init__(**kwargs)
```
jjhong922, line 375: why doesnt *args get passed here?

scvi/module/base/_base_module.py
```diff
@@ -343,3 +366,133 @@ def create_predictive(
...
+        In the case of variational inference, this function will perform steps related to
+        computing variational distribution parameters. In a VAE, this will involve running
+        data through encoder networks.
+
+        This function should return a dictionary with str keys and :class:`~jnp.ndarray` values.
+        """
+        pass
+
+    @abstractmethod
+    def generative(self, *args, **kwargs) -> dict:
```
jjhong922, line 477: For whole file, type the Dict return values

jjhong922 (commented):
Overall looks really good. JaxBaseModule is nicely consistent with BaseModule. It is a shame that beyond looking similar we can't tie together the pytorch and jax components (probably mostly bc of pytorch lightning). It would be cool if everything was agnostic to the backend with the exception of the very lowest level of calling flax modules.

scvi/model/_jaxscvi.py
adamgayoso: yes, but this whole API is experimental. Right now for scvi et al. this is part of **kwargs to the trainer. I think this is fine for now.

scvi/module/base/_base_module.py
adamgayoso: yes later :)

scvi/module/base/_base_module.py
adamgayoso: because this is some randomly generated vscode fill in

scvi/module/base/_base_module.py
```diff
@@ -343,3 +358,155 @@ def create_predictive(
...
+
+        This function uses the outputs of the inference and generative functions to compute
+        a loss. This many optionally include other penalty terms, which should be computed here.
+
+        This function should return an object of type :class:`~scvi.module.base.LossRecorder`.
+        """
+        pass
+
+
+def _generic_forward(
```
jjhong922, line 481: where else is this used?

adamgayoso:
@jjhong922 the error here is due to amortized lda save load error precision

scvi/module/base/_base_module.py
adamgayoso: ? Only inside this file, both by jax call and pytorch forward

jjhong922: Oh ok missed the pytorch one. Just wanted to warn against refactoring too early.

jjhong922:
@adamgayoso just rerunning it to see if its truly a flake. But I suppose we can manually increase the margin on the test too

PR #1415, opened 2022-03-10
autoclass template, document shared attrs in `BaseModel`

adamgayoso:
Doc build takes 360 seconds vs 900 from before. Classes take one page now (methods on right side content bar)

Example here: https://scvi--1415.org.readthedocs.build/en/1415/api/reference/scvi.model.TOTALVI.html#scvi.model.TOTALVI

### DIFF
--- a/docs/_templates/autosummary/class.rst
+++ b/docs/_templates/autosummary/class.rst
@@ -1,3 +1,6 @@
+.. role:: hidden
+    :class: hidden-section
+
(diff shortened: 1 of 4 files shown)


jjhong922 (commented):
image
As communicated on zulip, would like to see the right sidebar have a hierarchy with correct ordering

jjhong922 (approved):

PR #1416, opened 2022-03-10
Transfer when copy detected

### LINKED ISSUE #1409: `_validate_anndata` should deal with AnnData
If somebody subsets an anndata and makes a copy, it will grab the wrong AnnDataManager. Validate_anndata should handle this by copying in place and attempting transfer.

To copy in place: https://github.com/scverse/scvi-tools/blob/a5a7c596d1057a9fcd46dafd2a935b8ebbfbbc8d/scvi/model/base/_base_model.py#L358

Where to address: https://github.com/scverse/scvi-tools/blob/2472b76572672fb9c98755522fa3c7326a946444/scvi/model/base/_base_model.py#L209-L213

jjhong922:
Fixes #1409


adamgayoso (approved):
LGTM after release note

PR #1431, opened 2022-03-14
Fix warning for old model version when using scArches

grst:
Loading a model trained with 0.15.0 currently triggers `"Query integration should be performed using models trained with version >= 0.8"`.


adamgayoso:
Thanks!

PR #1434, opened 2022-03-14
add arches query data prep

adamgayoso:
This adds a new static method to the `ArchesMixin` that will pad and subset query data appropriately given a reference model (instance or path). Arches works well if not too many 0 genes are padded, so a warning is issued if too many are added.


scvi/model/base/_archesmixin.py
```diff
@@ -229,3 +300,17 @@ def requires_grad(key):
             par.requires_grad = True
         else:
             par.requires_grad = False
+
+
+def _get_loaded_data(reference_model, device=None):
```
jjhong922, line 305: typings

scvi/model/base/_archesmixin.py
```diff
@@ -155,6 +154,78 @@ def load_query_data(
...
+
+        This function will return a new AnnData object with padded zeros
+        for missing features, as well as correctly sorted features.
+
+        Parameters
+        ----------
+        adata
+            AnnData organized in the same way as data used to train model.
+            It is not necessary to run setup_anndata,
+            as AnnData is validated against the saved `scvi` setup dictionary.
```
jjhong922, line 175: Fix comment, refer to the registry instead (same from where it was copied from)

scvi/model/base/_archesmixin.py
```diff
@@ -155,6 +154,78 @@ def load_query_data(
...
+        _, var_names, _ = _get_loaded_data(reference_model)
+        var_names = pd.Index(var_names)
+
+        if return_reference_var_names:
+            return var_names
+
+        intersection = adata.var_names.intersection(var_names)
+        inter_len = len(intersection)
+        if inter_len == 0:
+            msg = "No reference var names found in query data. "
```
jjhong922, line 198: just wrap in () to stretch across lines

jjhong922: e.g.
```
raise ValueError("No reference ..."
"etc etc"
)
```

scvi/model/base/_archesmixin.py
```diff
@@ -155,6 +154,78 @@ def load_query_data(
...
+
+        Returns
+        -------
+        Query adata ready to use in `load_query_data` unless `return_reference_var_names`
+        in which case a pd.Index of reference var names is returned.
+        """
+        _, var_names, _ = _get_loaded_data(reference_model)
+        var_names = pd.Index(var_names)
+
+        if return_reference_var_names:
```
jjhong922, line 192: this is kind of strange sub sub functionality of this function. Just split into its own public function instead?

scvi/model/base/_archesmixin.py
```diff
@@ -155,6 +154,78 @@ def load_query_data(
...
+            msg = "No reference var names found in query data. "
+            msg += "Please rerun with return_reference_var_names=True "
+            msg += "to see reference var names."
+            raise ValueError(msg)
+
+        ratio = inter_len / len(var_names)
+        logger.info("Found {}% reference vars in query data.".format(ratio * 100))
+        if ratio < 0.8:
+            warnings.warn(
+                "Query data contains less than 80% of reference var names. "
```
jjhong922, line 207: nit: could make the comment a function of the ratio `f'{ratio_min:.0f}%'`

scvi/model/base/_archesmixin.py
```diff
@@ -155,6 +154,78 @@ def load_query_data(
...
+            [adata, adata_padding],
+            axis=1,
+            join="outer",
+            index_unique=None,
+            merge="unique",
+        )
+        adata_out._inplace_subset_var(var_names)
+
+        if inplace:
+            adata._init_as_actual(adata_out, dtype=adata._X.dtype)
```
jjhong922, line 225: if we start using this a lot, we might want to wrap it in a util since its not guaranteed stable

jjhong922 (approved):
mostly nits

scvi/model/base/_archesmixin.py
adamgayoso: I know... I didn't want to add another method for something so basic...

scvi/model/base/_archesmixin.py
adamgayoso: I'm terrible at long strings...

scvi/model/base/_archesmixin.py
adamgayoso: agree, though now as part of scverse we can do what we want! right?!

scvi/model/base/_archesmixin.py
adamgayoso: though could just be a method called `load_var_names`

scvi/model/base/_archesmixin.py
jjhong922: 🤔

jjhong922 (approved):

PR #1444, opened 2022-03-15
[WIP] MuData support

jjhong922:


scvi/data/fields/_protein.py
```diff
@@ -73,29 +74,30 @@ def _get_batch_mask_protein_data(self, adata: AnnData) -> Optional[dict]:
...
-        if np.sum([~b[1] for b in batch_mask.items()]) > 0:
-            logger.info("Found batches with missing protein expression")
-            return batch_mask
-
+        if self.use_batch_mask:
+            pro_exp = self.get_field_data(adata)
+            pro_exp = (
+                pro_exp.to_numpy() if isinstance(pro_exp, pd.DataFrame) else pro_exp
+            )
+            batches = self.batch_field.get_field_data(adata)
```
adamgayoso, line 82: does this field need to be registered before?

scvi/data/fields/_obs_field.py
```diff
@@ -110,30 +118,34 @@ class CategoricalObsField(BaseObsField):
...
         self._original_attr_key = obs_key or registry_key
-        super().__init__(registry_key, f"_scvi_{registry_key}")
+        super().__init__(registry_key, f"_scvi_{registry_key}", mod_key=mod_key)
 
         self.count_stat_key = f"n_{self.registry_key}"
 
     def _setup_default_attr(self, adata: AnnData) -> None:
         self._original_attr_key = self.attr_key
-        adata.obs[self.attr_key] = np.zeros(adata.shape[0], dtype=np.int64)
+        mod_adata = self._maybe_get_modality(adata)
```
adamgayoso, line 132: should we use bdata globally as the return val of this function?

tests/dataset/utils.py
```diff
@@ -70,3 +71,22 @@ def generic_setup_adata_manager(
...
+def generic_setup_mudata_manager(
+    mdata: MuData,
+    layer: Optional[str] = None,
+    layer_mod: Optional[str] = None,
+) -> AnnDataManager:
+    setup_args = locals()
+    setup_args.pop("mdata")
+    setup_method_args = {_MODEL_NAME_KEY: "TestModel", _SETUP_ARGS_KEY: setup_args}
+
+    anndata_fields = [
```
adamgayoso, line 85: we should add some global obs fields and mod specific obs fields for testing

adamgayoso (changes requested):
Overall pretty good I think.

scvi/data/fields/_obs_field.py
```diff
@@ -110,30 +118,34 @@ class CategoricalObsField(BaseObsField):
...
+        self.set_field_data(adata, np.zeros(mod_adata.n_obs, dtype=np.int64))
```
adamgayoso, line 133: should this be set in the mod_data?

adamgayoso: I see now that it is, maybe ambiguous from the way this is coded....

scvi/data/fields/_protein.py
jjhong922: yes order matters, they get run in the same order they are passed in

scvi/data/fields/_obs_field.py
jjhong922: are you just referring to the naming?

scvi/data/fields/_obs_field.py
jjhong922: yea this is what i dont like about this approach. `adata` can either be mudata or anndata at any point which can be very confusing

scvi/data/fields/_obs_field.py
adamgayoso: yes, trips my mind a bit when this returns the same value as the input var name but sometimes it's not equal.

jjhong922:
Closing in favor of https://github.com/scverse/scvi-tools/pull/1474

PR #1446, opened 2022-03-17
Simplify changing the training plan for pyro

vitkl:
I find myself copying the train function several times to add a different training. Would it make sense to add the training plan as an argument to train method?


adamgayoso:
@jjhong922 what do you think? We can add this in unsupervised training mixin too?

scvi/model/base/_pyromixin.py
```diff
@@ -123,7 +126,7 @@ def train(
                 batch_size=batch_size,
                 use_gpu=use_gpu,
             )
-        training_plan = PyroTrainingPlan(pyro_module=self.module, **plan_kwargs)
+        training_plan = training_plan(pyro_module=self.module, **plan_kwargs)
```
adamgayoso, line 129:
```suggestion
        training_plan = training_plan(self.module, **plan_kwargs)
```

making this general so that it wouldn't raise an error if someone used different kwarg names

adamgayoso (changes requested):
We should also add this to unsupervised train mixin (I can handle that)

vitkl:
> We should also add this to unsupervised train mixin (I can handle that)

Would be great if you can do that

adamgayoso (approved):

PR #1447, opened 2022-03-17
Scale ELBO by a scalar using poutine.scale

vitkl:
@adamgayoso as we discussed.


scvi/train/_trainingplans.py
```diff
@@ -679,23 +683,30 @@ def __init__(
...
+                optim=self.optim,
+                loss=self.loss_fn,
+            )
+        else:
+            self.svi = pyro.infer.SVI(
+                model=self.module.model,
+                guide=self.module.guide,
+                optim=self.optim,
+                loss=self.loss_fn,
             )
```
adamgayoso, line 709:
```suggestion
        def scale(pyro_obj):
            if scale_elbo == 1:
                return pyro_obj
            else:
                return pyro.poutine.scale(pyro_obj, scale_elbo)
                
                
        self.svi = pyro.infer.SVI(
            model=scale(self.module.model),
            guide=scale(self.module.guide),
            optim=self.optim,
            loss=self.loss_fn,
        )
```

scvi/train/_trainingplans.py
```diff
@@ -661,6 +664,7 @@ def __init__(
         optim_kwargs: Optional[dict] = None,
         n_steps_kl_warmup: Union[int, None] = None,
         n_epochs_kl_warmup: Union[int, None] = 400,
+        scale_elbo: Union[float, None] = 1.0,
```
jjhong922, line 667: Is this allowed to be None and if so what is the behavior? The docs make it seem like it requires a positive float.

vitkl: You are right - I copied this from another class where the default behaviour is to scale by the total number of cells.

adamgayoso (approved):

PR #1457, opened 2022-03-22
Updated destVI

cane11:
1. Fixed cell_type amortization (error with neural network structure). Now performs similar to no cell_type amortization. Changed the default to both amortized.
2. Added L1_sparsity similar to final version of manuscript.
3. Added dropout at condSCVI decoder and changed all layers to layer norm to increase stability.
4. Vamp_prior uses over clustering now to find better metapoints. Beforehand the selection of vamp_prior led to different results. Downside: Currently, adds dependency on scanpy and leiden. Do we want this? kmeans clustering was comparable in performance but more unusual choice. We can also remove the cluster generation step from the code and run it outside manually.
5. Changed library size to non-log in CondSCVI
6. Updated hyperparameters to match with reproducibility script from destVI paper.


scvi/model/_condscvi.py
```diff
@@ -3,6 +3,7 @@
 from typing import Optional, Union
 
 import numpy as np
+import scanpy
```
adamgayoso, line 6: we don't depend on scanpy directly, so we will need to do a lazy import of this inside the vamp prior function.

we also need to add leidenalg to our test dependencies.

scvi/model/_condscvi.py
```diff
@@ -128,34 +128,72 @@ def get_vamp_prior(
...
+
+        mean_cat, var_cat = np.array(torch.cat(mean)), np.array(torch.cat(var))
+        adata.obsm["X_CondSCVI"] = mean_cat
+
+        for ct in range(self.summary_stats["n_labels"]):
+            local_indices = np.where(adata.obs[key] == mapping[ct])[0]
+            sub_adata = adata[local_indices, :].copy()
+            if "overclustering_vamp" not in sub_adata.obs.columns:
+                from scanpy.pp import neighbors
+                from scanpy.tl import leiden
```
adamgayoso, line 152: this should look more like this:

https://github.com/scverse/scvi-tools/blob/4c4bd29c279c95fc85ab3ce4603a310debb57be0/scvi/model/utils/_mde.py#L53-L56

adamgayoso:
@cane11 we are losing some test coverage for condscvi -- can you add more tests for these new options?

cane11:
The coverage difference is related to the import error for scanpy (and differences in number of lines). Should I add a test for the ImportError? The new functions seem to be all covered.

romain-lopez:
Thanks, I'll review this ASAP. Ping me if I haven't done it soon enough please :)

adamgayoso:
> The coverage difference is related to the import error for scanpy (and differences in number of lines). Should I add a test for the ImportError? The new functions seem to be all covered.


if that's it then no, no need.

Also as an aside, this version is not backwards compatible with previously saved versions of destVI. I think this is personally fine, but in the future we should put experimental tags on models that exist before a publication is out.

cane11:
It's two bugs (that are not backward compatible), new Vamp_prior is backward compatible using (should we include it in the tutorial):

```
adata.obs['vamp_overclustering'] = list(range(adata.n_obs))
adata.obs['vamp_overclustering'] = adata.obs['vamp_overclustering'].astype('category')
```

I can make the additional parameter updates non hardcoded to increase compatibility. It was hardcoded in the previous version, which is why I kept it  hardcoded (dropout in destVI, layer norm in destVI, weights of bayesian prior on beta and eta). Especially, the weight for beta might be different for unusual data inputs (it regularizes how similar sc and spatial technology are).

scvi/model/_condscvi.py
```diff
@@ -92,9 +92,7 @@ def __init__(
 
     @torch.no_grad()
     def get_vamp_prior(
-        self,
-        adata: Optional[AnnData] = None,
-        p: int = 50,
+        self, adata: Optional[AnnData] = None, p: int = 50, resolution: float = 10.0
```
romain-lopez, line 95: 10 looks high, and it doesn't guarantee a certain number of clusters, or even the same number of clusters per cell type. How do you think about that? 

kmeans definitely seems like an easier alternative (coding-wise as well). Is the leiden choice justified? 

Also, should we just add the leiden package as an option to scvi-tools instead of using scanpy?

scvi/model/_condscvi.py
```diff
@@ -128,34 +128,79 @@ def get_vamp_prior(
...
+
+            var_cluster = np.zeros(
+                [
+                    len(np.unique(sub_adata.obs.overclustering_vamp)),
+                    self.module.n_latent,
+                ]
+            )
+            mean_cluster = np.zeros(
+                [
+                    len(np.unique(sub_adata.obs.overclustering_vamp)),
```
romain-lopez, line 174: len(np.unique(sub_adata.obs.overclustering_vamp)) <- give this a proper name. It's used twice and long

romain-lopez: refactor

scvi/model/_condscvi.py
```diff
@@ -128,34 +128,79 @@ def get_vamp_prior(
...
         key = labels_state_registry.original_key
         mapping = labels_state_registry.categorical_mapping
-        for ct in range(self.summary_stats.n_labels):
-            # pick p cells
-            local_indices = np.random.choice(
-                np.where(adata.obs[key] == mapping[ct])[0], p
+
+        scdl = self._make_data_loader(adata=adata, batch_size=p)
+
+        mean = []
```
romain-lopez, line 134: add a comment line to explain what's happening here (extracting latent, but getting vars in addition to just the mean)

scvi/model/_condscvi.py
```diff
@@ -106,6 +104,8 @@ def get_vamp_prior(
             AnnData object used to initialize the model.
         p
```
romain-lopez, line 105: looks like p has been removed here. Do we want it to be an option then, or do we want just the clustering?

I think clustering should be more robust, so I'd lean towards removing the option p if that makes sense to you. You should have only resolution (or p is you use k-means; but not both)

scvi/model/_destvi.py
```diff
@@ -96,7 +98,9 @@ def from_rna_model(
         cls,
         st_adata: AnnData,
         sc_model: CondSCVI,
-        vamp_prior_p: int = 50,
+        vamp_prior_p: int = 500,
+        vamp_prior_resolution: float = 10.0,
+        l1_sparsity: float = 60.0,
```
romain-lopez, line 103: I feel that this prior default should be zero, I mostly didn't use it in the paper. Did you have some test supporting a diff value?

scvi/model/_destvi.py
```diff
@@ -110,6 +114,10 @@ def from_rna_model(
             trained CondSCVI model
         vamp_prior_p
```
romain-lopez, line 115: same here, either you have vamp_prior_p or vamp_prior_resolution, not both (unless you support for mulitple modes)

scvi/module/_mrdeconv.py
```diff
@@ -155,7 +164,7 @@ def generative(self, x, ind_x):
         m = x.shape[0]
         library = torch.sum(x, dim=1, keepdim=True)
         # setup all non-linearities
-        beta = torch.nn.functional.softplus(self.beta)  # n_genes
+        beta = torch.exp(self.beta)  # n_genes
```
romain-lopez, line 167: softplus was meant for numerical stability. What supports this choice?

scvi/module/_mrdeconv.py
```diff
@@ -121,17 +123,24 @@ def __init__(
...
+            FCLayers(
+                n_in=self.n_genes,
+                n_out=n_hidden,
+                n_layers=2,
+                n_hidden=n_hidden,
+                dropout_rate=0.05,
+                use_layer_norm=True,
+                use_batch_norm=False,
+            ),
+            torch.nn.Linear(n_hidden, n_labels + 1),
```
romain-lopez, line 143: thanks for fixing this :)

scvi/module/_mrdeconv.py
```diff
@@ -217,14 +226,21 @@ def loss(
...
 
         reconst_loss = -NegativeBinomial(px_rate, logits=px_o).log_prob(x).sum(-1)
 
         # eta prior likelihood
         mean = torch.zeros_like(self.eta)
         scale = torch.ones_like(self.eta)
-        glo_neg_log_likelihood_prior = -Normal(mean, scale).log_prob(self.eta).sum()
-        glo_neg_log_likelihood_prior += torch.var(self.beta)
+        glo_neg_log_likelihood_prior = (
+            -1e-10 * Normal(mean, scale).log_prob(self.eta).sum()
```
romain-lopez, line 237: that seems extreme, what's supporting this choice?

scvi/module/_mrdeconv.py
```diff
@@ -217,14 +226,21 @@ def loss(
...
+        )
+        glo_neg_log_likelihood_prior += 5.0 * torch.var(self.beta)
+
+        v_sparsity_loss = (
+            self.l1_sparsity * torch.mean(torch.exp(self.beta)) * torch.abs(v).mean(-1)
```
romain-lopez, line 242: I don't understand what is the motivation behind having mean(exp(beta) here

scvi/module/_mrdeconv.py
```diff
@@ -245,16 +261,20 @@ def loss(
...
                 self.p
             )  # minibatch, n_labels
             neg_log_likelihood_prior = -log_likelihood_prior.sum(1)  # minibatch
             # mean_vprior is of shape n_labels, p, n_latent
 
-        loss = (
-            n_obs * torch.mean(reconst_loss + kl_weight * neg_log_likelihood_prior)
+        loss = n_obs * (
+            torch.mean(
+                reconst_loss + kl_weight * (neg_log_likelihood_prior + v_sparsity_loss)
```
romain-lopez, line 276: sparsity loss shouldn't be affected by kl annealing except if motivated. Also, sometimes it's a local param (if amortized), sometimes it's a global param (if proportions non amortized). Did we make sure it works in both settings?

scvi/module/_mrdeconv.py
```diff
@@ -329,7 +349,7 @@ def get_ct_specific_expression(
             integer for cell types
         """
         # cell-type specific gene expression, shape (minibatch, celltype, gene).
-        beta = torch.nn.functional.softplus(self.beta)  # n_genes
+        beta = torch.exp(self.beta)  # n_genes
```
romain-lopez, line 352: same question

scvi/module/_mrdeconv.py
```diff
@@ -69,14 +70,15 @@ def __init__(
...
+        self.l1_sparsity = l1_sparsity
         # unpack and copy parameters
         self.decoder = FCLayers(
             n_in=n_latent,
             n_out=n_hidden,
             n_cat_list=[n_labels],
             n_layers=n_layers,
             n_hidden=n_hidden,
-            dropout_rate=0,
+            dropout_rate=0.05,
```
romain-lopez, line 81: the zero here was actually super important. If you have layer norm you don't need dropout. And both batchnorm and dropout behaved wierdly for transfer learning so I had deactivated them (on this specific decoder). same for condscvi. Does this work? It's a bit dangerous because it behaves abnormally during .train() compared to .eval(). So I'd prefer to keep this zero unless it's motivated.

scvi/module/_vaec.py
```diff
@@ -84,7 +84,7 @@ def __init__(
             n_cat_list=[n_labels],
             n_layers=n_layers,
             n_hidden=n_hidden,
-            dropout_rate=0,
+            dropout_rate=dropout_rate,
```
romain-lopez, line 87: I'd see this to be zero here, see mrvdecon.py

scvi/model/_destvi.py
cane11: No, I agree with setting it zero. It might be beneficial in edge cases, but also in those cases it makes sense to first run without it. Everything above 120 breaks performance (Tregs are predicted to be not present at all in lymph node).

scvi/module/_mrdeconv.py
cane11: We regularize var(beta) in the loss but not mean(beta). In the generative model, we only use exp(beta)*v. Without also exp(beta) the sparsity loss will lead to bigger beta and thereby smaller v to minimize it. softmax(v) or normalizing v should also work here.

scvi/model/_condscvi.py
adamgayoso: agree kmeans would be ideal dependency wise, for now all this is optional dependency.

scvi/module/_mrdeconv.py
cane11: I tried both non-amortized as well as amortized. The issue is the amortized version: Without annealing the sparsity loss leads to cancelling complete cell types at the beginning of training and the gamma is then not learned correctly, so also at latter time points it's not introducing them. There is no justification for kl_weight and I could also introduce another adaptive weight.

scvi/module/_mrdeconv.py
cane11: Beta is regularized by var and the use of softplus leads to values in the range of 1e-3 (smaller change of beta lead to bigger change in softplus(beta), whereas exp leads to values close to 1. I prefer this but am open to keep it softplus.

scvi/module/_mrdeconv.py
cane11: It's essential to get comparable results with the old code after fixing the library size. It's working with both amortization schemes.

scvi/module/_vaec.py
cane11: Same motivation.

scvi/model/_condscvi.py
cane11: kmeans was my initial choice and I haven't seen performance gain by leiden. Reason to switch were - speed was faster (with 100 initializations for means), selecting n_cluster when there are less cells than clusters is not straightforward, leiden is "default" clustering for single cell.
The resolution is intended to be overclustering (Celltypist uses 5-30 for this purpose https://github.com/Teichlab/celltypist/blob/main/celltypist/classifier.py). I tried multiple values and got closest results to paper with 10 (compared to 5 and 20).
I think new and old code favor cell types with only few cells (sigmas are larger leading to lower loss).

scvi/model/_condscvi.py
cane11: p is still in use, sampling p cells from the constructed clusters based on cluster_probability (it's also beneficial for performance). Leiden tends to find small outlier clusters that are weighted higher when not sampling.

scvi/model/_condscvi.py
adamgayoso: can't you use some kind of other spectral clustering or minibatch kmeans in sklearn?

cane11: It doesn't really depend on the current implementation. I will update it most likely tomorrow (need to run the three datasets again).

scvi/module/_mrdeconv.py
cane11: Oh, we are speaking about different beta. I'm speaking about the output of softplus being 1e-3.As the rate of softplus is not constant, our variance regularization leads to shrinkage of the betas towards -4, which is the region where softplus is an exponential function. It's not an issue performance-wise and I can switch back to softplus. However, I don't really like this property (beta close to 0 is more reasonable to me for interpretation).

scvi/module/_mrdeconv.py
cane11: Works perfectly fine on simulation (there is a gain of 5-10% on gene correlation figure from the paper with the updated code though I wasn't looking what change is responsible for this).

scvi/module/_mrdeconv.py
cane11: Number of cell types yes. However, the sum over v is variable and having smaller sum leads to lower l1 loss, while for the generative model this can be compensated by larger beta values. It introduces a strange dependency.I agree that the current solution is not great. I guess dividing by sum(v) is the best solution.

cane11: By the way, softmax for v instead of softplus is worse in performance, so the model needs the flexibility that sum(v) is not fixed.

scvi/module/_mrdeconv.py
romain-lopez: I don't expect users to look at the betas so unclear if it all matters really. I just wanted to avoid numerical stability. Did you lower the learning rate to 0.003 because otherwise you have numerical instabilities? If yes, I'd lean towards returning to softplus. If you have no numerical instabilities and similar results we can keep exp.

scvi/module/_mrdeconv.py
romain-lopez: I expected that, softmax is not appropriate here. You may want to take the mean over cell types as well (i.e., divide by the sum).

scvi/module/_mrdeconv.py
cane11: No it's not unstable also with lr=0.01. Lower leaning rate led to slightly lower losses. I am pretty convinced the rather large weight on torch.var(beta) ensures there is no instability.

scvi/model/_condscvi.py
cane11: List was slower. However, now it uses a tensor with shape kmeans number of clusters, uses cluster probability as weights and sets values to a weight of 0 if less clusters are given for the corresponding cell type.

scvi/model/_condscvi.py
cane11: Implemented, it's 15 clusters and there is some minor speed gain

scvi/module/_mrdeconv.py
cane11: Even though I don't like the effect of beta, this effect is low when using reasonable values (below 100) and I adjusted the code at this position. Keep in mind that it reduces the estimated counts for each gene by a fixed values when the betas are smaller.

romain-lopez:
Starting to look great!

pyproject.toml
```diff
@@ -65,7 +65,7 @@ python = ">=3.7,<4.0"
 python-igraph = {version = "*", optional = true}
 pytorch-lightning = "~1.5"
 rich = ">=9.1.0"
-scanpy = {version = ">=1.6", optional = true}
+scanpy = {version = ">=1.6", optional = true, extras = ["leiden"]}
```
romain-lopez, line 68: that might not be necessary anymore

scvi/model/_condscvi.py
```diff
@@ -121,41 +119,88 @@ def get_vamp_prior(
...
+                    # Every cell is its own cluster
+                    overclustering_vamp = list(range(len(local_indices)))
+            else:
+                overclustering_vamp = adata[local_indices, :].obs["overclustering_vamp"]
+
+            keys, counts = np.unique(overclustering_vamp, return_counts=True)
+
+            n_labels_overclustering = len(keys)
+            if n_labels_overclustering > p:
+                raise ValueError(
```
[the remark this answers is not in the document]
cane11: It happens with user supplied cell-type subclustering (overclustering_vamp in obs). I think it's not great to change p inplace in these cases and better to have an error. Test for this is included in the test functions.

scvi/model/_destvi.py
```diff
@@ -96,7 +98,8 @@ def from_rna_model(
         cls,
         st_adata: AnnData,
         sc_model: CondSCVI,
-        vamp_prior_p: int = 50,
+        vamp_prior_p: int = 500,
```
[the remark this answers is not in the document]
cane11: My bad. I haven't updated the default here.

scvi/module/_mrdeconv.py
cane11: Up to 1e-4 it works (as it is multiplied by n_obs it's close to the old value).

scvi/model/_condscvi.py
```diff
@@ -121,41 +119,88 @@ def get_vamp_prior(
...
+                    np.where(overclustering_vamp == cluster)[0]
+                ]
+                var_cluster[index, :] = (
+                    np.mean(var_cat[indices_curr], axis=0)
+                    + np.mean(mean_cat[indices_curr] ** 2, axis=0)
+                    - (np.mean(mean_cat[indices_curr], axis=0)) ** 2
+                )
+                mean_cluster[index, :] = np.mean(mean_cat[indices_curr], axis=0)
+
+            mean_vprior[ct, 0:n_labels_overclustering, :] = mean_cluster
```
adamgayoso, line 195: you can use a slice object so this slice code is not repeated in the following three lines

scvi/module/_mrdeconv.py
cane11: 1e-4 also works (everything above leads to wrong cell type proportions). I don't have a strong reason that it helps with the three datasets to have it. 1e-4*n_obs is close to the old term. Beta_weighting is much bigger now and added as an input parameter.

cane11:
All comments should be addressed. I exposed weighting_beta_prior as values between 0.2 and 20 work and I think it is dataset dependent (more regularization possible and beneficial when reference and spatial data are more similar).

scvi/module/_mrdeconv.py
```diff
@@ -43,8 +43,14 @@ class MRDeconv(BaseModuleClass):
         Mean parameter for each component in the empirical prior over the latent space
     var_vprior
         Diagonal variance parameter for each component in the empirical prior over the latent space
+    weight_vprior
+        Weight parameter for each component in the empirical prior over the latent space
     amortization
         which of the latent variables to amortize inference over (gamma, proportions, both or none)
+    l1_sparsity
+        Weighting in loss of L1 regularization of cell type proportions
+    beta_prior_weighting
```
romain-lopez, line 52: polish the docstring (ex.)
scalar parameter indicating the strength of the empirical variance penalty for the multiplicative offset in gene expression values (beta parameter)

Do the same for l1_sparsity

scvi/module/_mrdeconv.py
```diff
@@ -43,8 +43,14 @@ class MRDeconv(BaseModuleClass):
         Mean parameter for each component in the empirical prior over the latent space
     var_vprior
         Diagonal variance parameter for each component in the empirical prior over the latent space
+    weight_vprior
+        Weight parameter for each component in the empirical prior over the latent space
```
romain-lopez, line 47: call this mp_vprior?

mixture proportion of each component in the empirical prior over the latent space

scvi/module/_mrdeconv.py
```diff
@@ -43,8 +43,14 @@ class MRDeconv(BaseModuleClass):
...
     amortization
         which of the latent variables to amortize inference over (gamma, proportions, both or none)
+    l1_sparsity
```
romain-lopez, line 50: call it l1_reg?
scalar parameter indicating the strength of the L1 penalization

scvi/module/_mrdeconv.py
romain-lopez: call it beta_reg?

romain-lopez: also, indicate the values you think are good to try based on your experiments

scvi/module/_mrdeconv.py
```diff
@@ -69,14 +78,16 @@ def __init__(
...
+        self.beta_prior_weighting = beta_prior_weighting
         # unpack and copy parameters
         self.decoder = FCLayers(
             n_in=n_latent,
             n_out=n_hidden,
             n_cat_list=[n_labels],
             n_layers=n_layers,
             n_hidden=n_hidden,
-            dropout_rate=0,
+            dropout_rate=0.05,
```
romain-lopez, line 90: Something else that's kind of funky. Here you use 0.05, but you already changed in DestVI the other dropout to 0.05. Do you want all of them to be destvi.dropout param, or would you like only this one to be stuck to 0.05 when users change destvi.dropout?

If this is the latter option, then please add a comment explaining that the dropout of the decoder should stay small and match the CondSCVI.

romain-lopez: Or just remove the dropout parameters from CondSCVI and DestVI and let them fixed everywhere?

scvi/model/_condscvi.py
```diff
@@ -57,7 +58,7 @@ def __init__(
         n_hidden: int = 128,
         n_latent: int = 5,
         n_layers: int = 2,
-        dropout_rate: float = 0.1,
+        dropout_rate: float = 0.05,
```
romain-lopez, line 61: is that even used anywhere? see my other discussions of this param. Consider either using it or removing it in all the objects.

scvi/module/_mrdeconv.py
```diff
@@ -121,17 +132,24 @@ def __init__(
                 n_cat_list=None,
                 n_layers=2,
                 n_hidden=n_hidden,
-                dropout_rate=0.1,
+                dropout_rate=0.05,
```
romain-lopez, line 135: same discussions about dropout

tests/models/test_models.py
```diff
@@ -1274,18 +1274,26 @@ def test_destvi(save_path):
...
     dataset = synthetic_iid(n_labels=n_labels)
+    dataset.obs["overclustering_vamp"] = list(range(dataset.n_obs))
     CondSCVI.setup_anndata(dataset, labels_key="labels")
     sc_model = CondSCVI(dataset, n_latent=n_latent, n_layers=n_layers)
     sc_model.train(1, train_size=1)
 
-    # step 2 learn destVI with multiple amortization scheme
+    # step 2 Check model setup
+    DestVI.setup_anndata(dataset, layer=None)
+    _ = DestVI.from_rna_model(dataset, sc_model, vamp_prior_p=10000)
```
romain-lopez, line 1284: that seems extreme as a value for the test. Maybe unimportant.

romain-lopez:
Hi Can, thanks for your newer work on this PR. I think this looks almost perfect except for the glitch with the dropout parameter. I might be wrong but it seems that it's not really used anywhere. It should be either user dependent or fixed everywhere or fixed in some places but not some others. I'd lean towards deleting it as an attribute and keeping it hardcoded!

After these minor points, we can proceed to merging this!

tests/models/test_models.py
cane11: I added a comment. This vamp_prior_p is only used to check that preclustering in line 1277 contains less cells than vamp_prior_p (successful clustering outside of DestVI). Next line is then leading to ValueError. And overclustering is deleted afterwards.

cane11: It has to be vamp_prior_p>dataset.n_obs to not have a value error, but I can set it to this value.

scvi/module/_mrdeconv.py
cane11: DestVI.dropout is currently not exposed. So I think they are already fixed in destVI. I also fixed them now for CondSCVI. I have no good intuition whether they should be the same here as in CondSCVI or whether this is by chance here.

cane11: Removed the parameter from both. Here it was working when both are the same for all three datasets. I don't have a good intuition whether it worked because all are small (0.03 was working worse though) or whether they should be the same (I wouldn't understand why they should be the same).

cane11: I kept it in the summary string of CondSCVI for consistence with other models.

tests/models/test_models.py
cane11: Actually it has to be vamp_prior_p>n_largest_celltype. I am not using this more tight value but n_obs here. I don't think this test is critical but I imagine it's a functionality that could go wrong in future releases.

scvi/module/_mrdeconv.py
```diff
@@ -217,14 +235,19 @@ def loss(
...
         mean = torch.zeros_like(self.eta)
         scale = torch.ones_like(self.eta)
-        glo_neg_log_likelihood_prior = -Normal(mean, scale).log_prob(self.eta).sum()
-        glo_neg_log_likelihood_prior += torch.var(self.beta)
+        glo_neg_log_likelihood_prior = (
+            -1e-4 * Normal(mean, scale).log_prob(self.eta).sum()
+        )
+        glo_neg_log_likelihood_prior += self.beta_prior_weighting * torch.var(self.beta)
+
+        v_sparsity_loss = self.l1_sparsity * torch.abs(v).mean(1)
```
adamgayoso: there is a mix of using positive and negative dimension indices throughout. I'd stick to one

adamgayoso (changes requested):

scvi/model/_condscvi.py
```diff
@@ -121,41 +120,86 @@ def get_vamp_prior(
...
 
-        return mean_vprior, var_vprior
+            for index, cluster in enumerate(np.unique(overclustering_vamp)):
+                indices_curr = local_indices[
+                    np.where(overclustering_vamp == cluster)[0]
+                ]
+                var_cluster[index, :] = (
+                    np.mean(var_cat[indices_curr], axis=0)
+                    + np.mean(mean_cat[indices_curr] ** 2, axis=0)
+                    - (np.mean(mean_cat[indices_curr], axis=0)) ** 2
```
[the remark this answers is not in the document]
cane11: Thanks for this suggestions. Better solution.

scvi/model/_condscvi.py
```diff
@@ -121,41 +120,86 @@ def get_vamp_prior(
...
+                    overclustering_vamp = list(range(len(local_indices)))
+            else:
+                overclustering_vamp = adata[local_indices, :].obs["overclustering_vamp"]
+
+            keys, counts = np.unique(overclustering_vamp, return_counts=True)
+
+            n_labels_overclustering = len(keys)
+            if n_labels_overclustering > p:
+                raise ValueError(
+                    "Given overclustering contains more clusters than vamp_prior_p. Increase value."
```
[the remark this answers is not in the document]
cane11: Please check this fix.

scvi/module/_mrdeconv.py
```diff
@@ -217,14 +235,19 @@ def loss(
...
 
         reconst_loss = -NegativeBinomial(px_rate, logits=px_o).log_prob(x).sum(-1)
 
         # eta prior likelihood
         mean = torch.zeros_like(self.eta)
         scale = torch.ones_like(self.eta)
-        glo_neg_log_likelihood_prior = -Normal(mean, scale).log_prob(self.eta).sum()
-        glo_neg_log_likelihood_prior += torch.var(self.beta)
+        glo_neg_log_likelihood_prior = (
+            -1e-4 * Normal(mean, scale).log_prob(self.eta).sum()
```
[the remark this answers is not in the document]
cane11: It doesn't make sense to make it accessible from my side. It's not really important to have it (1e-10 has quite similar performance). However, when you further increase this parameter (bigger than 1e-3) the models starts failing. I would vote against exposing it as I have no suggestion how and when to change it.

scvi/module/_mrdeconv.py
```diff
@@ -244,17 +267,24 @@ def loss(
             var_vprior = torch.transpose(self.var_vprior, 0, 1).unsqueeze(
                 0
             )  # 1, p, n_labels, n_latent
-            pre_lse = (
-                Normal(mean_vprior, torch.sqrt(var_vprior)).log_prob(gamma).sum(-1)
+            weight_vprior = torch.transpose(self.weight_vprior, 0, 1)  # p, n_labels
+            pre_lse = weight_vprior * torch.exp(
+                Normal(mean_vprior, torch.sqrt(var_vprior) + 1e-4)
+                .log_prob(gamma)
+                .sum(-1)
```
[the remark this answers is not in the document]
cane11: @romain-lopez It's LogSumExp in the original code. However, this is my custom function of weighted LogSumExp (not supported by PyTorch). I get the idea of sum instead of multiply. However, it would often lead to adding log(0) in cases where weight_vprior is 0, right? What's your suggestion to put there? Overall, you are right, that the current version is numerically less stable than LogSumExp and I am open to adapt it.

scvi/model/_condscvi.py
```diff
@@ -121,41 +120,86 @@ def get_vamp_prior(
...
+        var = []
+        for tensors in scdl:
+            x = tensors[REGISTRY_KEYS.X_KEY]
+            y = tensors[REGISTRY_KEYS.LABELS_KEY]
+            out = self.module.inference(x, y)
+            mean_, var_ = out["qz_m"], out["qz_v"]
+            mean += [mean_.cpu()]
+            var += [var_.cpu()]
+
+        mean_cat, var_cat = np.array(torch.cat(mean)), np.array(torch.cat(var))
```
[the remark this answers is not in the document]
adamgayoso: this code change is more optimal, uses shared memory, while np array will create new memory

scvi/module/_mrdeconv.py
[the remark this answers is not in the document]
adamgayoso: I would still define this as a variable, perhaps as a kwarg in loss, or a variable at top of file because it makes the code less readable right now

docs/user_guide/models/destvi.md
```diff
@@ -176,8 +182,8 @@ The loss is defined as:
 :nowrap: true
 
 \begin{align}
-     L(l, \alpha, \beta, f^g, \gamma, p, \eta) := &-\log p(X \mid l, \alpha, \beta, f^g, \gamma, p, \eta) - \log p(\eta) \\
-     &+ \mathrm{Var}(\alpha) - \log p(\gamma \mid \mathrm{VampPrior}) \tag{6} \\
+     L(l, \alpha, \beta, f^g, \gamma, p, \eta) := &-\log p(X \mid l, \alpha, \beta, f^g, \gamma, p, \eta) - eta_{reg} \log p(\eta) \\
+     &+ beta_{reg} \mathrm{Var}(\alpha) - \log p(\gamma \mid \mathrm{VampPrior}) + l1_{reg} sum_{c=1}^{C}\beta_{sc} \\
```
adamgayoso, line 186: can yoou use a proper norm notation and add back `\tag{6}`?

docs/user_guide/models/destvi.md
```diff
@@ -106,7 +106,10 @@ $p_g$ is the rate parameter for the negative binomial distribution.
 
 To avoid the latent variable $\gamma_s^c$ from incorporating variation attributed to experimental
 assay differences, we assign an empirical prior informed by the scLVM and a corresponding set of
-cells of the same cell type in the scRNA-seq dataset.
+cells of the same cell type in the scRNA-seq dataset. To compute this function, we subcluster the latent space of the
+scLVM for each cell type to k cell type specific clusters. For each cluster we compute an empirical mean and variance.
+The loss is weighted by the probability of a random cell from this cell type to be in the respective cluster in the
+scRNA-seq dataset (mixture probability, m_{kc}).
```
adamgayoso, line 112:
```suggestion
scRNA-seq dataset (mixture probability, $m_{kc}$).
```

scvi/module/_mrdeconv.py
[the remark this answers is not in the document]
adamgayoso: maybe you can use https://pytorch.org/docs/stable/generated/torch.Tensor.masked_scatter_.html#torch.Tensor.masked_scatter_

?

scvi/module/_vaec.py
```diff
@@ -71,7 +70,7 @@ def __init__(
             n_cat_list=[n_labels],
             n_layers=n_layers,
             n_hidden=n_hidden,
-            dropout_rate=dropout_rate,
+            dropout_rate=0.05,
```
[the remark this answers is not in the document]
cane11: Then I expose it again and expose. Dropout_rate for dec and amortization in mrdeconv are different values.

scvi/model/_condscvi.py
```diff
@@ -121,41 +120,86 @@ def get_vamp_prior(
...
+            if "overclustering_vamp" not in adata.obs.columns:
+                if p < len(local_indices) and p > 0:
+                    overclustering_vamp = KMeans(n_clusters=p, n_init=30).fit_predict(
+                        mean_cat[local_indices]
+                    )
+                else:
+                    # Every cell is its own cluster
+                    overclustering_vamp = list(range(len(local_indices)))
+            else:
+                overclustering_vamp = adata[local_indices, :].obs["overclustering_vamp"]
```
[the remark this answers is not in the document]
cane11: I put a paragraph about it in the tutorial.

docs/user_guide/models/destvi.md
```diff
@@ -230,8 +237,13 @@ impute the spatial pattern of the cell-type-specific gene expression with:
...
-can be found in the DestVI paper.
+from the parameters of the generative distribution predicted for each spot in question. 
+### Utilities function
+
+To explore the results of the output of the stLVM, we published a utilities function covering functions
+for automatic thresholding of cell type proportions, a spatial PCA analysis to find main axis of variation
+in spatial gene expression and the described frequentist test for differential expression. Further information
+can be found on [destvi_utils](https://destvi-utils.readthedocs.io/en/latest/installation.html)
 
 [^ref1]: Romain Lopez, Baoguo Li, Hadas Keren-Shaul, Pierre Boyeau, Merav Kedmi, David Pilzer, Adam Jelinski, Eyal David, Allon Wagner, Yoseph Addad, Michael I. Jordan, Ido Amit, Nir Yosef (2021),
```
romain-lopez, line 248: ah, we can change the title here

scvi/model/_destvi.py
```diff
@@ -110,6 +113,9 @@ def from_rna_model(
             trained CondSCVI model
```
romain-lopez, line 113: here we should make sure we apply to DestVI the dropout parameters from CondSCVI (for sure the dropout_decoder! they have to be the same!)

romain-lopez (approved):
two remaining minor things

docs/user_guide/models/destvi.md
romain-lopez: Romain Lopez, Baoguo Li, Hadas Keren-Shaul, Pierre Boyeau, Merav Kedmi, David Pilzer, Adam Jelinski, Ido Yofe, Eyal David, Allon Wagner, Can Ergen, Yoseph Addadi, Ofra Golani, Franca Ronchese, Michael I Jordan, Ido Amit, Nir Yosef. DestVI identifies continuums of cell types in spatial transcriptomics data. Nature Biotechnology (in press), 2022.

cane11: I made it consistent with other packages and also added link to VAMP prior paper

romain-lopez:
@adamgayoso ready to merge!

adamgayoso (approved):

adamgayoso:

adamgayoso:

PR #1473, opened 2022-03-29
pin memory False!

adamgayoso:
Turns out this doesn't really make things faster, and really messes things up in shared environments.


jjhong922 (approved):

adamgayoso:
@meeseeksdev backport to 0.16.x

PR #1475, opened 2022-03-29
Fix unnecessary warnings, logger propagate

adamgayoso:
This:

- Fixes a bug where totalVI was resetting the warnings filters accidentally, now it's all inside a context.
- Prevents logger for percolating upward to root logger (stops double logging due to jax)
- Fixes seurat v4 pbmc dataset creation fragmentation warning from pandas


jjhong922 (approved):
release notes then good to go

PR #1477, opened 2022-03-30
Lightning 1.6 official

adamgayoso:


adamgayoso:
@jjhong922 I think we should get this in staged for 0.16.0. There will be more weird messages from this lightning release we will have to handle as we work on other 0.16.0 changes

jjhong922 (approved):

PR #1496, opened 2022-04-07
Add batch embedding

wukathy:
Add option for 'use_batch_embedding' in VAE class that encodes batch covariates using nn.Embedding (defaults to an embedding_dim of 5)


scvi/model/base/_vaemixin.py
```diff
@@ -177,3 +177,7 @@ def get_latent_representation(
 
             latent += [z.cpu()]
         return torch.cat(latent).numpy()
+
+    @torch.no_grad()
+    def get_batch_embedding(self):
```
jjhong922, line 182: Need typing on the return. It seems like it would be a torch object which is not preferable at the model abstraction level. Convert to a numpy array before returning.

scvi/nn/_base_components.py
```diff
@@ -67,18 +69,23 @@ def __init__(
...
+
         if n_cat_list is not None:
             # n_cat = 1 will be ignored
             self.n_cat_list = [n_cat if n_cat > 1 else 0 for n_cat in n_cat_list]
         else:
             self.n_cat_list = []
 
         cat_dim = sum(self.n_cat_list)
+        if self.batch_embedding:
+            cat_dim = 5
```
jjhong922, line 88: this should be parametrizable right?

jjhong922: Also does this work with categorical covariates?

scvi/model/base/_vaemixin.py
jjhong922: Oh now i see whats going on with this torch.nn.Embedding typing. I think it would be more natural to take batch indices as a parameter and return the relevant embeddings to the user. If the indices=None then you just return the embedding for each of the batches used to train the model. Generally better to avoid returning torch level objects at the model abstraction level (allows for different backends, keeps user level backend agnostic, avoids issues with memory device management)

scvi/nn/_base_components.py
```diff
@@ -340,6 +351,7 @@ def __init__(
         inject_covariates: bool = True,
         use_batch_norm: bool = False,
         use_layer_norm: bool = False,
+        batch_embedding: nn.Embedding = None,
```
jjhong922, line 354: Little odd to me to take an Embedding object here.  Injecting external state into the constructor is inconsistent with the rest of the parametrization. Would prefer you convert to embedding in the `generative` function and then pass it in as a vector to the decoder. Then just make sure the decoder can accept this additional input by replacing this parameter with a int (similar to what `n_cat_list` does).

jjhong922 (commented):
Good initial PR on batch embedding! Just left a few comments on improving code style + abstraction. Adam might have some higher level comments on what to do with the code, but hopefully this is helpful regardless.

wukathy:
will use scvi-tools-skeleton instead

PR #1502, opened 2022-04-08
Fix backwards compatibility for TOTALVI

jjhong922:
Since TOTALVI.setup_anndata was modified to remove the unnecessary labels key, issues arose when a legacy totalvi model was loaded then saved, then loaded again.


adamgayoso (changes requested):
can you add this to the 0.16 release note?

We should talk about removing legacy save load -- I don't think we should get rid of it yet

adamgayoso (approved):

PR #1505, opened 2022-04-08
Add backup_url to load and move legacy loading functions out of load

jjhong922:
Fixes https://github.com/scverse/scvi-tools/issues/1484

### DIFF
--- a/scvi/model/_utils.py
+++ b/scvi/model/_utils.py
@@ -1,5 +1,4 @@
 import logging
-import os
 import warnings
 from collections.abc import Iterable as IterableClass
 from typing import Dict, List, Optional, Sequence, Tuple, Union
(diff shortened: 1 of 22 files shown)


scvi/model/_utils.py
```diff
@@ -10,10 +11,18 @@
 from scvi import REGISTRY_KEYS
 from scvi._types import Number
 from scvi.data import AnnDataManager
+from scvi.data._built_in_data._download import _download
 
 logger = logging.getLogger(__name__)
 
 
+def _download_if_missing(filepath: str, backup_url: Optional[str]):
```
adamgayoso, line 19: this function might not work in some cases where you can't simply add the filename to the url. I think we should remove backup url from the new legacy function and in the new load function, backup url should download the file with the user provided url and put the file inside the load path specified by the user (just like how backup url in scanpy read works)

adamgayoso: I think this situation arises in zenodo for example

adamgayoso (approved):

PR #1506, opened 2022-04-10
`JaxTrainingPlan` via PyTorch Lightning and `JaxModuleWrapper` abstraction

adamgayoso:
Adam: Pytorch lightning training plan sounds crazy, but seems this will be a moderately safe solution with minimal overhead for us.

Justin: This PR additionally includes the changes of https://github.com/scverse/scvi-tools/pull/1552. It implements the JaxModuleWrapper which interacts directly with top level scvi-tools model classes. It duck types methods required to save/load to avoid jax specific reimplementation of save/load. It also removes much of the jax specific complexity that was in JaxSCVI as well as in the training plan.

Additionally option to explore is https://github.com/deepmind/jaxline.

Note: with these changes, the `JaxSCVI` implementation will remain optional until we gain confidence in the `JaxModuleWrapper` abstraction as well as the use of pytorch lightning for training.


scvi/train/_trainingplans.py
```diff
@@ -853,3 +865,215 @@ def configure_optimizers(self):
...
+
+    @property
+    def rngs(self) -> Dict[str, jnp.ndarray]:
+        return self._rngs
+
+    @rngs.setter
+    def rngs(self, rngs: Dict[str, jnp.ndarray]):
+        self._rngs = rngs
+
+    def set_rngs(self, keys: List[str]):
```
jjhong922, line 961: I think we can operate with PRNGKeyArrays here instead of list of ndarrays https://github.com/google/jax/blob/b051be4a7a163502ab871cf16e5f749d49622042/jax/_src/prng.py#L99.

Seems pretty feature rich and the default typing for most functions. For example random.split takes an array and returns an array

jjhong922: This hides having explicit PRNGKeys floating around in a dict.

scvi/train/_trainingplans.py
```diff
@@ -853,3 +865,215 @@ def configure_optimizers(self):
...
+            loss_kwargs=self.loss_kwargs,
+        )
+        loss = torch.tensor(jax.device_get(loss))
+        elbo = torch.tensor(jax.device_get(elbo))
+        # TODO: Better way to get batch size
+        self.log(
+            "train_loss",
+            loss,
+            on_epoch=True,
+            batch_size=batch[REGISTRY_KEYS.X_KEY].shape[0],
```
jjhong922, line 1009: don't see a way around this one

scvi/train/_trainingplans.py
```diff
@@ -853,3 +865,215 @@ def configure_optimizers(self):
...
+        """Scaling factor on KL divergence during training."""
+        return _compute_kl_weight(
+            self.current_epoch,
+            self.global_step,
+            self.n_epochs_kl_warmup,
+            self.n_steps_kl_warmup,
+        )
+
+    def transfer_batch_to_device(self, batch, device, dataloader_idx):
+        """Bypass Pytorch Lightning device management."""
```
jjhong922, line 1066: where is the jax device management? Is it determined by the device of the PRNG key?

jjhong922 (changes requested):

scvi/train/_trainingplans.py
jjhong922: ignore comment. Moved in https://github.com/scverse/scvi-tools/pull/1552

scvi/train/_trainingplans.py
jjhong922: issue created for this https://github.com/scverse/scvi-tools/issues/1553

scvi/model/_jaxscvi.py
```diff
@@ -130,6 +131,67 @@ def module(self):
...
+        self,
+        max_epochs: Optional[int] = None,
+        use_gpu: Optional[Union[str, int, bool]] = None,
+        train_size: float = 0.9,
+        validation_size: Optional[float] = None,
+        batch_size: int = 128,
+        lr: float = 1e-3,
+        **trainer_kwargs,
+    ):
+        if max_epochs is None:
```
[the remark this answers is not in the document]
jjhong922: this is not a big deal after https://github.com/scverse/scvi-tools/pull/1552

jjhong922 (approved):
unblocking

PR #1511, opened 2022-04-19
Allow additional arguments for SCVI encoder and decoder

Koncopd:
related to https://github.com/scverse/scvi-tools/issues/1510

Allows to do this
`set_activation = dict(activation_fn=torch.nn.Softplus)`
`vae = scvi.model.SCVI(adata, encoder_kwargs=set_activation, decoder_kwargs=set_activation)`

### DIFF
--- a/scvi/module/_vae.py
+++ b/scvi/module/_vae.py
@@ -112,6 +112,8 @@ class VAE(BaseModuleClass):
         library_log_means: Optional[np.ndarray] = None,
         library_log_vars: Optional[np.ndarray] = None,
         var_activation: Optional[Callable] = None,
+        encoder_kwargs: dict = {},
+        decoder_kwargs: dict = {},
     ):
         super().__init__()
         self.dispersion = dispersion
@@ -177,6 +179,7 @@ class VAE(BaseModuleClass):
             use_batch_norm=use_batch_norm_encoder,
             use_layer_norm=use_layer_norm_encoder,
             var_activation=var_activation,
+            **encoder_kwargs
         )
         # l encoder goes from n_input-dimensional data to 1-d library size
         self.l_encoder = Encoder(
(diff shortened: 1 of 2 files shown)


scvi/module/_vae.py
```diff
@@ -112,6 +112,8 @@ def __init__(
         library_log_means: Optional[np.ndarray] = None,
         library_log_vars: Optional[np.ndarray] = None,
         var_activation: Optional[Callable] = None,
+        encoder_kwargs: dict = {},
+        decoder_kwargs: dict = {},
```
adamgayoso, line 116: not ideal to use mutable defaults, this should be `Optional[dict] = None`, also should be documented.

I would also considering calling these `extra_encoder_kwargs` because some init kwargs are passed to init of the encoder.

scvi/nn/_base_components.py
```diff
@@ -341,6 +341,7 @@ def __init__(
         use_batch_norm: bool = False,
         use_layer_norm: bool = False,
         scale_activation: Literal["softmax", "softplus"] = "softmax",
+        **kwargs,
```
adamgayoso, line 344: needs documentation

scvi/module/_vae.py
adamgayoso: @jjhong922 any thoughts on that?

adamgayoso (changes requested):
Thanks for this PR!

scvi/module/_vae.py
jjhong922: Yes agree with both your comments. I'm actually surprised codacy didnt complain about a mutable default!

Koncopd:
thx, i will make the changes.

wmacnair:
Hey, I would also be interested in trying out different activation functions. I guess this PR still isn't in the code base, is that right? Any chance it could be...? :D

Thanks!

adamgayoso:
I'm happy to get something here in. Let me discuss with @martinkim0. Ideally we would have configs, but this might not be doable.

PR #1519, opened 2022-04-22
scanvi scarches fixes

adamgayoso:
1. `allow_missing_labels` wasn't working properly if `_scvi_labels` was in obs because it should be looking for the original attr key. This can occur if the query data was already used for `SCVI` but without a labels key in setup anndata
2. We should not allow extending the categories of labels (seeing labels in query that weren't in reference)

### DIFF
--- a/scvi/data/fields/_scanvi.py
+++ b/scvi/data/fields/_scanvi.py
@@ -97,7 +97,7 @@ class LabelsWithUnlabeledObsField(CategoricalObsField):
             )
 
         transfer_state_registry = super().transfer_field(
-            state_registry, adata_target, **kwargs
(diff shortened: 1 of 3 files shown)


scvi/data/fields/_scanvi.py
```diff
@@ -98,7 +97,7 @@ def transfer_field(
             )
 
         transfer_state_registry = super().transfer_field(
-            state_registry, adata_target, extend_categories=extend_categories, **kwargs
+            state_registry, adata_target, **kwargs
```
jjhong922, line 100: Extend categories = False here

jjhong922 (approved):

adamgayoso:

PR #1520, opened 2022-04-22
0.16.1 release staging

adamgayoso:
also fixes to `prepare_query_data`

### DIFF
--- a/scvi/model/base/_archesmixin.py
+++ b/scvi/model/base/_archesmixin.py
@@ -206,21 +206,28 @@ class ArchesMixin:
                 "This may result in poor performance."
             )
         genes_to_add = var_names.difference(adata.var_names)
-        adata_padding = AnnData(csr_matrix(np.zeros((adata.n_obs, len(genes_to_add)))))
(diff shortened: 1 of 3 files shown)


adamgayoso:
@jjhong922 this is good to go, just going to update the scarches tutorial one more time

scvi/model/base/_archesmixin.py
```diff
@@ -206,21 +206,31 @@ def prepare_query_anndata(
...
+                axis=1,
+                join="outer",
+                index_unique=None,
+                merge="unique",
+            )
+        else:
+            adata_out = adata
+
+        # also covers the case when new adata has more var names than old
+        if not var_names.equals(adata_out.var_names):
```
jjhong922, line 228: think you can move both of these cases into the if case above

jjhong922: not super necessary though

jjhong922 (approved):
lgtm. Might want to add tests that fail on the old arches code though

adamgayoso:

PR #1529, opened 2022-05-03
clean up metrics code, use MetricCollection

adamgayoso:
By calling .update() explicitly, it should speed things up according to their documentation as we don't need the per step values.

This also fixes an issue where `n_obs_training` might not reflect the number of cells in an epoch, as in the large scale training params where we restrict an epoch to 20 minibatches

### DIFF
--- a/scvi/train/_metrics.py
+++ b/scvi/train/_metrics.py
@@ -4,14 +4,12 @@ from torchmetrics import Metric
 from scvi._compat import Literal
 
 
-class ElboMetric(Metric):
+class BaseElboMetric(Metric):
     """
     Elbo metric aggregator for scvi-tools experiments.
 
     Parameters
     ----------
-    n_obs_total
-        Number of total observations, for rescaling the ELBO
     mode
         Train or validation, used for logging names
     dist_sync_on_step
@@ -22,20 +20,16 @@ class ElboMetric(Metric):
 
     def __init__(
         self,
-        n_obs_total: int,
         mode: Literal["train", "validation"],
         dist_sync_on_step: bool = False,
         **kwargs
     ):
         super().__init__(dist_sync_on_step=dist_sync_on_step, **kwargs)
 
-        self.n_obs_total = 1 if n_obs_total is None else n_obs_total
         self._mode = mode
 
         default_val = torch.tensor(0.0)
-        self.add_state("reconstruction_loss", default=default_val)
-        self.add_state("kl_local", default=default_val)
-        self.add_state("kl_global", default=default_val)
+        self.add_state("elbo_component", default=default_val)
         self.add_state("n_obs", default=default_val)
         self.add_state("n_batches", default=default_val)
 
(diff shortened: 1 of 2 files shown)


scvi/train/_metrics.py
```diff
@@ -45,29 +39,115 @@ def mode(self):
...
+
+    def update(
+        self,
+        kl_global: torch.Tensor,
+        n_obs_minibatch: int,
+    ):
+        """Updates all metrics."""
+        super().update(kl_global, n_obs_minibatch)
+
+    def compute(self):
```
jjhong922, line 150: this looks like the only thing that sets apart this class from the rest. Otherwise a lot of copy paste. To avoid over generalizing, maybe you could just stick to having the base class, except it takes a metric name, and you add the kwarg for what to normalize by?

scvi/train/_trainingplans.py
```diff
@@ -136,14 +137,49 @@ def __init__(
...
+            self.rec_loss_val,
+            self.kl_local_val,
+            self.kl_global_val,
+            self.val_metrics,
+        ) = self._create_elbo_metric_components(
+            mode="validation", n_total=self.n_obs_validation
+        )
         self.elbo_val.reset()
 
+    def _create_elbo_metric_components(self, mode: str, n_total: Optional[int] = None):
```
jjhong922, line 164: nit: move to top of class (before usage)

scvi/train/_trainingplans.py
```diff
@@ -136,14 +137,49 @@ def __init__(
...
+        """Initialize ELBO metric and the metric collection."""
+        rec_loss = ReconstructionLossMetric(mode)
```
jjhong922, line 166: based on my first comment this could be replaced by `ElboMetric("reconstruction_loss", mode, "n_obs")` etc.

scvi/train/_trainingplans.py
```diff
@@ -136,14 +137,49 @@ def __init__(
...
+        kl_local = KLLocalMetric(mode)
+        kl_global = KLGlobalMetric(mode)
+        # n_total can be 0 if there is no validation set, this won't ever be used
```
jjhong922, line 169: can add an assertion statement here for this? `raise AssertionError()` to satisfy codacy

jjhong922:
@PierreBoyeau  @adamgayoso  updated and addressed my own comments. Condensed code by relying on this part of the metricscollection code https://github.com/PyTorchLightning/metrics/blob/11641b8558e713649743709d78d3ab363bcfc9aa/torchmetrics/metric.py#L681-L682

jjhong922:
Also noticed GIMVI doesnt use this and seems hard to adapt to it. Does it matter?

scvi/train/_trainingplans.py
```diff
@@ -209,44 +245,15 @@ def compute_and_log_metrics(
...
-            reduce_fx=torch.sum,
-            on_step=False,
-            on_epoch=True,
-            batch_size=n_obs_minibatch,
-        )
-        # default aggregation is mean
-        self.log(
-            f"kl_global_{mode}",
-            kl_global,
+        self.log_dict(
```
PierreBoyeau, line 255: One naive question as this is not fresh in my mind anymore. Where are the metrics explicitly reset to 0?

jjhong922: They get set to 0 on __init__ https://github.com/scverse/scvi-tools/pull/1529/files/2eb5cb855bfaa3fc2a7a45f057928fe1742dabc6#diff-8fc39d9964444c66caf3545a773bef7402a7a73386aa2761f06930d7759353daR135 as well as when `n_obs_training` is set, which happens in https://github.com/scverse/scvi-tools/blob/master/scvi/train/_trainrunner.py#L70.

scvi/train/_metrics.py
```diff
@@ -10,64 +10,81 @@ class ElboMetric(Metric):
...
+
     def update(
         self,
-        reconstruction_loss_sum: torch.Tensor,
-        kl_local_sum: torch.Tensor,
-        kl_global: torch.Tensor,
-        n_obs_minibatch: int,
+        **kwargs,
     ):
         """Updates all metrics."""
```
adamgayoso, line 74: might be good to explain somehwere that in a metric collection that arguments get passed to each individual metric

PR #1535, opened 2022-05-10
Create release.yml github action

adamgayoso:
this came from https://github.com/python-poetry/poetry/blob/master/.github/workflows/release.yml

My token has been added to the secrets in the repo.

Since this action will initiate whenever a tag with name meeting `"*.*.*"` is met, I'm inclined to force everyone to use forks for all PRs. Only admins would have write permissions.


adamgayoso:

Issue #1483, opened 2022-04-01
Model load does not pass in train parameters for pyro module


chaichontat:
Hi, I was loading a previously trained model with a small batch size. In the `load` function, attempts to train the model by one step. This fails as there is not enough memory.

https://github.com/scverse/scvi-tools/blob/6e734bc828521d58d92265548c0fdc338ecb22cf/scvi/model/base/_base_model.py#L632-L636

The `train` function that I'm using accepts other keyword arguments including batch size and passing that in resolves this issue. Would it be possible to add keyword arguments to the `load` function that would be forwarded into the `train` function? I could make a PR if that is ok.

adamgayoso:
thanks for the issue, what kwargs exactly would you want to pass? If it's just batch size we can probably do that internally. Also it's important to consider not all pyro models will train with minibatches. @vitkl any ideas?

chaichontat:
Like you said, since this is an abstract class, I'm thinking in terms of 
```python
  @classmethod
  def load(
      cls,
      dir_path: str,
      adata: Optional[AnnData] = None,
      use_gpu: Optional[Union[str, int, bool]] = None,
      prefix: Optional[str] = None,
      train_kwargs: Optional[dict[str, Any]] = None,
  ):
```
where the user is free to pass their own `dicts`.

adamgayoso:
I'm more inclined to have a solution that looks more like a callback. Suppose we had `model.on_load_module_start` @jjhong922 any thoughts?

jjhong922:
Given that this train step only happens in the Pyro model for a very specific reason, I would avoid putting it in the load signature of the BaseModelClass. Agree with @adamgayoso about having a callback solution or just fixing it for this case internally. Alternatively, you can also make this Module specific and put an `on_load()` callback on the Module level, then have that called by `model.load()`. This way you can just put it on the `PyroBaseModuleClass` and fix it across any of the relevant modules and, in turn, models.

adamgayoso:
@jjhong922 we can also write a simple callback class specific to loading, and this class can be associated with the model class as like a property or something. This way in this case, devs can overwrite the one method they want to in the callback and tell the model to use that callback class

vitkl:
I agree that would be good to pass train parameters to the load function. The current solution is that all models are loaded with minibatch training and default minibatch side (based on `batch_size = None`). Would be good to have an option to do otherwise. 

[...]

PR #1542, opened 2022-05-16
Add `on_load()` callback to PyroBaseModuleClass

jjhong922:
Addresses #1483 

Note this PR also includes a bunch of small fixes to the backticks in this file that are unrelated to the issue.


adamgayoso (commented):
what's the tradeoff of this being a model/module method?

PR #1552, opened 2022-06-02
Implement `JaxModuleWrapper`

### LINKED ISSUE #1388: Implement `JaxModuleWrapper`
Refer to #1366 for overview. This should allow us to remove all the notimplementederrors from `JaxSCVI`

jjhong922:
Fixes #1388 

This PR implements the `JaxModuleWrapper` which interacts directly with top level scvi-tools model classes. It duck types methods required to save/load to avoid jax specific reimplementation of save/load. It also removes much of the jax specific complexity that was in `JaxSCVI` as well as in the training plan.

Beyond this PR, there still needs to be better device management (see https://github.com/scverse/scvi-tools/issues/1553).


scvi/train/_trainingplans.py
```diff
@@ -1018,24 +984,26 @@ def training_step(self, batch, batch_idx):
...
-        outputs = module.apply(vars_in, batch, rngs=rngs, **kwargs)
+        outputs = self.module.apply(vars_in, batch, rngs=rngs, **kwargs)
         loss_recorder = outputs[2]
         loss = loss_recorder.loss
         elbo = jnp.mean(loss_recorder.reconstruction_loss + loss_recorder.kl_local)
 
         return loss, elbo
 
     def validation_step(self, batch, batch_idx):
+        self.module.eval()
```
adamgayoso, line 1001: we could probably call this once in the init right? I know it's not intensive, but probably is worth it.

Same for the train version

scvi/module/base/_jax_module_wrapper.py
```diff
@@ -0,0 +1,125 @@
...
+from flax.core import FrozenDict
+from flax.training import train_state
+from jax import random
+
+from scvi.utils._jax import device_selecting_PRNGKey
+
+from ._base_module import JaxBaseModuleClass
+
+
+class BatchTrainState(train_state.TrainState):
```
adamgayoso, line 14: batchtrainstate sounds like minibatching, how about `TrainStateWithBatchNorm`?

scvi/model/base/_base_model.py
```diff
@@ -637,6 +637,12 @@ def load(
                 model.history_ = old_history
                 pyro.clear_param_store()
                 model.module.load_state_dict(model_state_dict)
+            elif isinstance(model.module, JaxModuleWrapper):
```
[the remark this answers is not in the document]
adamgayoso: also, why is this necessary?

scvi/model/_jaxscvi.py
```diff
@@ -420,10 +143,12 @@ def get_latent_representation(
             adata=adata, indices=indices, batch_size=batch_size, iter_ndarray=True
         )
 
+        bound_module = self.module.bound_module
+
         @jax.jit
         def _get_val(array_dict):
```
adamgayoso, line 149: We should add this to the wrapper so that devs don't need to interact with the bound module in most cases. Also, Flax does not recommend using these bound modules extensively for some reason.

scvi/model/base/_base_model.py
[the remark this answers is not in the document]
adamgayoso: In my mind, since we load the train state, you should be able to run apply right away?

scvi/train/_trainingplans.py
adamgayoso: ah I see it's cached behind the scenes, that's fine then.

adamgayoso (commented):
Looks pretty good overall, added some comments

scvi/model/base/_base_model.py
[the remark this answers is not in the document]
jjhong922: The load function needs a template object to load the state dict into. From what I can tell there is no better way to get an empty train state object so went with this way.

And yes, will rebase and use the on_load instead.

scvi/model/_jaxscvi.py
jjhong922: sounds good

scvi/model/base/__init__.py
```diff
@@ -16,4 +17,5 @@
     "PyroSampleMixin",
     "PyroJitGuideWarmup",
     "DifferentialComputation",
+    "JaxTrainingMixin",
```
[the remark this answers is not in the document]
jjhong922: all of the other training mixins are in this package though

scvi/module/_jaxvae.py
```diff
@@ -76,15 +76,15 @@ def setup(self):
...
-            momentum=0.99, epsilon=0.001, use_running_average=training
+            momentum=0.99, epsilon=0.001, use_running_average=is_eval
         )
         self.batchnorm2 = nn.BatchNorm(
-            momentum=0.99, epsilon=0.001, use_running_average=training
+            momentum=0.99, epsilon=0.001, use_running_average=is_eval
         )
-        self.dropout1 = nn.Dropout(self.dropout_rate, deterministic=training)
-        self.dropout2 = nn.Dropout(self.dropout_rate, deterministic=training)
+        self.dropout1 = nn.Dropout(self.dropout_rate, deterministic=is_eval)
```
PierreBoyeau, line 86: Do we need dropout for the decoder if we never use it?

jjhong922: Since its just a generalized decoder component I think we can leave it. We just dont happen to use it in JaxVAE specifically

PR #1565, opened 2022-06-10
Ray.tune for parameter optimization. Included more functionality

nrclaudio:
Hi @adamgayoso and @PierreBoyeau. As we discussed recently via email, in our lab we were able to use the logic in [michael/autotune](https://github.com/scverse/scvi-tools/tree/michael/autotune) to do hyperparameter optimization.  

I've implemented the changes I applied to the original `Autotune` module. I also included the option to test for different number of HVGs, use of covariates in the model HVG filters. In our implementation we also included a descriptive plot for the `tune `experiment. I didn't include this in the module but we can discuss if it's something worth including. 

The environment needs to have **ray tune installed** with `pip install -U "ray[tune]"`  but I wasn't sure how to add dependencies to the PR. 

*y ax: silhouette score
*x ax: epoch

I'm not sure if I should've PRd to [master](https://github.com/scverse/scvi-tools) instead, let me know if that's the case and I can open a new one there.


jjhong922:
Hi @nrclaudio, thanks for providing this PR. If you could PR to master instead that would be great. Right now the changes include everything on master since the original autotune branch which makes it hard to parse out the contributed code.

nrclaudio:
Hi @jjhong922, I think it should be up to date now

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,346 @@
...
+        self,
+        adata: anndata.AnnData,
+        model: BaseModelClass,
+        num_epochs: int = 2,
+        training_metrics: Optional[list[str]] = None,
+        metric_functions: Optional[dict] = None,
+        model_hyperparams: Optional[dict] = None,
+        trainer_hyperparams: Optional[dict] = None,
+        plan_hyperparams: Optional[dict] = None,
+        continious_covariates: Optional[list[str]] = None,
```
jjhong922, line 71: These covariates are not present in every model's `setup_anndata`, so we would need to restrict these values to be non-None only in the case that they are present in the signature of the `setup_anndata` method. Additionally, I'm wondering if for kwargs like these covariates or hvg selection below, we should raise warnings if users attempt to autotune over a metric like `elbo_validation`. Since the model architectures would be entirely different, it would not be a fair comparison. However, I can see using the silhouette for batch being a reasonable one.


also there is a typo in: continuous*

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,346 @@
...
+        categorical_covariates: Optional[list[str]] = None,
+        test_effect_covariates: bool = False,
+        batch_key: Optional[str] = None,
+        test_effect_hvg: bool = False,
+        top_hvg: Optional[list[int]] = None,
+    ):
+        if not training_metrics:
```
jjhong922, line 78: nit: these can be abbreviated like so: `training_metrics = training_metrics or []`.

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,346 @@
...
+        if "subset_adata" in hvg_config is True:
+            sc.pp.highly_variable_genes(
+                self.adata,
+                flavor="seurat_v3",
+                n_top_genes=hvg_config["top_n"],
+                batch_key=self.batch_key,
+                subset=True,
+            )
+
+        self.model.setup_anndata(
```
jjhong922, line 325: to generalize to other models, we'll have to allow for passing in additional setup_args. They don't have to be tunable parameters, but maybe we could just add a `setup_args: Optional[Dict] = None` kwarg to the init fn

scvi/autotune/_callbacks.py
```diff
@@ -0,0 +1,51 @@
...
+    def __init__(self, model):
+        super()
+        self.model = model
+
+    def on_validation_epoch_end(self, trainer, pl_module, outputs=None):
+        if trainer.sanity_checking:
+            return
+        step = f"epoch={trainer.current_epoch}-step={trainer.global_step}"
+        with tune.checkpoint_dir(step=step) as checkpoint_dir:
+            self.model.save(dir_path=checkpoint_dir + "/checkpoint")
```
jjhong922, line 16: use os.path.join here to be safe

scvi/autotune/_callbacks.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class _TuneReportMetricFunctionsCallback(TuneCallback):
+    def __init__(
+        self,
+        metrics=None,
```
jjhong922, line 22: nit: add typings

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,346 @@
...
+            elif key in self.categorical_covariates:
+                categorical_covariate_config[key] = config[key]
+            elif (
+                (self.test_effect_hvg and self.top_hvg) is not None
+                and key in self.test_effect_hvg
+                or self.top_hvg
+            ):
+                hvg_config[key] = config[key]
+
+        if "subset_adata" in hvg_config is True:
```
jjhong922, line 178: I would split up the above and below into separate functions for readability purposes, the above being formatting the config dictionaries, and the below actually running the model with the respective configuration.

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,346 @@
...
+            ],
+            categorical_covariate_keys=categorical_covariate_config[
+                "categorical_covariates"
+            ],
+        )
+
+        best_model = self.model(self.adata, **model_config)
+        best_model.load(adata=self.adata, dir_path=best_checkpoint + "checkpoint")
+        if train:
+            best_model.train(
```
jjhong922, line 339: The code in this function is repeated from `_trainable`, please factor it out

scvi/autotune/_metrics.py
```diff
@@ -0,0 +1,86 @@
...
+    Sum of asw for batch and for labels. Scores are scaled such that
+    2 is the best score and 0 is the worst.
+    Notes
+    -----
+    This function is influence by the following code:
+    https://github.com/theislab/scib/blob/main/scib/metrics/silhouette.py
+    """
+    model.is_trained_ = True
+    latent = model.get_latent_representation()
+    model.is_trained_ = False
```
jjhong922, line 40: What is the reason for switching this off?

jjhong922: Oh I see, could you add a note for the reason behind it?

scvi/autotune/_wrappers.py
```diff
@@ -0,0 +1,42 @@
+import anndata
+from ray import tune
+from ray.tune.schedulers import ASHAScheduler
+
+from scvi.autotune import Autotune
+from scvi.model import SCVI
+
+
+def tune_scvi(adata: anndata.AnnData, n_epochs, run_kwargs):
```
jjhong922, line 9: nit: typing for n_epochs and run_kwargs

tests/autotune/test_autotune.py
```diff
@@ -0,0 +1,324 @@
...
+    )
+
+
+def test_complex(save_path):
+    adata = synthetic_iid()
+    adata.obs["continious_cov"] = np.linspace(0, 1, adata.n_obs)
+    adata.obs["categorical_cov"] = pd.Categorical(
+        ["cov_1", "cov_2"] * (round(adata.n_obs / 2))
+    )
+    metrics = [
```
jjhong922, line 292: There's a lot of repeated code in this file. Could you move out some constants like metric_functions, model_config, plan_config to the top of the file since they are consistent across all tests?

jjhong922 (commented):
Thanks for the update on this PR and the extensive testing! To add ray tune as a dependency, we would want to add it to the `pyproject.toml` file as an optional dependency, then create a new `extras` keyword `tune` under which you put `ray`. You can see how `pymde` was handled in that file as an example.

I'll let @adamgayoso  or @PierreBoyeau take a pass as well.

scvi/autotune/_autotune.py
nrclaudio: Good catch. I've included a `self.setup_signature` to the Autotune object that we can use to set covariates to non-None values. I've also added a user warning for the latter (very valid) point you make. Maybe we can consider raising an error if we just don't want to support this behaviour?

scvi/autotune/_metrics.py
nrclaudio: Included the reasoning as a comment, were you referring to a note in the docstring instead?

nrclaudio:
> Thanks for the update on this PR and the extensive testing! To add ray tune as a dependency, we would want to add it to the `pyproject.toml` file as an optional dependency, then create a new `extras` keyword `tune` under which you put `ray`. You can see how `pymde` was handled in that file as an example.
> 
> I'll let @adamgayoso or @PierreBoyeau take a pass as well.

I've tried changing `pyproject.toml`, but it seems that the job it's not installing `tune`'s dependencies during `pytest`. Let me know if I did something wrong there

jjhong922:
We'd like to retain the commit history for the original ray tune PR. Hopefully this isn't too much work, but could you try rebasing on `michael/autotune` and updating this PR? You wouldn't change the base since we would just merge the commits together in one PR to master. I just updated the `michael/autotune` PR to latest master, but tests are failing since I didn't adapt it to the new setup scheme which you did in this PR.

nrclaudio:
I see @adamgayoso included autotune.srt already for documentation, I can help draft it if you want.

scvi/autotune/_wrappers.py
```diff
@@ -0,0 +1,47 @@
...
+    metrics = [
+        "elbo_validation",
+        "reconstruction_loss_validation",
+    ]
+    model_config = {
+        "dropout_rate": tune.loguniform(1e-4, 1e-1),
+        "n_layers": tune.choice([i for i in range(1, 5)]),
+        "n_hidden": tune.choice([64, 128, 256]),
+        "n_latent": tune.choice([5, 10]),
+    }
```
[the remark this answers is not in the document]
nrclaudio: Indeed 5 and 10 seem like too low of a value. I usually use values between 20-50. Maybe we change this to increments of 5 between 20 and 50?

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,307 @@
...
+            else None
+        }
+
+        if test_effect_hvg:
+            self.test_effect_hvg = {"subset_adata": tune.choice([True, False])}
+        else:
+            self.test_effect_hvg = {"subset_adata": False}
+
+        # Using conditional search spaces is not supported in all space search algorithms
+        # If necessary we would need to re-factor to Optuna / HyperOptSearch spaces
```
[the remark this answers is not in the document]
nrclaudio: Yes, of course. It's a bit complicated to grasp but I hope I can explain it well.

First, `test_effect_hvg` can be set by the user to either `True` or `False`. If it is set to `True` we want to reflect this in the search space with the `self.test_effect_hvg` dictionary key. Since we want to test the effect of filtering for HVGs, we randomly generate `True` or `False` values that will be assigned to `self.test_effect_hvg`. If the user sets the `test_effect_hvg` to `False` ,  `self.test_effect_hvg` will be just set to `False` without using ray search spaces. 

Here the conditional bit is found in lines 142-148:

```
 self.top_hvg = {
            "top_n": tune.sample_from(
                lambda spec: np.random.choice(top_hvg)
                if spec.config["subset_adata"]
                else None
            )
        }
```
See: [ray tune conditional search spaces](https://docs.ray.io/en/master/tune/tutorials/tune-search-spaces.html#how-to-use-custom-and-conditional-search-spaces). Here `if spec.config["subset_adata"]` is checking the value of this variable in the search space **while** it is being created. 

I think it's easier to understand if we imagine the case without conditional search spaces. In this case the snippet would look like this:

```
 self.top_hvg = {
            "top_n": tune.sample_from(
                lambda spec: np.random.choice(top_hvg)
                if self.test_effect_hvg
                else None
            )
        }
```

Note the use of `self.test_effect_hvg` instead of `spec.config["subset_adata"]`. In this case, if `self.test_effect_hvg` is set to `True` (i.e. the user wants to test the effect of hvg filtering) then self.top_hvg will always be set to a given number of `top_hvg`, even though we would like to test also the effect of not filtering. When using `spec.config["subset_adata"]` is checking the value of this variable in the search space **while** it is being created, so even if the user sets `test_hvg_effect` to `True`, the search space will still consider not filtering for `top_hvg`. 

I reckon it is a bit too much over text, let me know if we need to discuss this further

scvi/autotune/_wrappers.py
```diff
@@ -0,0 +1,47 @@
...
+from ray.tune.schedulers import ASHAScheduler
+
+from scvi.autotune import Autotune
+from scvi.model import SCVI
+
+
+def tune_scvi(adata: anndata.AnnData, n_epochs: int, run_kwargs: Optional[dict] = None):
+    """
+    Tune scvi with defaults for `tune.run` and return the best model.
+    Parameters
```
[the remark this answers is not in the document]
jjhong922: there should be a note in this function's docstring specifying that dropout_rate, n_layers, n_hidden, and n_latent are optimized. The autotune class is more generalized so I don't think you need to add anything there

scvi/autotune/_autotune.py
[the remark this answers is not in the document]
jjhong922: thanks for the clarification, we can stick to using this implementation for now and refactor if necessary

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,307 @@
...
+        top_hvg = top_hvg or [None]
+
+        if test_effect_hvg and not top_hvg:
+            raise ValueError(
+                "test_effect_hvg is set to True but no list of top_hvgs was given. Please"
+                " provide a list"
+            )
+
+        self.adata = adata
+        self.model = model
```
jjhong922, line 103: lets name this `model_cls` to avoid confusion with an instance

scvi/autotune/_callbacks.py
```diff
@@ -0,0 +1,62 @@
...
+from ray import tune
+from ray.tune.integration.pytorch_lightning import TuneCallback
+
+from scvi.model.base import BaseModelClass
+
+
+class ModelSave(Callback):
+    def __init__(self, model: BaseModelClass):
+        super()
+        self.model = model
```
jjhong922, line 15: `model_cls` here also

jjhong922 (commented):
thanks for addressing all my comments, I think once these is rebased on the latest master and some minor comments are resolved itll be close to being merged! @adamgayoso let me know if you have any reservations about getting this in

jjhong922:
Also, please resolve the codacy checks if you can

jjhong922 (approved):
Approving to unblock. @adamgayoso, my proposal is to get this into master and have it ready for v0.18.  We can use the master branch to test our hyperparams and to make sure everything works smoothly and include fixes as necessary for v0.18. Just need to ignore the existing codacy errors to get it merged.

jjhong922:
Also the docs are still broken, that should be fixed before merging

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,289 @@
...
+
+from ._utils import apply_model_config, fetch_config, train_model
+
+logger = logging.getLogger(__name__)
+
+
+class Autotune:
+    """
+
+    Hyperparameter tuning using Ray Tune.
```
adamgayoso, line 27:
```suggestion
    """
    Hyperparameter tuning using Ray Tune.
    
    PARAGRAPH DESCRIBING THIS CLASS
```

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,289 @@
...
+            )
+
+        if not resources_per_trial:
+            if torch.cuda.is_available():
+                resources_per_trial = {"cpu": 1, "gpu": 1}
+            else:
+                resources_per_trial = {"cpu": 1}
+
+        if resources_per_trial["cpu"] > 1 and resources_per_trial["gpu"]:
+            resources_per_trial["gpu"] /= resources_per_trial["cpu"]
```
adamgayoso, line 233: what is this doing? Can you add a comment?

scvi/autotune/_utils.py
```diff
@@ -0,0 +1,126 @@
+from typing import Tuple, Union
+
+import scanpy as sc
```
adamgayoso, line 3: scanpy is an optional dependency of scvi-tools, so we need to do a lazy import

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,289 @@
+"""Main class and helper functions."""
+import inspect
+import logging
+import os
+import warnings
+from typing import List, Optional, Union
+
+import anndata
+import numpy as np
+import ray
```
adamgayoso, line 10: since this is an optional dependency we should be able to capture the error and tell people to install it

adamgayoso (changes requested):

scvi/autotune/_autotune.py
```diff
@@ -0,0 +1,289 @@
...
+    continuous_covariates
+        continuous covariates from `adata` to tune the model on
+    categorical_covariates
+        Categorical covariates from `adata` to tune the model on
+    test_effect_hvg
+        Whether to test the effect of filtering adata with highly variable genes on model performance or not. Default is `False`.
+    top_hvg
+        Top hvgs to test. If `test_effect_hvg` is set, defaults to top 2000, 2500, 3000 and 3500 highly variable genes
+    batch_key_hvg
+        Column in adata.obs specifying the different batches in the data used for highly_variable_gene selection
```
adamgayoso, line 60: I think this is a bit overloaded. I think this API would be easier to use if Autotune did not alter the data/ setup anndata hyperparams at all, because one can always run multiple autotune sessions for different data configs. These kwargs are a bit too specific to scvi as it currently is.

An alternative is to provide a function that alters the data as input and a config or something, but I think we can leave that for future work

adamgayoso: Therefore I think we should simplify this functionality and only have setup_kwargs. Wyt? @PierreBoyeau @jjhong922

adamgayoso (changes requested):

scvi/autotune/_autotune.py
nrclaudio: Indeed. As it is, it might be too much functionality packed in the class initialization. However, I think the ability to test different anndata configurations is quite handy, so your suggestion to factor it out as a separate function can be the sweet spot, maybe something like `.test_configurations()`? 

Let me know what you all think about this, so I can factor it out and put it as a separate block of code for future use.

pyproject.toml
```diff
@@ -67,6 +68,7 @@ python = ">=3.7,<4.0"
 python-igraph = {version = "*", optional = true}
 pytorch-lightning = ">=1.6.4,<1.7.0"
 rich = ">=9.1.0"
+ray = {extras = ["tune"], version = ">=1.12.1", optional = true}
```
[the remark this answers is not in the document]
nrclaudio: As is right now, the API is not using HyperOpt due to https://github.com/scverse/scvi-tools/pull/1565#discussion_r914928902. However, if we factor the anndata configurations out from `.__init__()` we can probably include it, although I haven't tested this myself.

adamgayoso:
@nrclaudio we will merge this into a staging branch and complete it on our end. Thank you for the contribution!!

PR #1574, opened 2022-06-22
Break up all the anndata tests

RK900:


tests/dataset/test_anndata.py
```diff
@@ -304,6 +310,8 @@ def test_setup_anndata():
         adata_manager.get_from_registry(REGISTRY_KEYS.X_KEY), true_x
     )
 
+
+def test_setup_anndat_create_label_batch():
```
jjhong922, line 314: typo

jjhong922 (approved):
lgtm

jjhong922:
agree w/ rohan here, caching the same anndata object across tests can lead to unexpected behavior. At best, we can have a fixture to create an anndata object consistent across tests

PR #1576, opened 2022-06-23
Use pytest fixtures for anndata tests

RK900:
referenced in #1574

### DIFF
--- a/tests/dataset/conftest.py
+++ b/tests/dataset/conftest.py
@@ -3,6 +3,7 @@ import pytest
 
 from scvi.data import synthetic_iid
 
+
 @pytest.fixture
 def adata():
     anndata = synthetic_iid()
@@ -14,5 +15,6 @@ def adata():
     anndata.obs["cat2"] = np.random.randint(0, 5, size=(anndata.shape[0],))
 
     return anndata
+
 
 adata1 = adata2 = adata
--- a/tests/dataset/conftest.py
+++ b/tests/dataset/conftest.py
@@ -1,9 +1,7 @@
 import pytest
 import numpy as np
 
-import scvi
 from scvi.data import synthetic_iid
-from .utils import generic_setup_adata_manager
 
 @pytest.fixture
 def adata():
--- a/tests/dataset/conftest.py
+++ b/tests/dataset/conftest.py
@@ -1,0 +1,21 @@
+import pytest
+import numpy as np
+
+import scvi
+from scvi.data import synthetic_iid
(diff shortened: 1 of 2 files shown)


tests/dataset/test_anndata.py
```diff
@@ -329,25 +310,22 @@ def test_setup_anndat_create_label_batch():
...
     adata.obs["batch"][:10] = np.nan
     with pytest.raises(ValueError):
         generic_setup_adata_manager(adata, batch_key="batch")
 
 
-def test_setup_anndata_cat():
+def test_setup_anndata_cat(adata):
     # test error is thrown when categorical joint obsm field contains nans
-    adata = synthetic_iid()
     adata.obs["cat1"] = np.random.randint(0, 5, size=(adata.shape[0],))
```
jjhong922, line 322: this stuff for example can be factored out

tests/dataset/test_anndata.py
```diff
@@ -489,8 +462,7 @@ def test_anntorchdataset_getitem_pro_exp():
         assert type(value) == np.ndarray
 
 
-def test_view_registry():
-    adata = synthetic_iid()
+def test_view_registry(adata):
     adata.obs["cont1"] = np.random.normal(size=(adata.shape[0],))
```
jjhong922, line 466: same here, wont comment for rest of the file

jjhong922 (commented):
thanks! Left some comments, look for more places to extract code. Remember the fixture can contain more setup than currently exists in every tests. We want each fixture to cover as much of the setup for 90% of tests that need them.

RK900:
I tried factoring out the model training into a fixture, but that resulted in test_anndata taking 4->14 seconds to run

tests/dataset/conftest.py
```diff
@@ -0,0 +1,20 @@
+import numpy as np
+import pytest
+
```
jjhong922, line 3: we will want to use this fixture in other test files like test_models.py for example, so probably want to put it in a higher level conftest

jjhong922 (approved):
looks good to me, you can ignore the comment for now and move it up when you make the changes to other test files.

jjhong922:
> I tried factoring out the model training into a fixture, but that resulted in test_anndata taking 4->14 seconds to run

did you use the fixture on every test, or was it only used for tests that did this anyways? If its the latter I dont understand why the runtime would increase so much.

RK900:
It was only for the tests that required a model to be trained.

PR #1580, opened 2022-06-23
[Jax] Remove is_training from JaxVAE constructor

### LINKED ISSUE #1564: [Jax] Pass in `is_training` during module call rather than on module init
Based on https://flax.readthedocs.io/en/latest/design_notes/arguments.html, we should be able to avoid the pattern of reinitializing modules with `is_training` reset to the desired state: https://github.com/scverse/scvi-tools/blob/master/scvi/model/_jaxscvi.py#L121-L130.

Rather than having `is_training` as a [dataclass attribute](https://github.com/scverse/scvi-tools/blob/73952af10292852b8e5b7319409678cc66d85a47/scvi/module/_jaxvae.py#L117), we can have it passed down as a kwarg to [`__call__`](https://github.com/scverse/scvi-tools/blob/master/scvi/module/base/_base_module.py#L399). All linen modules with these dynamic hyperparameters have optional dataclass attributes with the suggested pattern (e.g. dropout https://flax.readthedocs.io/en/latest/_modules/flax/linen/stochastic.html#Dropout.__call__).

The main con is that this will require explicit piping of the `is_training` kwarg through all parts of `inference` and `generative` methods.

RK900:
Fixes #1564


scvi/model/_jaxscvi.py
```diff
@@ -158,6 +158,12 @@ def get_latent_representation(
 
         return np.array(jax.device_get(latent))
 
+    def train(self):
```
jjhong922, line 161: remove

scvi/module/_jaxvae.py
```diff
@@ -125,16 +117,20 @@ def setup(self):
...
         )
 
         self.decoder = FlaxDecoder(
             n_input=self.n_input,
             dropout_rate=0.0,
             n_hidden=self.n_hidden,
-            is_training=self.is_training,
         )
 
+    def train(self):
```
jjhong922, line 128: I don't think this is ever called right now? This PR should require changes to JaxModuleWrapper which has train() eval() functions. The trainer calls train() and eval() on the JaxModuleWrapper which currently wont be doing anything with respect to this parameter. Definitely add some tests to ensure this is working before going further.

scvi/module/_jaxvae.py
```diff
@@ -146,7 +142,7 @@ def _get_inference_input(self, tensors: Dict[str, jnp.ndarray]):
         return input_dict
 
     def inference(self, x: jnp.ndarray, n_samples: int = 1) -> dict:
-        mean, var = self.encoder(x)
+        mean, var = self.encoder(x, is_training=self.is_training)
```
jjhong922, line 145: I prefer if is_training is pushed through the inference and generative functions. Right now it would require new JaxModules to handle is_training state on its own. Ideally, the Flax module should not hold any state at all and remain entirely pure. This is why we have the JaxModuleWrapper to hold state and wrap the flax module.

jjhong922 (changes requested):
Left some comments on how to clean up the flow of the is_training parameter. I suggest taking a close look at JaxTrainingPlan to understand what happens when a user calls model.train() from front to end.

scvi/module/base/_jax_module_wrapper.py
```diff
@@ -40,6 +40,7 @@ class JaxModuleWrapper:
     def __init__(
         self,
         module_cls: JaxBaseModuleClass,
+        is_training: bool = False,
```
jjhong922, line 43: this doesnt need to be here, just initialize self.is_training to True (pytorch convention)

jjhong922 (approved):

jjhong922:
meeseeksdev backport to 0.17.x

PR #1595, opened 2022-07-09
Max kl weight

### LINKED ISSUE #1569: kl_weight in scVI for more than 20,000 cells as input
Hi all,

thanks for providing such a great tool! I think there is a problem in scVI with handling n_epochs_kl_warmup. If the number of cells in the AnnData is larger than 20,000,the maximum number of epochs defaults to a value smaller than n_epochs_kl_warmup. This basically turns scVI into a betaVAE with beta < 1. The problem is that in UnsupervisedTrainingMixin it is never validated that the n_epochs_kl_warmup is smaller than max_epochs.

Code to reproduce the issue (runs ~10 minutes with a GPU):

```import numpy as np
import scvi
import anndata
import pytest
import scvi.train._trainingplans

_compute_kl_weight = scvi.train._trainingplans._compute_kl_weight


def kl_weight(*args, **kwargs):
    global weight
    kl_weight = _compute_kl_weight(*args, **kwargs)
    weight = kl_weight
    return kl_weight


def test_linear_annealing(monkeypatch):
    monkeypatch.setattr("scvi.train._trainingplans._compute_kl_weight", kl_weight)
    adata = anndata.AnnData(X=np.ones((25_000, 2)))
    scvi.model.SCVI.setup_anndata(adata)
    model = scvi.model.SCVI(adata)
    model.train()
    assert np.isclose(weight, 1.0)
```

This problem is easily fixed by setting the number of n_epochs_kl_warmup to min(n_epochs_kl_warmup, max_epochs). This should be done in the train method of UnsupervisedTrainingMixin because we need access to the batch size and the number of cells to calculate max steps to also ensure that n_steps_kl_warmup is smaller than max steps.

I am happy to provide a PR that fixes this.

Best,
Florian

FlorianBarkmann:
Fixes #1569. 

It introduces the max_kl_weight and exposes the min_kl_weight. Furthermore, it adds a warning if max_kl_weight is not reached during training.


scvi/model/base/_training_mixin.py
```diff
@@ -1,11 +1,76 @@
...
+        max_steps = max_epochs * steps_per_epoch
+        if n_steps_kl_warmup and max_steps < n_steps_kl_warmup:
+            warnings.warn(
+                _WARNING_MESSAGE.format(
+                    mode="steps", max=max_steps, warm_up=n_steps_kl_warmup
+                )
+            )
+        return
+
+    if n_epochs_kl_warmup:
```
jjhong922, line 60: this can just be an elif and the return above can be removed

tests/models/base/test_traning_mixin.py
```diff
@@ -0,0 +1,41 @@
+import warnings
```
jjhong922, line 1: typo in the file name

tests/models/base/test_traning_mixin.py
```diff
@@ -0,0 +1,41 @@
...
+
+import pytest
+
+from scvi.model.base._training_mixin import _check_warmup
+
+
+@pytest.mark.parametrize(
```
jjhong922, line 8: this is wonderful!

jjhong922 (commented):
This looks great! If you can just fix the nits and add an entry into the v0.17.0 release note (feel free to add yourself as a contributor for this release), we will include it in our upcoming this week (hopefully this week).

FlorianBarkmann:
Thanks for your review. I have made all the changes and updated the release notes.

jjhong922 (approved):

PR #1596, opened 2022-07-11
Set full state update to false in `ElboMetric`

jjhong922:
Removes warning from torchmetrics, checked with their util function `check_forward_full_state_property`


adamgayoso (approved):
LGTM can you just add a comment as to why that is there?

Issue #1578, opened 2022-06-23
Implementation of PeakVI in Jax


justjhong:
Reimplementation of PeakVI in Jax. This is meant to both see if we can get a similar speed up in PeakVI, as well as understanding the limitations of the `JaxModuleWrapper` abstraction.

PR #1608, opened 2022-07-18
Basic JaxPEAKVI class

RK900:
Partially addresses #1578


jjhong922 (changes requested):
good start, but need to be very sure the model is implemented in parallel with the torch implementation. Implementing get_latent should help test this.

scvi/module/_jaxpeakvae.py
```diff
@@ -0,0 +1,100 @@
...
+        tensors,
+        inference_outputs,
+        generative_outputs,
+        kl_weight: float = 1.0,
+    ):
+        x = tensors[REGISTRY_KEYS.X_KEY]
+        px = generative_outputs["px"]
+        qz = inference_outputs["qz"]
+
+        reconstruction_loss = -px.log_prob(x).sum(-1)
```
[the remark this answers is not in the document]
RK900: https://optax.readthedocs.io/en/latest/api.html?highlight=binary#optax.sigmoid_binary_cross_entropy

Do we want to use optax for these sorts of things?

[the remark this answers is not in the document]
jjhong922: that's fine as long as the sign and normalization match torch's

RK900:
Screen Shot 2022-07-28 at 8 08 49 PM

Latent space from JPVI, matching the one from the [PeakVI tutorial](https://docs.scvi-tools.org/en/0.15.1/tutorials/notebooks/PeakVI.html)

scvi/module/_jaxpeakvae.py
```diff
@@ -0,0 +1,201 @@
...
+    dropout_rate: float
+    n_hidden: int
+    n_output: int = None
+
+    def setup(self):
+        self.dense1 = Dense(self.n_hidden)
+        self.dense2 = Dense(self.n_hidden)
+        self.dense3 = Dense(self.n_hidden)
+        self.dense4 = Dense(self.n_hidden)
+        if self.n_output is None:
```
jjhong922, line 62: do we need n_input if it's inferred? Seems like we just need n_output to be defined for every instance

scvi/module/_jaxpeakvae.py
```diff
@@ -0,0 +1,201 @@
...
+            self.dense5 = Dense(self.n_input)
+        else:
+            self.dense5 = Dense(self.n_output)
+
+        self.batchnorm1 = nn.BatchNorm()
+        self.batchnorm2 = nn.BatchNorm()
+        self.dropout1 = nn.Dropout(self.dropout_rate)
```
jjhong922, line 69: dropout is deterministically set to 0 in peakvae, we can just remove this? same with dropout2

scvi/module/_jaxpeakvae.py
```diff
@@ -0,0 +1,201 @@
...
+class FlaxEncoder(nn.Module):
+    n_input: int
+    n_latent: int
+    n_hidden: int
+    dropout_rate: int
+
+    def setup(self):
+        self.dense1 = Dense(self.n_hidden)
+        self.dense2 = Dense(self.n_hidden)
+        self.dense3 = Dense(self.n_latent)
```
jjhong922, line 23: provide more descriptive names given that these are not sequential like the names would suggest otherwise

scvi/module/_jaxpeakvae.py
```diff
@@ -0,0 +1,201 @@
...
+    n_input: int
+    dropout_rate: float
+    n_hidden: int
+    n_output: int = None
+
+    def setup(self):
+        self.dense1 = Dense(self.n_hidden)
+        self.dense2 = Dense(self.n_hidden)
+        self.dense3 = Dense(self.n_hidden)
+        self.dense4 = Dense(self.n_hidden)
```
jjhong922, line 61: i think this should only be 2 layers by default

scvi/module/_jaxpeakvae.py
```diff
@@ -0,0 +1,201 @@
...
+
+        input_dict = dict(x=x, batch_index=batch_index)
+        return input_dict
+
+    def inference(self, x: jnp.ndarray, batch_index, n_samples: int = 1) -> dict:
+        mean, var = self.encoder(x, training=self.training)
+        stddev = jnp.sqrt(var) + self.eps
+
+        batch = jax.nn.one_hot(batch_index, self.n_batch).squeeze(-2)
+        d = self.d_encoder(x, batch, training=self.training)
```
jjhong922, line 145: can we make this an optional encoder with the `model_depth` kwarg

tests/models/test_models.py
```diff
@@ -87,6 +88,23 @@
 }
 
 
+def test_jax_peakvi():
+    n_latent = 5
+
+    adata = synthetic_iid()
```
jjhong922, line 94: use fixture

jjhong922 (commented):
latent space is looking good! some comments on more discrepancies with the torch implementation. Doesn't have to have full functionality, but the least we can do is get the defaults right. Another necessary check is to ensure the elbo values are in the same range as the torch implementation.

scvi/module/_jaxpeakvae.py
```diff
@@ -0,0 +1,201 @@
...
+
+import jax
+import jax.numpy as jnp
+import numpyro.distributions as dist
+from flax import linen as nn
+
+from scvi import REGISTRY_KEYS
+from scvi.module.base import JaxBaseModuleClass, LossRecorder
+
+from ._jaxvae import Dense
```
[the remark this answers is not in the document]
RK900: Moved it to _base_components

tests/models/test_models.py
RK900: I'm gonna refactor these tests in a separate PR, I think I can add fixtures for the scvi tests as well

scvi/module/_jaxpeakvae.py
RK900: In peakvi the default dropout is 0.1

scvi/module/_jaxpeakvae.py
RK900: I think we need all of these layers, one for x and one for batch

scvi/module/_jaxpeakvae.py
jjhong922: This is only true in the encoder, for the decoder its set to 0: https://github.com/scverse/scvi-tools/blob/master/scvi/module/_peakvae.py#L62

PR #1618, opened 2022-07-28
Likelihood params bugfix

tommycelsius:
Bug:
When using `gene_likelihood!="zinb"`, the function `get_likelihood_parameters()` would fail.

To reproduce:
```
model = SCVI(adata, gene_likelihood="nb")
model.get_likelihood_parameters()
```

Fix:
Added `if self.module.gene_likelihood == "zinb":` to a few lines.


adamgayoso (approved):
LGTM, thanks!

PR #1636, opened 2022-08-11
fix pyro global step tracking

### LINKED ISSUE #1646: Error in scvi.model.AmortizedLDA.load
When trying to load saved model using 'scvi.model.AmortizedLDA.load' it again starts training the model. 

```python
model_loaded = scvi.model.AmortizedLDA.load("SCC_Topic_model/", adata=adata, use_gpu=True)
```

```pytb
INFO     File SCC_Topic_model/model.pt already downloaded                                    
GPU available: True, used: True
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs
/home/nileshkr/.conda/envs/Hackin_Omics/lib/python3.10/site-packages/pytorch_lightning/trainer/configuration_validator.py:131: UserWarning: You passed in a `val_dataloader` but have no `validation_step`. Skipping val loop.
  rank_zero_warn("You passed in a `val_dataloader` but have no `validation_step`. Skipping val loop.")
LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]
```

#### Versions:

> '0.17.1'

adamgayoso:
The issue here is that PyTorch Lightning uses its wrapper of an optimizer to track steps. It uses post step hooks to increment the global step of the trainer.

Here we provide a shimmy optimizer for it to do this.

Fixes #1646


vitkl:
Thanks for solving this issue! I also get this issue with non-cell2location scvi-tools model. So would be great if this fix is approved and merged.

watiss (approved):

PR #1637, opened 2022-08-12
Update doctrings for continuous and categorical covariate docstrings

watiss:
There were sources of confusion among more than one users about how the model consumes this, so this is adding some more context to the docstrings to provide clarity on that.


RK900 (approved):

adamgayoso:
meeseeksdev backport to 0.17.x

PR #1639, opened 2022-08-16
Fix toalvi `max_epochs`

### LINKED ISSUE #1638: `max_epochs` is never determined heuristically for totalvi, instead it always defaults to 400
Below is an excerpt of the code in totalvi. The intention seems to be to determine `max_epochs` based on a heuristic, but instead the value always defaults to 400 due to the way the kwarg is declared.

`def train(max_epochs: Optional[int] = 400):
        if max_epochs is None:
            n_cells = self.adata.n_obs
            max_epochs = np.min([round((20000 / n_cells) * 400), 400])`

watiss:
Fixes #1638

### DIFF
--- a/scvi/model/_totalvi.py
+++ b/scvi/model/_totalvi.py
@@ -192,7 +192,7 @@ class TOTALVI(RNASeqMixin, VAEMixin, ArchesMixin, BaseModelClass):
 
     def train(
         self,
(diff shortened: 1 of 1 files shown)


adamgayoso:
We should note that this is a breaking change , but it should be fine

adamgayoso (approved):

PR #1645, opened 2022-08-22
Remove poetry lock, use newer build system

adamgayoso:
- Update poetry build system
- Remove lock file, as it's not useful given the many number of ways pytorch/jax can be installed

The first change still allows `pip install -e .` and this is the recommended way to install (in a virtualenv)

See https://github.com/python-poetry/poetry/issues/34 for more discussion (specifically the end)


watiss:
Do we need to update scvi documentation https://docs.scvi-tools.org/en/stable/contributing/index.html ?

watiss (approved):

adamgayoso:
> Do we need to update scvi documentation https://docs.scvi-tools.org/en/stable/contributing/index.html ?

Nope, all these things still work, maybe the poetry install is not recommended, but I think this is fine.

PR #1649, opened 2022-08-23
Fix `encode`/`decode` calls in JVAE by replacing them with `inference`/`generative` calls

### LINKED ISSUE #1588: missing "encode" function in JVAE class of gimVI
The following `encode` function used to be in the 
`jvae.py` in scvi-tools version 0.8.0 is missing in the current scvi.external.GIMVI module:

```python 
def encode(
        self, x: torch.Tensor, mode: int
    ) -> Tuple[
        torch.Tensor,
        torch.Tensor,
        torch.Tensor,
        Optional[torch.Tensor],
        Optional[torch.Tensor],
        torch.Tensor,
    ]:
        x_ = x
        if self.log_variational:
            x_ = torch.log(1 + x_)

        qz_m, qz_v, z = self.z_encoder(x_, mode)
        ql_m, ql_v, library = None, None, None
        if self.model_library_bools[mode]:
            ql_m, ql_v, library = self.l_encoders[mode](x_)
        else:
            library = torch.log(torch.sum(x, dim=1)).view(-1, 1)

        return qz_m, qz_v, z, ql_m, ql_v, library
```

Causing GIMVI.get_imputed_values(normalized = False) to fail in the following logic of imputing unnormalized counts:

```python
    GIMVI.setup_anndata(spatial_data_partial)
    GIMVI.setup_anndata(seq_data)
    model = GIMVI(seq_data, spatial_data_partial)    
    model.train(200)
    _, imputation = model.get_imputed_values(normalized = False) # calls JVAE.sample_rate(), involving a call to self.encode, which is missing

```pytb
[ERROR HERE]
```

#### Versions:

> 0.16.4

watiss:
Fixes #1588

For posterity: `encode` and `decode` functions were removed in [this](https://github.com/scverse/scvi-tools/commit/014ec33a9cd51bd6d9dd15fca4d8ef2b95d4f930) commit (but their call sites were not updated)

### DIFF
--- a/scvi/external/gimvi/_module.py
+++ b/scvi/external/gimvi/_module.py
@@ -227,10 +227,12 @@ class JVAE(BaseModuleClass):
             tensor of shape ``(batch_size, 1)``
 
         """
-        _, _, _, ql_m, _, library = self.encode(x, mode)
-        if deterministic and ql_m is not None:
-            library = ql_m
-        return library
+        inference_out = self.inference(x, mode)
+        return (
+            inference_out["ql"].loc
+            if (deterministic and inference_out["ql"] is not None)
+            else inference_out["library"]
+        )
 
     def sample_scale(
         self,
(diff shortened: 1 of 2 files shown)


adamgayoso (changes requested):
Can we add a test to ensure this doesn't happen again?

adamgayoso (approved):

PR #1654, opened 2022-08-24
Fix more missing `max_epochs` heuristics

watiss:
Much like #1639


adamgayoso:
Closing as we should not make so many breaking changes all at once and should discuss other options before going this route.

PR #1655, opened 2022-08-24
Update 0.17.2 rel notes

watiss:


adamgayoso (changes requested):
We should change the date to friday

adamgayoso (approved):

PR #1660, opened 2022-08-26
Fix typo in the logic for throwing exception in the VAE module

### LINKED ISSUE #1659: Small typo error in exception throwing of VAE module class
In the VAE module class line 130, if ´self.use_observed_lib_size´== False, the condition for throwing the exception should be
```if library_log_means is None or library_log_vars is None``` instead of ```if library_log_means is None or library_log_means is None```
https://github.com/scverse/scvi-tools/blob/04186b50b3f929d6bdd4c7c65d9ef37bbec047fd/scvi/module/_vae.py#L130

#### Versions:
master branch

watiss:
Fixes #1659


adamgayoso (approved):
LGTM

Issue #1670, opened 2022-09-13
report [and proposed fix] for autoreload error caused by explicit module.super(ChildClass,self)__init__()


mkarikom:
I noticed that autoreload fails when updating class methods in the api_overview.ipynb, and that this can be fixed by removing the explicit child class parameter in the `super()__init__()` call.  In my case, I was debugging an updated TrainingPlan, but I suspect this will be effective for other classes as well.

I can submit a PR for this but wanted to make sure it wasn't already being addressed somehow.

1. add a cell in `api_overview.ipynb` with
```python
%load_ext autoreload
%autoreload 2
```
2. instantiate and train the `model`
3. introduce a bug (eg in `train.TrainingPlan`, inherited by the already-instantiated `model`) like:
```python
    def training_step(self, batch, batch_idx, optimizer_idx=0):
        if "kl_weight" in self.loss_kwargs:
            self.loss_kwargs.update({"kl_weight": self.kl_weight})
        _, _, scvi_loss = self.forward(batch, loss_kwargs=self.loss_kwargs)
        self.log("train_loss", scvi_loss_not_defined.loss, on_epoch=True) # this will cause an error
        self.compute_and_log_metrics(scvi_loss, self.train_metrics, "train")
        return scvi_loss.loss
```
which tries to log an undefined key and gives the following error
```python
/workspaces/scvi_work_pip/scvi-tools/scvi/model/base/_training_mixin.py:67: UserWarning: max_epochs=4 is less than n_epochs_kl_warmup=400. The max_kl_weight will not be reached during training.
  warnings.warn(
GPU available: True, used: True
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs
LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0,1]
Epoch 1/4:   0%|          | 0/4 [00:00<?, ?it/s]/workspaces/scvi_work_pip/scvi-tools/scvi/distributions/_negative_binomial.py:64: UserWarning: Specified kernel cache directory could not be created! This disables kernel caching. Specified directory is /home/vscode/.cache/torch/kernels. This warning will appear only once per process. (Triggered internally at  ../aten/src/ATen/native/cuda/jit_utils.cpp:860.)
  + torch.lgamma(x + theta)
[...]

PR #1671, opened 2022-09-13
change to super().__init__()

mkarikom:
global super().__init__() referenced in #1670

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,10 +1,10 @@
 repos:
     - repo: https://github.com/psf/black
-      rev: 22.6.0
+      rev: 22.8.0
       hooks:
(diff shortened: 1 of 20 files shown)


mkarikom:
Just noticed the [contributing guidelines](https://docs.scvi-tools.org/en/0.15.1/contributing/index.html#writing-a-github-release) after the docs check failed, and tried to make a release with an existing tag (0.17.1), then update the commit.

I admit that I don't really understand what Releases are and haven't made one before, lol.

adamgayoso:
Thanks, this looks good. We are having some issue with one of our tutorials that is causing the docs to fail so this might hang for a bit.

adamgayoso (approved):

PR #1672, opened 2022-09-13
Add latent mode support in SCVI

watiss:
Fixes https://github.com/scverse/scvi-tools/issues/1663


scvi/module/base/_base_module.py
```diff
@@ -197,6 +201,11 @@ def inference(
         """
         pass
 
+    @abstractmethod
+    def inference_no_encode(self, qz_m, qz_v, latent_data_type, n_samples=1):
+        """Runs the inference routine but skips encoding, used in latent mode."""
+        pass
+
```
adamgayoso, line 208: this is going to break a lot of peoples code, we should not make this abstract and instead can just add a notimplementederror?

scvi/model/_scvi.py
```diff
@@ -188,4 +203,72 @@ def setup_anndata(
...
+            adata_manager.register_new_fields(cls._get_latent_fields(latent_mode))
         cls.register_manager(adata_manager)
+
+    def _get_latent_adata_from_adata(
+        self,
+        mode: LatentDataType,
+        use_latent_key: str = "X_latent",
+        use_latent_qzm_key: str = "X_latent_qzm",
+        use_latent_qzv_key: str = "X_latent_qzv",
+    ) -> AnnData:
```
adamgayoso, line 218: to_latent_mode() should just by default add the latents to the existing adata object attached to the model

we can add a memory reduce option that inplace deletes adata.raw and substitutes layers/x for csr matrix. If the user doesn't want inplace there can be a copy option

scvi/model/_scvi.py
```diff
@@ -188,4 +203,72 @@ def setup_anndata(
...
+        all_zeros = csr_matrix(self.adata.X.shape, dtype=np.int8).toarray()
```
adamgayoso, line 219:
```suggestion
        all_zeros = csr_matrix(self.adata.X.shape, dtype=np.float32)
```
 toarray will create a dense array?

scvi/model/_scvi.py
```diff
@@ -188,4 +203,72 @@ def setup_anndata(
...
+            layers={layer: all_zeros.copy() for layer in self.adata.layers},
+        )
+        bdata.obs = self.adata.obs
+        bdata.obsm = self.adata.obsm
+        bdata.obsp = self.adata.obsp
+        bdata.uns = self.adata.uns
+        bdata.var_names = self.adata.var_names
+        # add latent related metadata
+        if mode == "sampled":
+            bdata.obsm[_SCVI_LATENT_SAMPLES] = self.adata.obsm[use_latent_key]
```
adamgayoso, line 231: I will not comment this everywhere, but I don't see the purpose of having multiple modes, most of the code added assumes parameterized posteriors, so we should always assume dist mode. This will make the code a bit simpler as well.

scvi/model/_scvi.py
```diff
@@ -188,4 +203,72 @@ def setup_anndata(
...
+            raise ValueError(f"Unknown latent mode: {mode}")
+        return latent_fields
+
+    def to_latent_mode(
+        self,
+        mode: LatentDataType,
+        use_latent_key: str = "X_latent",
+        use_latent_qzm_key: str = "X_latent_qzm",
+        use_latent_qzv_key: str = "X_latent_qzv",
+    ):
```
adamgayoso, line 270: we should add the `latent_mode` to setup_anndata, at the moment I'm not seeing how load works properly (though I see it's tested). How do the latent fields get added when setup anndata gets run during the load method?

adamgayoso: another benefit of this is we can always use the manager to check if an anndata was used for latent mode.

scvi/module/_totalvae.py
```diff
@@ -550,6 +550,10 @@ def inference(
             untran_l=untran_l,
         )
 
+    @auto_move_data
+    def inference_no_encode(self, qz_m, qz_v, latent_data_type, n_samples=1):
+        raise NotImplementedError()
+
```
adamgayoso, line 556: Please remove these. This will break a lot of people's code

adamgayoso (changes requested):

scvi/module/base/_base_module.py
```diff
@@ -523,18 +533,31 @@ def _generic_forward(
...
     inference_inputs = module._get_inference_input(
         tensors, **get_inference_input_kwargs
     )
-    inference_outputs = module.inference(**inference_inputs, **inference_kwargs)
+    if is_latent:
+        inference_outputs = module.inference_no_encode(
+            **inference_inputs, **inference_kwargs
+        )
+    else:
```
adamgayoso, line 559: I think it would be best not to include this new function, so as to isolate these changes to just changes within the respective model/module classes.

Inference in the VAE could look like this:

```python
def inference(...):
    if dist_params is not None:
        self._inference_no_encode(...)
    else:
        self._inference_with_encode(...)
```

The way the base module is written already permits this flexibility. Once the anndatamanager is registered with the latent fields it will start adding these tensors into the minibatches, so this is a way for the module to know which one to route to.

scvi/model/_scvi.py
adamgayoso: Can we create another abstract class that augments `BaseModelClass` with some functionality that one would need to add for latent mode?

scvi/utils/_decorators.py
```diff
@@ -0,0 +1,26 @@
+from functools import wraps
+from typing import Callable
+
+from anndata import AnnData
+from mudata import MuData
+
+from scvi.data._utils import _get_latent_adata_type
+
+
+def unsupported_in_latent_mode(fn: Callable) -> Callable:
```
adamgayoso, line 10: if we create an extended version of `BaseModelClass` with some latent mode functionality we can add a property to the model that this decorator can then use to check if it's supported

scvi/model/base/_vaemixin.py
```diff
@@ -14,6 +16,7 @@ class VAEMixin:
     """Univseral VAE methods."""
 
     @torch.no_grad()
+    @unsupported_in_latent_mode
```
adamgayoso, line 19: Because latent mode doesn't necessarily mean the count data has been set to 0, this would technically be supported (especially with the other changes mentioned). However, if the `memory_reduce` option has been used, the counts will be 0, so the ELBO will be nonsensical. Therefore the decorator can instead add a warning that memory reduce has been used so the output of this function might not make sense.

scvi/model/base/_vaemixin.py
```diff
@@ -118,13 +123,15 @@ def get_reconstruction_error(
         return reconstruction_error
 
     @torch.no_grad()
+    @unsupported_in_latent_mode
```
adamgayoso, line 126: Based on my proposed changes, this would be supported in latent mode because inference would work just fine.

adamgayoso (changes requested):

scvi/model/_scvi.py
watiss: I will remove it the next pr (i.e. after this one is merged in)

scvi/model/_scvi.py
```diff
@@ -188,4 +200,8 @@ def setup_anndata(
             fields=anndata_fields, setup_method_args=setup_method_args
         )
         adata_manager.register_fields(adata, **kwargs)
+        # register new fields for latent mode if needed
+        latent_mode = _get_latent_adata_type(adata)
+        if latent_mode is not None:
+            adata_manager.register_new_fields(cls._get_latent_fields(latent_mode))
```
adamgayoso, line 206: can you check if latent mode and then append the latent fields to anndata_fields on line 185? This way validation will be performed on loading

scvi/model/base/_latent_mode_mixin.py
```diff
@@ -0,0 +1,90 @@
...
+            raise ValueError(f"Unknown latent mode: {mode}")
+        latent_fields.append(
+            StringUnsField(
+                REGISTRY_KEYS.LATENT_MODE_KEY,
+                _ADATA_LATENT_UNS_KEY,
+            ),
+        )
+        return latent_fields
+
+    def to_latent_mode(
```
adamgayoso, line 79: can you write detailed documentation of these methods in this class?

scvi/model/base/_base_model.py
```diff
@@ -64,6 +69,8 @@ class BaseModelClass(metaclass=BaseModelMetaClass):
     """Abstract class for scvi-tools models."""
 
     def __init__(self, adata: Optional[AnnOrMuData] = None):
+        latent_adata = adata is not None and _get_latent_adata_type(adata) is not None
+        _raise_if_missing_latent_mode_support(type(self).__name__, latent_adata)
```
adamgayoso, line 73: why is this necessary?

tests/models/test_models.py
```diff
@@ -1606,3 +1607,356 @@ def test_early_stopping():
...
+    # load saved non-latent model with latent adata
+    loaded_model = SCVI.load(save_path, adata=model.adata)
+
+    scvi.settings.seed = 1
+    params_latent = loaded_model.get_likelihood_parameters()
+    assert params_latent["mean"].shape == adata.shape
+    assert np.array_equal(params_latent["mean"], params_orig["mean"])
+
+
+def test_scvi_latent_mode_get_latent_representation():
```
adamgayoso, line 1893: can we put all latent mode tests in their own file?

scvi/_types.py
```diff
@@ -5,7 +5,10 @@
 import mudata
 import torch
 
+from ._compat import Literal
+
 Number = Union[int, float]
 AnnOrMuData = Union[anndata.AnnData, mudata.MuData]
 Tensor = Union[torch.Tensor, jnp.ndarray]
 LossRecord = Union[Dict[str, Tensor], Tensor]
+LatentDataType = Literal["sampled", "dist"]
```
adamgayoso, line 14: I would vote for removing sampled mode asap so we don't introduce a bunch of code we will end up removing anyway

adamgayoso (changes requested):
Really close! It looks really clean now.

scvi/model/base/_latent_mode_mixin.py
```diff
@@ -0,0 +1,81 @@
...
+from scvi.data.fields import ObsmField, StringUnsField
+
+logger = logging.getLogger(__name__)
+
+_SCVI_LATENT_SAMPLES = "_scvi_latent_samples"
+_SCVI_LATENT_QZM = "_scvi_latent_qzm"
+_SCVI_LATENT_QZV = "_scvi_latent_qzv"
+
+
+class LatentModeMixin:
```
adamgayoso, line 18: abstract class, move this code to scvi

scvi/dataloaders/_anntorchdataset.py
```diff
@@ -115,6 +116,9 @@ def __getitem__(self, idx: List[int]) -> Dict[str, np.ndarray]:
                 sliced_data = cur_data.iloc[idx, :].to_numpy().astype(dtype)
             elif issparse(cur_data):
                 sliced_data = cur_data[idx].toarray().astype(dtype)
+            # for latent mode anndata
```
adamgayoso, line 119: TODO: Adata manager should have a list of which fields it will load

adamgayoso: but for now I would put a better comment so we don't forget why we did it

scvi/data/fields/_uns_field.py
```diff
@@ -0,0 +1,92 @@
...
+from anndata import AnnData
+
+from scvi.data import _constants
+
+from ._base_field import BaseAnnDataField
+
+logger = logging.getLogger(__name__)
+
+
+class BaseUnsField(BaseAnnDataField):
```
adamgayoso, line 14: let's make sure we add these to documentation indexes

scvi/dataloaders/_anntorchdataset.py
adamgayoso: and the todo!

scvi/model/_scvi.py
adamgayoso: let's purse this change

scvi/model/base/_base_model.py
adamgayoso: let's think if this works for developers in other codebases

scvi/module/_vae.py
```diff
@@ -310,6 +334,17 @@ def inference(self, x, batch_index, cont_covs=None, cat_covs=None, n_samples=1):
         outputs = dict(z=z, qz=qz, ql=ql, library=library)
         return outputs
 
+    @auto_move_data
+    def _inference_no_encode(self, qzm, qzv, n_samples=1):
```
adamgayoso, line 338: _cached_inference

scvi/module/base/_latent_mode_module_mixin.py
```diff
@@ -0,0 +1,26 @@
+import logging
+from typing import Optional
+
+from scvi._types import LatentDataType
+from scvi.module.base import auto_move_data
+
+logger = logging.getLogger(__name__)
+
+
+class LatentModeModuleMixin:
```
adamgayoso, line 10: make this one abstract? Or keep as a mixin?

adamgayoso: abstract so we force that they implement cached_inference

adamgayoso (changes requested):

scvi/model/base/_base_model.py
watiss: good call

scvi/model/base/__init__.py
```diff
@@ -1,7 +1,9 @@
 from ._archesmixin import ArchesMixin
-from ._base_model import BaseModelClass
+from ._base_model import BaseLatentModeModelClass, BaseModelClass
 from ._differential import DifferentialComputation
 from ._jaxmixin import JaxTrainingMixin
+
+# from ._base_latent_mode_model import BaseLatentModeModelClass
```
adamgayoso, line 6: nit: remove commented code

scvi/model/base/_base_model.py
```diff
@@ -64,6 +64,11 @@ class BaseModelClass(metaclass=BaseModelMetaClass):
     """Abstract class for scvi-tools models."""
 
     def __init__(self, adata: Optional[AnnOrMuData] = None):
+        latent_adata = adata is not None and _get_latent_adata_type(adata) is not None
```
adamgayoso, line 67: nit: can we add a comment here so it's clear what this does?

scvi/model/_scvi.py
```diff
@@ -184,8 +206,62 @@ def setup_anndata(
...
+            ),
+        )
+        return latent_fields
+
+    def to_latent_mode(
+        self,
+        mode: LatentDataType,
+        use_latent_qzm_key: str = "X_latent_qzm",
+        use_latent_qzv_key: str = "X_latent_qzv",
+    ):
```
adamgayoso, line 264: can we document this? It should mention that the model.adata will be modified inplace and user should make a copy if they want to preserve it.

scvi/module/_vae.py
```diff
@@ -310,6 +329,17 @@ def inference(self, x, batch_index, cont_covs=None, cat_covs=None, n_samples=1):
         outputs = dict(z=z, qz=qz, ql=ql, library=library)
         return outputs
 
+    @auto_move_data
+    def _cached_inference(self, qzm, qzv, n_samples=1):
+        if self.latent_data_type == "dist":
+            dist = Normal(qzm, qzv.sqrt())
+            untran_z = dist.rsample() if n_samples == 1 else dist.sample((n_samples,))
```
adamgayoso, line 336: this should be sample and not rsample

adamgayoso (changes requested):
Just a few more little things!

adamgayoso (approved):

PR #1673, opened 2022-09-13
Switch to sphinx book theme

adamgayoso:
This aligns us with the scverse cookiecutter. The theme also now has dark mode.

One thing we need to change is to make the tutorial notebooks use myst-nb style admonitions.

The way the theme handles colab urls isn't compatible with how we use git submodules, so I created a simple sphinx extension to handle this and edit the colab url post hoc


adamgayoso:
@martinkim0 we can get this in but not backport it to 0.20.x, so it would come out at the next major release

PR #1683, opened 2022-09-15
Added scAR as an external model for ambient RNA removal

ricomnl:
I'm proposing to add scAR as an external model. It's based on this [publication](https://www.biorxiv.org/content/10.1101/2022.01.14.476312v4) and this is the original [Github](https://github.com/Novartis/scar). I used a slightly modified version of SCVI's VAE and, other than that, only adjusted the `generative()` function.


.github/workflows/test.yml
```diff
@@ -5,7 +5,7 @@ name: scvi-tools
 
 on:
   push:
-    branches: [master, 0.15.x, 0.16.x, 0.17.x]
+    branches: [master, scar, 0.15.x, 0.16.x, 0.17.x]
```
adamgayoso, line 8: no need for this

tests/external/test_scar.py
```diff
@@ -0,0 +1,19 @@
...
+    adata = synthetic_iid()
+    SCAR.setup_anndata(adata, batch_key="batch", labels_key="labels")
+
+    model = SCAR(adata, ambient_profile=None, n_latent=n_latent)
+    model.train(1, check_val_every_n_epoch=1, train_size=0.5)
+    model.get_elbo()
+    model.get_latent_representation()
+    model.get_marginal_ll(n_mc_samples=5)
+    model.get_reconstruction_error()
+    model.history
```
adamgayoso, line 16: how do you get the denoised counts out of the model?

scvi/external/scar/_model.py
```diff
@@ -0,0 +1,435 @@
...
+            ),
+        ]
+        adata_manager = AnnDataManager(
+            fields=anndata_fields, setup_method_args=setup_method_args
+        )
+        adata_manager.register_fields(adata, **kwargs)
+        cls.register_manager(adata_manager)
+
+    @classmethod
+    def get_ambient_profile(
```
adamgayoso, line 230: how is this function used in the workflow? It would be good to put the workflow in the user guide

scvi/external/scar/_model.py
```diff
@@ -0,0 +1,435 @@
...
+        # update ambient profile
+        for ft in feature_type:
+            tmp = emptydrops[:, emptydrops.var["feature_types"] == ft]
+            adata.uns[f"ambient_profile_{ft}"] = pd.DataFrame(
+                tmp.X.sum(axis=0).reshape(-1, 1) / tmp.X.sum(),
+                index=tmp.var_names,
+                columns=[f"ambient_profile_{ft}"],
+            )
+
+        if kneeplot:
```
adamgayoso, line 378: we make it a point to not include any plotting code in this package, so I would prefer to remove this (as well as in the docstring)

adamgayoso (changes requested):
Thanks for the contribution. I added some initial comments

scvi/external/scar/_model.py
adamgayoso: can this function be a static method? User can put the ambient profile in `.varm`, and then the `ambient_profile_varm_key` can go in `setup_anndata`.

.github/workflows/test.yml
ricomnl: just been using it because I was working on the `scar` instead of `master` branch in my repo. will remove it in time for the merge

scvi/external/scar/_model.py
ricomnl: I couldn't find any varm keys in the `scvi.data.fields`. You think its worth adding a varm class?

scvi/external/scar/_model.py
ricomnl: sounds good, I removed it in the latest commit

tests/external/test_scar.py
ricomnl: added as a test in latest commit with an example in the user guide

scvi/external/scar/_model.py
```diff
@@ -0,0 +1,310 @@
...
+    @setup_anndata_dsp.dedent
+    def setup_anndata(
+        cls,
+        adata: AnnData,
+        layer: Optional[str] = None,
+        batch_key: Optional[str] = None,
+        labels_key: Optional[str] = None,
+        size_factor_key: Optional[str] = None,
+        categorical_covariate_keys: Optional[List[str]] = None,
+        continuous_covariate_keys: Optional[List[str]] = None,
```
adamgayoso, line 193: I haven't read scAR too in depth -- does it make sense to allow this functionality? My impression is that it's run once per 10x lane?

scvi/external/scar/_model.py
adamgayoso: hmm I see I think you solution is adequate for this model

scvi/external/scar/_model.py
adamgayoso: Therefore, no batch correction, etc is necessary. If this is the case, can you remove this as well as all related functionality at the PyTorch module level?

scvi/external/scar/_module.py
```diff
@@ -0,0 +1,460 @@
...
+        library_size
+            library size
+        cat_list
+            list of category membership(s) for this sample
+
+        Returns
+        -------
+        4-tuple of :py:class:`torch.Tensor`
+            parameters for the ZINB distribution of expression
+
```
adamgayoso, line 174: Please remove. Or make correct

scvi/external/scar/_module.py
```diff
@@ -0,0 +1,460 @@
...
+        # dropout
+        self.px_dropout_decoder = nn.Linear(n_hidden, n_output)
+
+    def forward(
+        self,
+        dispersion: str,
+        z: torch.Tensor,
+        library: torch.Tensor,
+        *cat_list: int,
```
adamgayoso, line 145: related to an earlier comment, do we need this functionality?

scvi/external/scar/_module.py
```diff
@@ -0,0 +1,460 @@
...
+        px_noise_ratio = self.px_noise_decoder(px)
+
+        # Clamp to high value: exp(12) ~ 160000 to avoid nans (computational stability)
+        px_rate = torch.exp(library) * px_scale  # torch.clamp( , max=12)
+        px_r = self.px_r_decoder(px) if dispersion == "gene-cell" else None
+
+        return px_scale, px_noise_ratio, px_r, px_rate, px_dropout
+
+
+class SCAR_VAE(VAE):
```
adamgayoso, line 191: Are the architectures exactly matched? If not, can you comment on this in the docstring of the model as well as in the user guide? At least this way people know it's not exactly the same but the outputs were checked to be similar on some examples.

scvi/external/scar/_module.py
```diff
@@ -0,0 +1,460 @@
...
+            y,
+        )
+        if self.dispersion == "gene-label":
+            px_r = F.linear(
+                one_hot(y, self.n_labels), self.px_r
+            )  # px_r gets transposed - last dimension is nb genes
+        elif self.dispersion == "gene-batch":
+            px_r = F.linear(one_hot(batch_index, self.n_batch), self.px_r)
+        elif self.dispersion == "gene":
+            px_r = self.px_r
```
adamgayoso, line 369: Related to earlier comments -- I especially think we don't need this sort of functionality. So I would try to remove.

scvi/external/scar/_module.py
```diff
@@ -0,0 +1,460 @@
...
+        # remove estimated noise
+        px_scale = px_scale * (1 - px_noise_ratio)
+        px_rate = px_rate * (1 - px_noise_ratio)
+        pamb_scale = self.ambient_profile.to(px_scale.device) * px_noise_ratio
+        pamb_rate = pamb_scale * torch.exp(size_factor)
+        px_r = torch.exp(px_r)
+
+        if self.gene_likelihood == "zinb":
+            px = ZeroInflatedNegativeBinomial(
+                mu=px_rate,
```
adamgayoso, line 380: I get it, posterior_predictive_sample works because of what you're doing here, pretty clever. I would add a comment somewhere about this because a priori it wouldn't make sense that `posterior_predictive_sample` removes background.

scvi/external/scar/_model.py
```diff
@@ -0,0 +1,310 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SCAR(
+    RNASeqMixin, ArchesMixin, VAEMixin, UnsupervisedTrainingMixin, BaseModelClass
+):
+    """
+    Ambient RNA removal in scRNA-seq data [Sheng22]_.
+    Original Github: https://github.com/Novartis/scar.
```
adamgayoso, line 41: If the model's are not parameter matched exactly I would comment that here.

adamgayoso (changes requested):

scvi/external/scar/_module.py
ricomnl: I think I'll actually change this functionality and create a separate function to get the denoised counts. This function will have two options for retrieving denoised counts:

(background: scAR models both an estimated native and ambient count distribution)

1. Use the estimated native counts directly (as it is done in scAR)
2. Use the estimated ambient counts and remove them from the original counts (more conservative, this is how [cellbender is doing it](https://github.com/broadinstitute/CellBender/blob/cb2d209d5aedffe71e28947bc5b7859600aef64d/cellbender/remove_background/infer.py#L531))

scvi/external/scar/_model.py
CaibinSh: > I haven't read scAR too in depth -- does it make sense to allow this functionality? My impression is that it's run once per 10x lane?

Hi @adamgayoso , I am the developer of scAR. It's run once per experiment/batch -- the same experiment and batch should have the same ambient profile (a.k.a same composition of contamination), it is lane-independent.

scvi/external/scar/_model.py
CaibinSh: Agree, please comment it. We tested and optimised a unique set of parameters based on and for multiple data modalities.

scvi/external/scar/_model.py
adamgayoso: Thanks @CaibinSh, so it's fair to say that there is no notion of batch correction with this model so we should ten remove all these fields, except perhaps the size factor.

scvi/external/scar/_model.py
adamgayoso: On the other hand, I'm curious how much more code is required to parameter match them exactly.

ricomnl: By parameters, you guys mean both hyperparameters and model architecture parameters (latent dimension, hidden layers) right?

adamgayoso: yes, and then secondarily activations etc.

scvi/external/scar/_model.py
ricomnl: I removed all of the keys and fields related to batch, label and covariates in the latest commit. I did leave them in the `setup_anndata()` function though in order to use the `generic_forward()` and other useful scVI utility functions that would break otherwise. That should be fine right? Alternatively we could rewrite these helper functions in the model to not rely on batch and label keys

scvi/external/scar/_module.py
ricomnl: Corrected in latest commit

scvi/external/scar/_module.py
ricomnl: No, removed in latest commit

scvi/external/scar/_module.py
ricomnl: Removed in latest commit

scvi/external/scar/_module.py
adamgayoso: @ricomnl offered to implement it as well (and using jax perhaps so much faster), but of course this kind of contribution takes a significant amount of work so I can't promise any timeline.

ricomnl: @cnk113 agreed! I guess the main difference between scAR and Cellbender is that scAR uses heuristics to define an ambient distribution and then learns to recreate it, while cellbender learns it de-novo from data. Cellbender also includes barcode swapping so the architecture is a bit more involved which is why I'd rather implement it separately. 
That being said, it would be quite simple to add the approach I mentioned above in case folks prefer a more conservative removal method in scAR

scvi/external/scar/_model.py
ricomnl: Checklist of parameters matched:
- [x] nn_layer1=150
- [ ] nn_layer2=100 (set to 150 as well bc as far as I can tell scVI only allows uniform sizes for hidden layers)
- [x] latent_dim=15
- [x] dropout_prob=0
- [x] sparsity=0.9
- [ ] count_model="binomial" (I think we should go for the standard `zinb` in scVI for this one)
- [x] batch_norm (used in both encoder and decoder with `momentum=0.01, eps=0.001`
- [x] activation functions

Training params:
I'd propose sticking with the scVI training defaults here (using heuristics to determine number of epochs and kl warmup)
- [ ] kld_weight=1e-5
- [ ] epochs=800

ricomnl:
Some tests I ran to make sure everything works. In this latest iteration I created a replica of scAR's `inference()` function to estimate denoised counts. Seems like the `remove_ambient_counts` does not work as expected. Perhaps we should remove it and wait for the cellbender implementation. @adamgayoso thoughts?

## Simulated data
Used the data generator from the [original repository](https://github.com/Novartis/scar/blob/main/scar/main/_data_generater.py#L13) to generate native counts with a known underlying ambient distribution. 
Trained a model and estimated the original counts with both flavors. Then calculated the mean wasserstein difference between the denoised counts and the known native counts to make sure they get closer.
```python
SCAR.setup_anndata(adata)
model = SCAR(adata, ambient_profile=None)
model.train(early_stopping=True)

# flavor 1
adata.layers["denoised_signals"] = model.get_denoised_counts(
    adata, 
    n_samples=1, 
    flavor='sample_denoised_counts'
)

# flavor 2
adata.layers["denoised_signals"] = model.get_denoised_counts(
    adata, 
    n_samples=1, 
    flavor='remove_ambient_counts'
)
```

### Results

#### 1. flavor='sample_denoised_counts'
Original wasserstein distance: 2.534
Denoised wasserstein distance: 2.053

download-88



#### 2. flavor='remove_ambient_counts'
Original wasserstein distance: 2.534
Denoised wasserstein distance: 2.034

download-87


## HGMM20k data from [tutorial](https://scar-tutorials.readthedocs.io/en/latest/tutorials/scAR_tutorial_mRNA_denoising.html)

```python
# used same pre-processing steps to get hgmm_20k and hgmm_20k_raw (see tutorial linked above)

SCAR.get_ambient_profile(
    adata=hgmm_20k,
    raw_adata=hgmm_20k_raw,
    prob=0.995,
)

# architecture parameter defaults are matched
model = SCAR(
    adata=hgmm_20k,
    ambient_profile="ambient_profile",
    sparsity=1
)

# flavor 1
hgmm_20k.layers["denoised_counts"] = model.get_denoised_counts(
    hgmm_20k, 
    flavor='sample_denoised_counts'
)

# flavor 2
hgmm_20k.layers["denoised_counts"] = model.get_denoised_counts(
    hgmm_20k, 
    flavor='remove_ambient_counts'
)
```

### Results

#### Visualization of raw counts
download-83

#### Original `scar` model retrained on the same data
download-85

#### 1. Our ported model with flavor='sample_denoised_counts'
download-90


#### 2. Our ported model with flavor='remove_ambient_counts'
download-89

scvi/external/scar/_model.py
adamgayoso: >  nn_layer2=100 (set to 150 as well bc as far as I can tell scVI only allows uniform sizes for hidden layers)

This is a fine compromise to me

>  count_model="binomial" (using nb - negative binomial from scVI)

This I think is an important distinction, can we make binomial default? Should just be using the torch binomial distribution?

> Training params:
I'd propose sticking with the scVI training defaults here (using heuristics to determine number of epochs and kl warmup)

This also seems fine to me

ricomnl: In the latest commit I changed it to a binomial distribution (using torch's binomial) and it pretty much exactly recapitulates the original results (at least qualitatively -  see comparison below)

download-91

scvi/external/scar/_module.py
ricomnl: I removed the cellbender flavor as the results seem a bit off (see comparison below) and as this is untested with this model I think I'd rather add this functionality with the cellbender model itself

scvi/external/scar/_model.py
CaibinSh: Great to see this @ricomnl. ZINB is more flexible than Binomial and Poisson distributions. It requires one more parameter to learn. We also found that Binomial distribution produces better results.

CaibinSh (approved):

scvi/external/scar/_model.py
CaibinSh: Great to see this @ricomnl .

I approve these changes since you seem to recapitulate our results. I have not review all codes though.

Agreed. The default kld_weight in scVI seems to be 1.0. This should produce very good results for mRNA denoising with scAR, but not the optimal results for sgRNA assignment in scCRISPR-seq (which is not the case here). 
The epoch number with scVI's default setting should also produce good results for scAR. Generally, about 150 are enough in our test. 

> Training params:
I'd propose sticking with the scVI training defaults here (using heuristics to determine number of epochs and kl warmup)
  kld_weight=1e-5
 epochs=800

scvi/external/scar/_model.py
```diff
@@ -0,0 +1,334 @@
...
+        sample = int(sample)
+        idx = np.random.choice(
+            raw_adata.shape[0], size=min(raw_adata.shape[0], sample), replace=False
+        )
+        raw_adata = raw_adata[idx]
+        print(f"Randomly sampling {sample} droplets to calculate the ambient profile.")
+        # initial estimation of ambient profile, will be updated
+        ambient_prof = raw_adata.X.sum(axis=0) / raw_adata.X.sum()
+
+        for _ in tqdm(range(iterations)):
```
adamgayoso, line 237: scvi.utils.track is a wrapper on tqdm and rich I'm trying to use codebase wide

.github/workflows/test.yml
adamgayoso, line 8: don't forget to remove!

scvi/external/scar/_model.py
```diff
@@ -0,0 +1,334 @@
...
+        self.init_params_ = self._get_init_params(locals())
+
+    @classmethod
+    @setup_anndata_dsp.dedent
+    def setup_anndata(
+        cls,
+        adata: AnnData,
+        layer: Optional[str] = None,
+        batch_key: Optional[str] = None,
+        labels_key: Optional[str] = None,
```
adamgayoso, line 162: can't these be removed from here but kept internally in the function with None arguments?

adamgayoso (changes requested):
Just a few more little things and then we can merge this!

.github/workflows/test.yml
ricomnl: done in last commit

scvi/external/scar/_model.py
ricomnl: done in latest commit

scvi/external/scar/_model.py
ricomnl: done in latest commit

adamgayoso (approved):

PR #1689, opened 2022-09-16
Add register_new_fields method to manager

adamgayoso:
This allows updating an instance of an AnnDataManager with new fields (assuming it has already been registered with an AnnData object).

We will use this to help put models in latent mode, in which latent dists are cached in anndata and used in forward passes instead of encoding.

### DIFF
--- a/scvi/data/_manager.py
+++ b/scvi/data/_manager.py
@@ -278,8 +278,7 @@ class AnnDataManager:
             of the setup method that will be used to update the existing values
             in the registry of this instance.
         """
-        if setup_method_args is not None:
-            self._registry[_constants._SETUP_ARGS_KEY].update(setup_method_args)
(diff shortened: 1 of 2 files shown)


scvi/data/_manager.py
```diff
@@ -224,6 +267,20 @@ def validate(self) -> None:
...
+        Update setup method args.
+
+        Parameters
+        ----------
+        setup_method_args
+            This is a bit of a misnomer, this is a dict representing kwargs
+            of the setup method that will be used to update the existing values
+            in the registry of this instance.
+        """
+        if setup_method_args is not None:
```
watiss, line 281: nit: do we ever expect to call this with a None?

watiss (approved):

scvi/data/_manager.py
adamgayoso: No whoops, should've removed this

watiss (approved):

Issue #13, opened 2018-06-12
[CLOSED] Importance Weighted Autoencoder


jeff-regier:
<a href="https://github.com/jeff-regier"><img src="https://avatars2.githubusercontent.com/u/8953355?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jeff-regier](https://github.com/jeff-regier)**
_Friday Mar 30, 2018 at 17:00 GMT_
_Originally opened as https://github.com/YosefLab/scVI-dev/issues/8_

----

https://arxiv.org/pdf/1509.00519.pdf

jeff-regier:
<a href="https://github.com/jeff-regier"><img src="https://avatars2.githubusercontent.com/u/8953355?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [jeff-regier](https://github.com/jeff-regier)**
_Thursday Apr 12, 2018 at 20:58 GMT_

----

This code I wrote for another project may be useful for this issue:
```
    def iwae_loss(self, ys, k=10):
        tall_shape = [ys.size(0) * k] + list(ys.size()[1:])
        ys_tall = ys.unsqueeze(1).repeat(1, k, 1, 1).view(tall_shape)

        wide_shape = [ys.size(0), k]
        vae_losses = self.vae_losses(ys_tall).view(wide_shape)

        iwae_gains = logsumexp(-1 * vae_losses, dim=1) - np.log(k)
        return -1 * iwae_gains.sum()
```
Above `ys` are the data/observations and `self.vae_losses` is a method that returns a vector of vae losses, one for each observation in the minibatch. (VAE loss approximates `-1 * E_q p(x, z, l) / q(z, l)` ).

For some reason PyTorch doesn't have an logsumexp function yet, so I use
```
def logsumexp(inputs, dim=None, keepdim=False):
    return (inputs - f.log_softmax(inputs, dim=dim)).mean(dim, keepdim=keepdim)
```

PR #1696, opened 2022-09-19
Replace custom `attrdict` with `ml_collections` implementation.

### LINKED ISSUE #1690: Sub custom attrdict for ml collections
https://github.com/google/ml_collections

martinkim0:
Closes #1690.


martinkim0:
Seems that `ml_collections` does not allow `ConfigDict`s to be initialized with nested dictionaries where inner dictionaries have non-iterable types (e.g. `int`s) as keys ([#13](https://github.com/google/ml_collections/issues/13)). This comes up as a problem when initializing `~scvi.model.TOTALVI`, since `"protein_batch_mask"` has `int`s as keys.

docs/api/developer.md
```diff
@@ -246,5 +246,4 @@ Utility functions used by scvi-tools.
 
    utils.track
    utils.setup_anndata_dsp
-   utils.attrdict
```
adamgayoso, line 249: Removing the old one completely would be a breaking change. Can we just create a dummy attrdict object that inherits completely an ml collection?

adamgayoso (changes requested):

scvi/utils/_attrdict.py
```diff
@@ -1,40 +1,11 @@
...
-    Parameters
-    ----------
-    recursive
-        If True, recursively converts nested dictionaries into :class:`~scvi.utils.attrdict` objects.
-
-    Notes
-    -----
-    Based off of https://stackoverflow.com/questions/38034377/object-like-attribute-access-for-nested-dictionary.
+    Dummy class to allow for backwards compatibility with old attrdict
+    Now uses ml-collections FrozenConfigDict.
```
adamgayoso, line 7: we should be able to add ml colletions to intersphinx in conf.py and do a proper :class: link to the object.

https://ml-collections.readthedocs.io/en/latest/

docs/api/developer.md
martinkim0: Addressed with new changes

adamgayoso (changes requested):
LGTM but can you add the package to pyproject?

PR #1706, opened 2022-09-21
Make DataSplitter more transparent to the existing Lightning Data Loader API wrt kwargs

mkarikom:
This allows DataSplitter and SemiSupervisedDataSplitter to pass kwargs to the DataLoader constructor in keeping with the DataSplitter API which states "**kwargs: Keyword args for data loader".  

The original implementation didn't completely observe the above design, since passing `drop_last` as a `kwarg` to the Splitter caused an error due to the the hard-coded default of `drop_last=3` in `data_loader_kwargs`.

In addition to allowing default values for all `data_loader_kwargs` to be defined under `defaultKwargs`, the main benefit of this refactoring is that DataSplitter is now transparent wrt the existing Lightning Data Loader API (like `drop_last=True`).

Because this refactoring does not affect supervised training, new tests added to `tests/models/test_models.py::test_data_splitter` run directly on `DataSplitter`, but not on `SemiSupervisedDataSplitter` (the implementation is identical)


mkarikom:
@adamgayoso I'm happy to add more unit tests if required, just lmk and thanks for looking!

scvi/dataloaders/_data_splitting.py
```diff
@@ -94,10 +94,12 @@ def __init__(
         **kwargs,
     ):
         super().__init__()
+
+        defaultKwargs = {"drop_last": 3}
```
adamgayoso, line 98:
```suggestion
        default_data_loader_kwargs = {"drop_last": 3}
```

scvi/dataloaders/_data_splitting.py
```diff
@@ -94,10 +94,12 @@ def __init__(
...
         self.adata_manager = adata_manager
         self.train_size = float(train_size)
         self.validation_size = validation_size
-        self.data_loader_kwargs = kwargs
+        self.data_loader_kwargs = {**defaultKwargs, **kwargs}
```
adamgayoso, line 102:
```suggestion
        self.data_loader_kwargs = default_data_loader_kwargs.update(kwargs)
```

tests/models/test_models.py
```diff
@@ -922,6 +922,50 @@ def test_data_splitter():
         ds.train_dataloader()
         ds.val_dataloader()
 
+    # test that data_loader_kwargs get default drop_last=3 when respective keyword arg is not provided to DataSplitter
+    ds = DataSplitter(adata_manager, train_size=400 / a.n_obs, batch_size=397)
```
adamgayoso, line 926: can all of these tests be moved into the data testing code? and each test should be its own function

scvi/dataloaders/_data_splitting.py
adamgayoso: I'd also move this line next to the other line that updates it

adamgayoso (changes requested):
Thanks, I added some comments!

tests/models/test_models.py
mkarikom: yes, now implemented but pls see comment below

mkarikom:
Hi @adamgayoso ,
I'm not sure what is going on with pytest throwing tons of errors.
As a sanity check I've created a vanilla cuda dev container which replicates all pytest errors in my environment in the following steps:

1. clone https://github.com/mkarikom/cuda_devcon
2. open the `cuda_devcon` folder in vscode
3. confirm "open folder in container"
4. wait for build
5. `python -m pip install pytest && git clone --recurse-submodules git@github.com:scverse/scvi-tools.git && cd scvi-tools && pytest tests/models/test_models.py`

mkarikom:
I hit a run-time bug with the proposed `default_data_loader_kwargs.update(kwargs)` syntax so reverted to the previous splatting syntax

mkarikom:
Hey, @adamgayoso!  
I think this is now current and all checks passing!

Sorry about the delayed changes!

A style issues in the pre-commit.ci that I can't see at all with my flake8/pylint plugins.
The only other style issue was "leading _" but adding those led to undefined errors in the tests so I've ignored it.

Hope your last few weeks have been as fun as ours ;)

PR #1747, opened 2022-10-17
Allow custom optimizers, improve `JaxTrainingPlan`

adamgayoso:
- Easily add optax/pytorch custom optimizers easily
- Breaking change where we force keyword args to the training plans


tests/models/test_models_latent_mode.py
```diff
@@ -79,7 +81,11 @@ def run_test_scvi_latent_mode_dist(
         assert params_latent[k].shape == adata.shape
 
     for k in keys:
-        assert np.array_equal(params_latent[k], params_orig[k])
+        # Allclose because on GPU, the values are not exactly the same
+        # as latents are moved to cpu in latent mode
+        np.testing.assert_allclose(
+            params_latent[k], params_orig[k], rtol=3e-1, atol=5e-1
+        )
```
adamgayoso, line 88: cc @watiss needed to change this. LMK if the comment makes sense!

martinkim0 (approved):
LGTM

PR #1756, opened 2022-10-20
Latent mode refactoring

adamgayoso:
This performs some cleanup:

1. User anndata no longer edited in place


scvi/model/_scvi.py
```diff
@@ -280,7 +280,18 @@ def to_latent_mode(
...
+        bdata = self._get_reduced_adata(mode)
+        if mode == _SCVI_LATENT_MODE:
+            bdata.obsm[_SCVI_LATENT_QZM] = self.adata.obsm[use_latent_qzm_key]
+            bdata.obsm[_SCVI_LATENT_QZV] = self.adata.obsm[use_latent_qzv_key]
+        else:
+            raise ValueError(f"Unknown latent mode: {mode}")
+        new_manager = self.adata_manager.transfer_fields(bdata)
+        new_manager.register_new_fields(self._get_latent_fields(mode))
+        self._adata_manager = new_manager
+        self.adata = bdata
```
watiss, line 296: I am a bit confused. Aren't we overwriting the user's data by doing this, and if so then what is the point in making any of these changes? More precisely:

- If `id(self.adata) == id(user's adata)` then this is essentially overwriting the user's data _and_ the model's adata, so there is no point in making a copy since we are losing the original dataset regardless.

- If` id(self.adata) != id(user's adata) `then the user's data is anyways not altered with this call the way that it is today in the main branch, so again there is no point in making a copy.

🤔

scvi/model/_scvi.py
```diff
@@ -258,9 +257,10 @@ def _get_latent_fields(mode: LatentDataType):
 
     def to_latent_mode(
```
watiss, line 258: Here is what I might do: Keep the contents of this function as it is today (per the main branch). Add an option in this call to return a copy of the adata. If the user asks for that option, then return make and return a copy of the adata otherwise leave everything as is (i.e. override the data). I believe this would address my comment below as well (the comment regarding whether the copy is making any difference the way it is currently implemented).

scvi/module/base/_base_module.py
```diff
@@ -348,6 +348,8 @@ class BaseLatentModeModuleClass(BaseModuleClass):
     @property
     def latent_data_type(self) -> Optional[LatentDataType]:
         """The latent data type associated with this module."""
+        if not hasattr(self, "_latent_data_type"):
```
watiss, line 351: Is it common practice to set a property in a getter method? How about just returning `None` if `not hasattr(self, "_latent_data_type")`?

scvi/model/_scvi.py
```diff
@@ -30,7 +30,7 @@
 
 _SCVI_LATENT_QZM = "_scvi_latent_qzm"
 _SCVI_LATENT_QZV = "_scvi_latent_qzv"
-
+_SCVI_LATENT_MODE = "posterior_parameters"
```
watiss, line 33: nit-ish: We might choose to have other latent mode types for scvi, so I would name this something more to the effect of "posterior_parameters" like `POSTERIOR_PARAMS_LATENT_MODE` or something rather than to the fact that this is scvi's latent mode because in theory scvi might have different types of latent mode if we want it. What do you think?

watiss (changes requested):

scvi/model/_scvi.py
```diff
@@ -280,7 +280,18 @@ def to_latent_mode(
             Key to use in `adata.obsm` where the latent qzm params are stored
         use_latent_qzv_key
             Key to use in `adata.obsm` where the latent qzv params are stored
+        attach_reduced_data
+            If `True`, sets `.adata` attribute with a memory-reduced anndata
+            in which all count data is set to zeros
         """
-        self._get_latent_adata_from_adata(mode, use_latent_qzm_key, use_latent_qzv_key)
-        self.adata_manager.register_new_fields(self.__class__._get_latent_fields(mode))
+        bdata = self._get_reduced_adata(mode)
```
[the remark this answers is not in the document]
adamgayoso: to be clear, there is no copying here. We are passing references to .obs, .uns etc to a new anndata, and the overhead of creating the sparse matrices is very minimal (there's no data other than a shape basically)

scvi/model/_scvi.py
adamgayoso: bdata is always a new anndata object in this code (with references to adata's objects). 

What's happening here is we have a new bdata object, we validate it by creating a new anndata manager (via transfer).

Then we add the new latent fields to the new manager, and finally swap out the manager and adata associated with the model (so the model is no longer associated with adata)

scvi/module/base/_base_module.py
adamgayoso: Done.

scvi/model/_scvi.py
adamgayoso: I left this as a TODO

scvi/model/_scvi.py
```diff
@@ -290,7 +292,28 @@ def to_latent_mode(
...
+        this method with `minify_adata=True`. This adata does not contain any of
+        the original count data, but instead contains the latent representation
+        of the original data and metadata.
         """
-        self._get_latent_adata_from_adata(mode, use_latent_qzm_key, use_latent_qzv_key)
-        self.adata_manager.register_new_fields(self.__class__._get_latent_fields(mode))
+        if minify_adata:
+            bdata = self._get_reduced_adata(mode)
+        else:
+            bdata = self.adata
```
watiss, line 309: Why should we allow a latent mode where the data is not dehydrated?

scvi/model/_scvi.py
watiss: I think. my question still remains. In particular, I dont understand why we are creating a new `adata` object (`bdata`), only to have it mirror `self.adata` 1:1.

watiss (changes requested):

scvi/model/_scvi.py
adamgayoso: for cached inference without altering data

scvi/_types.py
```diff
@@ -5,8 +5,10 @@
...
 import torch
 
+from scvi._compat import Literal
+
 Number = Union[int, float]
 AnnOrMuData = Union[anndata.AnnData, mudata.MuData]
 Tensor = Union[torch.Tensor, jnp.ndarray]
 LossRecord = Union[Dict[str, Tensor], Tensor]
-LatentDataType = str
+LatentDataType = Union[Literal["posterior_parameters"], None]
```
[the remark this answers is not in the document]
adamgayoso: I'm not following this comment, prior to this PR it can be None, but the type didn't specify that

scvi/model/_scvi.py
adamgayoso: There are two options:

1. `minify_adata` is False and nothing happens w.r.t. user data expect the new fields
2. `minify_adata` is True and a truly minimal adata is attached to the model. No data is copied as it's all references to adata objects in memory.

scvi/_types.py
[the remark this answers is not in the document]
adamgayoso: Are you suggesting to change to `Optional[Literal["posterior_parameters"]]`? Or use optional in the code?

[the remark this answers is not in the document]
adamgayoso: Ok I think I got it with these changes

scvi/model/_scvi.py
watiss: I dont think we should mix the two notions under the umbrella of latent data.
The codebase (and developer interface) currently assumes that latent == no count data. If you want to shove both states under the same notion (which is feasible but would probably not something I would do) then at a minimum the LatentDataType or other data "tag" needs to capture that. Unless you would want call sites to check if the data is "real latent" or "half latent".
What I'd do for this is I'd probably use LatentDataType to include whether the full data is present as well or not.

scvi/model/_scvi.py
watiss: I get that. But my question is why create a new anndata object, call it bdata, (optionally "minify" it,) and replace adata with bdata. Why not just do everything directly on self.adata?

watiss (changes requested):

scvi/model/_scvi.py
adamgayoso: good point, should be fixed now

scvi/model/_scvi.py
```diff
@@ -226,26 +227,31 @@ def setup_anndata(
...
+            X=all_zeros,
+            layers=layers,
+            uns=self.adata.uns,
+            obs=self.adata.obs,
+            var=self.adata.var,
+            varm=self.adata.varm,
+            obsm=self.adata.obsm,
+            obsp=self.adata.obsp,
+        )
+        del bdata.uns[_SCVI_UUID_KEY]
```
watiss, line 247: Add a code comment to explain why we do this?

watiss (approved):
Just pending the one comment on adding a code comment

scvi/model/_scvi.py
adamgayoso, line 247:
```suggestion
        # Remove scvi uuid key to make bdata fresh w.r.t. the model's manager
        del bdata.uns[_SCVI_UUID_KEY]
```

PR #1758, opened 2022-10-20
Fix Kl weight

adamgayoso:
Kl weight is not working for jax models due to new inheritance of the training plans and that jax modules are dataclasses

### DIFF
--- a/scvi/train/_trainingplans.py
+++ b/scvi/train/_trainingplans.py
@@ -187,7 +187,7 @@ class TrainingPlan(pl.LightningModule):
         self._n_obs_validation = None
 
         # automatic handling of kl weight
-        self._loss_args = signature(self.module.loss).parameters
+        self._loss_args = signature(self.module.loss).parameters.keys()
         if "kl_weight" in self._loss_args:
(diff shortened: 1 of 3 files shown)


scvi/train/_trainingplans.py
```diff
@@ -187,7 +187,7 @@ def __init__(
         self._n_obs_validation = None
 
         # automatic handling of kl weight
-        self._loss_args = getfullargspec(self.module.loss)[0]
+        self._loss_args = signature(self.module.loss).parameters.keys()
```
adamgayoso, line 190:
```suggestion
        self._loss_args = set(list(signature(self.module.loss).parameters.keys()))
```

scvi/train/_trainingplans.py
```diff
@@ -187,7 +187,7 @@ def __init__(
         self._n_obs_validation = None
 
         # automatic handling of kl weight
-        self._loss_args = getfullargspec(self.module.loss)[0]
+        self._loss_args = set(list(signature(self.module.loss).parameters.keys()))
```
adamgayoso, line 190:
```suggestion
        self._loss_args = set(signature(self.module.loss).parameters.keys())
```

PR #1779, opened 2022-11-03
Implement latent mode for SCANVI + allow models w/o size_factor_key to be latent-ify-able

watiss:


scvi/model/utils/_latent.py
```diff
@@ -0,0 +1,64 @@
...
+    adata: AnnData,
+    mode: LatentDataType,
+    scvi_latent_qzm_key: str,
+    scvi_latent_qzv_key: str,
+    scvi_obs_lib_size_key: str,
+    use_latent_qzm_key: str = "X_latent_qzm",
+    use_latent_qzv_key: str = "X_latent_qzv",
+):
+    """TODO add docstring"""
+    if mode == "dist":
```
adamgayoso, line 21: the latent mode is called posterior parameters now?

scvi/model/utils/_latent.py
```diff
@@ -0,0 +1,64 @@
...
+        adata.obsm[scvi_latent_qzm_key] = adata.obsm[use_latent_qzm_key]
+        adata.obsm[scvi_latent_qzv_key] = adata.obsm[use_latent_qzv_key]
+        adata.obs[scvi_obs_lib_size_key] = np.squeeze(np.asarray(adata.X.sum(axis=1)))
```
adamgayoso, line 24: this currently does not handle the case in which a user passed a size factor key to setup anndata

adamgayoso: in which case you'd have to get that from the anndata manager

adamgayoso: ok actually this is confusing code in the way vae is structured. I feel like we should lightly refactor the pytorch module to have a simpler control flow....

adamgayoso: the simple hack would be to check if the user has set the size factor key, and if they haven't, compute the library size and store it in our existing size factor key.

then the cached inference would return a dummy library which would actually be unused.

adamgayoso (changes requested):

scvi/model/_scvi.py
```diff
@@ -304,12 +285,21 @@ def to_latent_mode(
...
+            raise ValueError(
+                "Latent mode not supported when use_observed_lib_size is False"
+            )
+        reduced_adata = get_reduced_adata(self.adata, mode)
         if mode == _SCVI_LATENT_MODE:
-            self.adata.obsm[_SCVI_LATENT_QZM] = self.adata.obsm[use_latent_qzm_key]
-            self.adata.obsm[_SCVI_LATENT_QZV] = self.adata.obsm[use_latent_qzv_key]
+            reduced_adata.obsm[_SCVI_LATENT_QZM] = self.adata.obsm[use_latent_qzm_key]
+            reduced_adata.obsm[_SCVI_LATENT_QZV] = self.adata.obsm[use_latent_qzv_key]
+            counts = self.adata.X if layer is None else self.adata.layers[layer]
```
adamgayoso, line 296: this should get the adata_manager associated with adata and pull from the registry using `REGISTRY_KEYS.X_KEY`, which would then allow us to pull out the layer argument

```
manager = self.get_anndata_manager(adata)
manager.get_from_registry(..)
```

scvi/model/utils/_latent.py
```diff
@@ -0,0 +1,28 @@
+from anndata import AnnData
+from scipy.sparse import csr_matrix
+
+from scvi._types import LatentDataType
+from scvi.data._constants import _ADATA_LATENT_UNS_KEY, _SCVI_UUID_KEY
+
+
+def get_reduced_adata(
```
adamgayoso, line 8:
```suggestion
def get_reduced_adata_scrna(
```

adamgayoso: also we should add this to the documentation

adamgayoso: and improve the docstring to say that this should work for many scrna cases

adamgayoso (changes requested):

adamgayoso (approved):

Issue #1721, opened 2022-10-06
scvi-tools autotuning with Ray Tune


martinkim0:
This issue will serve as a parent thread for all issues related to supporting autotuning with Ray Tune.

### Basic
- [x] Implement a basic working `ModelTuner` API #1720 

### Docs and tutorials
- [x] Add autotune to docs
- [x] Add tutorial: basic usage with scVI
- [x] Add tutorial: advanced usage with new model class

### Metrics
- [ ] Refactor PL logged metric keys for better visibility and consistency #1798 
- [ ] Allow usage of scib-metrics
- [x] Allow for custom metrics with a callback

### Miscellaneous
- [x] Refactor training mixins for consistency #1796 
- [x] Define tunables at the base class level for simplicity #1799
- [x] More comprehensive testing for the module
- [x] Enable logging with wandb for monitoring
- [ ] Discover better defaults for models #1616

PR #1785, opened 2022-11-10
Implement a basic working `ModelTuner` API with Ray Tune

### LINKED ISSUE #1720: Basic working API of autotuning with Ray Tune

martinkim0:
Closes #1720

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,7 +37,7 @@ jobs:
               id: dependencies
               run: |
                   pip install pytest-cov
-                  pip install .[dev,pymde]
+                  pip install .[autotune,dev,pymde]
 
             # Following checks are independent and are run even if one fails
             - name: Lint with flake8
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,9 +5,9 @@ name: scvi-tools
 
 on:
     push:
-        branches: [main, 0.15.x, 0.16.x, 0.17.x, 0.18.x]
+        branches: [main, 0.15.x, 0.16.x, 0.17.x, 0.18.x, 0.19.x]
     pull_request:
-        branches: [main, 0.15.x, 0.16.x, 0.17.x, 0.18.x]
+        branches: [main, 0.15.x, 0.16.x, 0.17.x, 0.18.x, 0.19.x]
 
 jobs:
     build:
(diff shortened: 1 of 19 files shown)


martinkim0:
This PR just contains the bare essentials for the autotune module. More complexity will be addressed as listed in #1721.

.github/workflows/test.yml
```diff
@@ -37,7 +37,7 @@ jobs:
               id: dependencies
               run: |
                   pip install pytest-cov
-                  pip install .[dev,pymde]
+                  pip install .[autotune,dev,pymde]
```
martinkim0, line 40:
```suggestion
                  pip install .[dev,pymde,autotune]
```

martinkim0:
@adamgayoso Looks like the underlying issue where fork subprocesses were being started "randomly", causing CUDA to not work since it cannot initialize in a fork context, was being caused by this [one line](https://github.com/Lightning-AI/lightning/blob/1.7.7/src/pytorch_lightning/utilities/device_parser.py#L359) in Lightning 1.7. The issue goes away when upgrading to Lightning 1.8, where that line has been replaced. Blocked until we can upgrade to >=1.8.

scvi/autotune/_manager.py
```diff
@@ -0,0 +1,492 @@
...
+from collections import OrderedDict
+from typing import Any, Callable, List, Optional, Tuple
+
+import rich
+
+try:
+    from ray import air, tune
+    from ray.tune.integration.pytorch_lightning import TuneReportCallback
+except ImportError:
+    pass
```
adamgayoso, line 13: should we put a better message here?

martinkim0: Ray import error messages are handled by the `@dependencies` decorator right now, otherwise there would be a message every time scvi is imported. Should I still include it here though?

adamgayoso: no all good I missed that sorry

tests/autotune/test_tuner.py
```diff
@@ -0,0 +1,29 @@
...
+    model_cls = scvi.model.SCVI
+    tuner = scvi.autotune.ModelTuner(model_cls)
+
+    # adata should be setup before passing to `fit`
+    adata = scvi.data.synthetic_iid()
+    with pytest.raises(Exception):
+        tuner.fit(adata)
+
+    model_cls.setup_anndata(adata)
+    results = tuner.fit(adata)
```
adamgayoso, line 21: is there a way to make sure this completes quickly?

martinkim0: good point, we can reduce the number of samples for tests

adamgayoso (approved):

PR #1793, opened 2022-11-17
Add criticism metrics to scvi-tools

watiss:
The main idea is to compute statistics on model-generated posterior predictive samples and on the observed (raw) counts and then compare the two. This is also known as posterior predictive checks (some insights [here](https://www.nxn.se/valent/2022/3/9/vaes-are-explainable-differential-expression-in-scvi)).

Currently the following statistics are implemented:
- Coefficient of variation, cell-wise and gene-wise
- Mann–Whitney U test
- DE genes log-fold change (coming)
- DE genes Expression levels coming)
- DE gene sets overlap (coming)

### DIFF
--- a/scvi/criticism/__init__.py
+++ b/scvi/criticism/__init__.py
@@ -1,0 +1,3 @@
+from ._ppc import PPC, run_ppc
+
+__all__ = ["PPC", "run_ppc"]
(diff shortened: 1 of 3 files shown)


watiss:
Moving this into a separate package...

PR #1802, opened 2022-11-21
Modify `ModelTuner` to be setup method agnostic

martinkim0:
Previously, `ModelTuner` was hard-coded to only use `setup_anndata` for datasets. Now, it should work with `setup_mudata` too.

### DIFF
--- a/scvi/autotune/_manager.py
+++ b/scvi/autotune/_manager.py
@@ -332,14 +332,16 @@ class TunerManager:
         return _reporter
 
     def _validate_resources(self, resources: dict) -> dict:
-        """Validates a resources-use specification."""
+        """Validates a resource-use specification."""
         # TODO: perform resource checking
         return resources
 
-    def _get_setup_kwargs(self, adata: AnnOrMuData) -> dict:
-        """Retrieves the kwargs used for setting up `adata` with the model class."""
+    def _get_setup_info(self, adata: AnnOrMuData) -> Tuple[str, dict]:
+        """Retrieves the method and kwargs used for setting up `adata` with the model class."""
         manager = self._model_cls._get_most_recent_anndata_manager(adata)
-        return manager._get_setup_method_args().get("setup_args", {})
+        setup_method_name = manager._registry.get("_SETUP_METHOD_NAME", "setup_anndata")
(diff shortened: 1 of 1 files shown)


scvi/autotune/_manager.py
```diff
@@ -355,15 +358,16 @@ def _get_trainable(
...
-            max_epochs: int = None,
+            model_cls: BaseModelClass,
+            adata: AnnOrMuData,
+            metric: str,
+            setup_method_name: str,
+            setup_kwargs: dict,
+            max_epochs: int,
         ) -> None:
             model_kwargs, train_kwargs = self._get_search_space(search_space)
             # TODO: generalize to models with mudata
```
watiss, line 369:
```suggestion

```

scvi/autotune/_manager.py
```diff
@@ -332,21 +332,24 @@ def _validate_reporter(
...
         return resources
 
-    def _get_setup_kwargs(self, adata: AnnOrMuData) -> dict:
-        """Retrieves the kwargs used for setting up `adata` with the model class."""
+    def _get_setup_info(self, adata: AnnOrMuData) -> Tuple[str, dict]:
+        """Retrieves the method and kwargs used for setting up `adata` with the model class."""
         manager = self._model_cls._get_most_recent_anndata_manager(adata)
-        return manager._get_setup_method_args().get("setup_args", {})
+        setup_method_name = manager._registry.get("_SETUP_METHOD_NAME", "setup_anndata")
+        setup_args = manager._get_setup_method_args().get("_SETUP_ARGS_KEY", {})
```
watiss, line 343: don't we have a constant for this and `_SETUP_METHOD_NAME` ? If so, shouldn't this just be `_SETUP_ARGS_KEY` without the quotes?

watiss (changes requested):
Just for the comment regarding `_SETUP_ARGS_KEY` vs `"_SETUP_ARGS_KEY"`

scvi/autotune/_manager.py
martinkim0: Good catch, I forgot to take out the quotes

watiss (approved):

PR #1805, opened 2022-11-24
Fix observed variable detection

### LINKED ISSUE #1801: Posterior sampling with Messenger Pyro guides -> huge memory use.
Hi

Posterior sampling with Messenger Pyro guides does not remove observed variables leading to huge memory use.
https://github.com/scverse/scvi-tools/blob/main/scvi/model/base/_pyromixin.py#L184
I don't know whether when sampling from Messenger guides it is possible to easily detect and exclude observed variables. @fritzo any recommendations?

This can be addressed 
1. By adding an additional variable which allows users to exclude observed variables by name (like done here https://github.com/BayraktarLab/cell2location/blob/improved_posterior_quantile/cell2location/models/base/_pyro_mixin.py#L250). 
2. By deleting https://github.com/scverse/scvi-tools/blob/main/scvi/model/base/_pyromixin.py#L182-L185

Related to https://github.com/BayraktarLab/cell2location/pull/144/

vitkl:
Create a separate method to detect observed sites and apply it everywhere necessary to automatically exclude observed sites.

Fixes #1801


vitkl:
I think my pre-commit configs are out of date, sorry. My version of flake doesn't see anything wrong. I can fix this later.

adamgayoso (commented):
What's the status of this @vitkl ?

vitkl:
Still a work in progress, unfortunately, the solution I added doesn't seem to exclude observed variables. I will do a deeper dive some time this year.

vitkl:
I think this should be addressable by removing different procedures for different AutoGuide classes (https://github.com/scverse/scvi-tools/pull/2695).

PR #1810, opened 2022-11-29
New fields: numerical var and varm

vitkl:
These fields are needed when the dataset is split into batches of both cells and variables.

### DIFF
--- a/scvi/data/fields/_var_field.py
+++ b/scvi/data/fields/_var_field.py
@@ -1,0 +1,96 @@
+import logging
+from typing import Optional
+
+import numpy as np
+import rich
+from anndata import AnnData
+from pandas.api.types import CategoricalDtype
+
+from scvi.data import _constants
+from scvi.data._utils import _make_column_categorical, get_anndata_attribute
+
+from ._base_field import BaseAnnDataField
+from ._mudata import MuDataWrapper
+
+logger = logging.getLogger(__name__)
+
+
+class BaseVarField(BaseAnnDataField):
+    """
+    An abstract AnnDataField for .var attributes in the AnnData data structure.
+    Parameters
+    ----------
+    registry_key
+        Key to register field under in data registry.
+    var_key
+        Key to access the field in the AnnData var mapping. If None, defaults to `registry_key`.
+    required
+        If False, allows for `var_key is None` and marks the field as `is_empty`.
+    """
+
+    _attr_name = _constants._ADATA_ATTRS.VAR
+
+    def __init__(
+        self, registry_key: str, var_key: Optional[str], required: bool = True
+    ) -> None:
+        super().__init__()
+        if required and var_key is None:
+            raise ValueError(
+                "`var_key` cannot be `None` if `required=True`. Please provide an `var_key`."
+            )
+        self._registry_key = registry_key
+        self._attr_key = var_key
+        self._is_empty = var_key is None
+
+    @property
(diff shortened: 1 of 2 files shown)


scvi/data/fields/_var_field.py
```diff
@@ -0,0 +1,92 @@
...
+
+from ._base_field import BaseAnnDataField
+
+logger = logging.getLogger(__name__)
+
+
+class BaseVarField(BaseAnnDataField):
+    """
+    An abstract AnnDataField for .var attributes in the AnnData data structure.
+    Parameters
```
martinkim0, line 17:
```suggestion
    An abstract AnnDataField for .var attributes in the AnnData data structure.
    
    Parameters
```

scvi/data/fields/_var_field.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @property
+    def is_empty(self) -> bool:  # noqa: D102
+        return self._is_empty
+
+
+class NumericalVarField(BaseVarField):
+    """
+    An AnnDataField for numerical .var attributes in the AnnData data structure.
+    Parameters
```
martinkim0, line 61:
```suggestion
    An AnnDataField for numerical .var attributes in the AnnData data structure.
    
    Parameters
```

scvi/data/fields/_varm_field.py
```diff
@@ -0,0 +1,167 @@
...
+    def attr_name(self) -> str:  # noqa: D102
+        return self._attr_name
+
+
+class VarmField(BaseVarmField):
+    """
+    An AnnDataField for an .varm field in the AnnData data structure.
+    In addition to creating a reference to the .varm field, stores the column
+    keys for the varm field in a more accessible .uns attribute.
+    Parameters
```
martinkim0, line 47:
```suggestion
    An AnnDataField for an .varm field in the AnnData data structure.
    
    In addition to creating a reference to the .varm field, stores the column
    keys for the varm field in a more accessible .uns attribute.
    
    Parameters
```

scvi/data/fields/_varm_field.py
```diff
@@ -0,0 +1,167 @@
...
+        if self.is_count_data and not _check_nonnegative_integers(varm_data):
+            warnings.warn(
+                f"adata.varm['{self.attr_key}'] does not contain unnormalized count data. "
+                "Are you sure this is what you want?"
+            )
+
+    def _setup_column_names(self, adata: AnnData) -> Union[list, np.ndarray]:
+        """
+        Returns a list or NumPy array of column names that will be used for the relevant .varm data.
+        If the ``colnames_uns_key`` was specified, then the columns stored in that
```
martinkim0, line 107:
```suggestion
        Returns a list or NumPy array of column names that will be used for the relevant .varm data.
        
        If the ``colnames_uns_key`` was specified, then the columns stored in that
```

martinkim0 (approved):
LGTM, ready to merge after precommit and Codacy issues are addressed!

adamgayoso (changes requested):
Thanks! Would you be able to add the classes to our docs too?

adamgayoso:
Actually, thinking about this again, there's going to be a lot of copy/paste between the obs analogs. I think it makes sense to go deeper with the abstraction so we can reduce code duplication. We are on it from our side @vitkl

martinkim0:
Closed in favor of #1821

PR #1821, opened 2022-12-12
Generalize `obs` and `obsm` fields to `var` and `varm`

martinkim0:
Opened in favor of #1810. These fields are needed when the dataset is split into batches of both cells and variables. Thank you @vitkl for the initial PR!


adamgayoso (changes requested):
This looks good but two things.

1. How do you like the names?
2. Very nitpicky but can you git mv the _obsm_field file and then copy paste your new changes into it so we can keep the history of Justin's work?

martinkim0:
@adamgayoso 
1. Hmm I feel like the names are pretty verbose right now, but not sure how to change them and keep them unambiguous. Any suggestions?
2. Yeah of course

scvi/data/fields/_matrix_field.py
```diff
@@ -0,0 +1,660 @@
...
+    _verify_and_correct_data_format,
+)
+
+from ._base_field import BaseAnnDataField
+from ._mudata import MuDataWrapper
+
+logger = logging.getLogger(__name__)
+
+
+class BaseMatrixField(BaseAnnDataField):
```
adamgayoso, line 24:
```suggestion
class BaseArrayLikeField(BaseAnnDataField):
```

Thoughts?

martinkim0: Sounds reasonable, although the reason it was named BaseMatrixField was because I thought varm and obsm both stand for matrix.

adamgayoso: ah I see, I'm not sure. But when I think of matrix I think of np matrix, where arraylike better captures the data structures

martinkim0: Fair enough, I think I'm fine with either so I'll go with your suggestion.

PR #1823, opened 2022-12-12
Remove poetry in favor of hatch

adamgayoso:
This move sets us up to mirror the consolidated scverse cookiecutter structure

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,20 +20,13 @@ jobs:
                   python-version: "3.9"
 
             - name: Install poetry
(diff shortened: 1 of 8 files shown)


.github/workflows/release.yml
```diff
@@ -20,20 +20,13 @@ jobs:
                   python-version: "3.9"
 
             - name: Install poetry
```
martinkim0, line 22:
```suggestion
            - name: Install hatch
```

martinkim0 (approved):
LGTM

PR #1831, opened 2023-01-04
Add `TunableMixin` and `TuneAnalysis`

### LINKED ISSUE #1799: Define `_tunables` class property at the base level
This would make it easier to expose model classes to `autotune.ModelTuner` as any model class inheriting from `BaseModelClass` would work properly out of the box.

martinkim0:
Closes #1799


scvi/autotune/_manager.py
```diff
@@ -86,31 +88,58 @@ def _cls_to_tunable_type(cls: Any) -> str:
                     return tunable_type
             return "unknown"
 
+        def _parse_func_params(func: Callable, parent: Any, tunable_type: str) -> dict:
+            # parse function docstring for parameter descriptions
+            docstring = None
+            if func.__name__ == "__init__":
+                # __init__ func docstrings are the class's docstring
+                docstring = parse(parent.__doc__)
```
adamgayoso, line 96: I think we might be able to use docrep for this instead of introducing another dependency?

martinkim0: Might just remove the functionality to print out keyword argument descriptions, I think it's too overloaded with it.

scvi/train/_trainer.py
```diff
@@ -1,22 +1,21 @@
...
+from scvi.autotune._types import Tunable, TunableMixin
 
 from ._callbacks import LoudEarlyStopping
 from ._logger import SimpleLogger
 from ._progress import ProgressBar
 from ._trainingplans import PyroTrainingPlan
 
 
-class Trainer(pl.Trainer):
+class Trainer(pl.Trainer, TunableMixin):
```
adamgayoso, line 18: why is the mixin going last here? This applies to a lot of them.

martinkim0: Should it be going first? Wasn't sure if the order mattered

adamgayoso: It does matter if there were collisions in method names (lookup MRO python), it's probably best to have the mixins before the base class that provides most functionality

martinkim0:
@adamgayoso Any idea what could be happening with docs fail? Haven't been able to figure it out

martinkim0:
Looks like all new docs builds are failing with the same error

.gitignore
```diff
@@ -113,6 +113,7 @@ ENV/
 .floo
 .flooignore
 
+tests/data
```
adamgayoso, line 116: why is this in gitignore?

adamgayoso (changes requested):

.gitignore
martinkim0: The autotune tests leave some log files there

martinkim0: I can change where they're logged

adamgayoso (approved):

PR #1847, opened 2023-01-11
more context for pyro modules

adamgayoso:


vitkl:
Why did you decide in favour of `pyro_guide(*args, **kwargs)` over `pl_module.differentiable_loss_fn(pyro_model, pyro_guide, *args, **kwargs)`? @adamgayoso

adamgayoso:
Running the guide runs the model. I figured it's no worse and potentially cheaper computationally (no elbo computation)

vitkl:
Is this true for the Messenger guides? I guess yes because the test doesn't fail.

PR #1849, opened 2023-01-12
WIP: Testing AutoHierarchicalNormalMessenger

vitkl:
This tests the use of a more complex AutoHierarchicalNormalMessenger guide both with and without local variables.
This guide constructs the posterior distribution conditional on parent variables - automatically replicating hierarchical variable dependencies of the model. This can improve the performance of hierarchical models. However, this guide operates in a very different way from standard AutoNormal (including lazy creation of new parameters if they appear in the model) - required to be tested.


adamgayoso:
Closing this for now.

Issue #1854, opened 2023-01-14
potential miscomputation in get_reconstruction_error of scVI


eroell:
When calling `get_reconstruction_error` of the `scVI` model, `compute_reconstruction_error` gets called.

Filling/summing an empty dictionary `log_lkl` starting from line [48](https://github.com/scverse/scvi-tools/blob/9e13c6e6d2b12de3afe5f9ec5bb4b69e19b4be64/scvi/model/base/_log_likelihood.py#L48) looks as follows:

```python
        for key, value in rec_loss_dict.items():
            if key in log_lkl:
                log_lkl[key] += torch.sum(value).item()
            else:
                log_lkl[key] = 0.0
```
I think the correct computation would be

```python
        for key, value in rec_loss_dict.items():
            if key in log_lkl:
                log_lkl[key] += torch.sum(value).item()
            else:
                log_lkl[key] = torch.sum(value).item()
```

My apologies in case its correct & I misunderstand the computation involved.

adamgayoso:
Indeed this looks like a bug, thank you for spotting. It seems like the reconstruction loss would be rescaled incorrectly as it would not consider the loss of the first minibatch.

Would you be able to make a PR with this change asap? Thanks!

adamgayoso:
it would also be great to add a test

eroell:
should such a test evaluate the numeric result `get_reconstruction_error`? it appears to me that generally the result of `model` methods are not further tested [here](https://github.com/scverse/scvi-tools/blob/9e13c6e6d2b12de3afe5f9ec5bb4b69e19b4be64/tests/models/test_models.py#L182)...

if yes to numeric result evaluation, a first thought on a test I could insert there, relying on `get_likelihood_parameters`:

```python
params = model.get_likelihood_parameters(adata, give_mean=True)
px = scvi.distributions._negative_binomial.ZeroInflatedNegativeBinomial(mu=torch.tensor(params['mean']),
                                                            theta=torch.tensor(params['dispersions']),
                                                            zi_logits=torch.logit(torch.tensor(params['dropout'])))
zinb_ll = torch.mean(px.log_prob(torch.tensor(adata.X)).sum(axis=1))

assert np.isclose(rec_error['reconstruction_loss'], zinb_ll)
```
the two quantities `zinb_ll` and `rec_error` do deviate by 1-2 percent when I tested this. I suppose numeric issues? (compared to arbitrarily big deviation before the fix, depending on the batch size)

[...]

PR #1857, opened 2023-01-18
reconstruction error fix for #1854

eroell:
small change as suggested in the issue #1854


PierreBoyeau (approved):
this sounds right! thanks so much for spotting this!

eroell:
you're more than welcome! kindly let me know if more action from my side would be required to close this workflow.

PR #1863, opened 2023-01-21
update lightning to >=1.9, bump other minimums, stage 0.20.0b1

adamgayoso:


scvi/__init__.py
```diff
@@ -3,12 +3,6 @@
 # Set default logging handler to avoid logging with logging.lastResort logger.
 import logging
 
-try:
-    # necessary as importing scvi after ray causes kernel crash
-    from ray import tune  # noqa
-except ImportError:
-    pass
-
```
martinkim0, line 11: Does it work now without this block?

adamgayoso: I was having no issues. Can you check as well?

martinkim0: Yeah I'm still having the issue unfortunately. I usually check by running
```
import scvi
import ray
ray.init()
ray.shutdown()
```

adamgayoso: I'm not having that issue and the tests are passing, I'm wondering what could be happening

martinkim0 (approved):

PR #1884, opened 2023-01-31
Use `cellxgene` download in HubModel when needed

### LINKED ISSUE #1878: [Hub] read large training data should work with cellxgene urls
https://github.com/scverse/scvi-tools/blob/b2858d1fe33c897f8d6f46d71ecad4c6a7f72b4a/scvi/hub/hub_model.py#L299

We can use this gist for an idea of how to get cellxgene data

https://gist.github.com/ivirshup/f1a1603db69de3888eacb4bdb6a9317a

In fact, we can write a function in `scvi.data` that takes a cellxgene link and downloads the data

watiss:
Fixes #1878

### DIFF
--- a/scvi/data/__init__.py
+++ b/scvi/data/__init__.py
@@ -63,4 +63,5 @@ __all__ = [
     "organize_multiome_anndatas",
     "pbmc_seurat_v4_cite_seq",
     "add_dna_sequence",
+    "cellxgene",
 ]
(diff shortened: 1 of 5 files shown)


adamgayoso (changes requested):
It would be great to document in the docstrings that this url can be a cxg explorer session url

adamgayoso (approved):

PR #1907, opened 2023-02-17
Upperbound jax, jaxlib due to jit/pjit issue

martinkim0:
`JaxTrainingPlan` fails in the first epoch with an error related to jit/pjit:
```
Received incompatible devices for jitted computation. Got argument  of JaxTrainingPlan.jit_training_step with shape float32[2,40000] and device ids [0] on platform GPU and pjit inside jit with device ids [0] on platform CPU
```


justjhong (approved):

adamgayoso:
meeseeksdev backport to 0.20.x

PR #1936, opened 2023-03-01
Log individual loss terms in MultVI LossOutput

martinkim0:
@marianogabitto


adamgayoso (approved):

adamgayoso:
@martinkim0 can you add a test to check that it gets into the history?

Issue #1926, opened 2023-02-23
Implement PoissonVI to model scATAC-seq fragment counts


lauradmartens:
**Is your feature request related to a problem? Please describe.**
PeakVI currently only supports modelling of binarized scATAC-seq data. In our [preprint](https://www.biorxiv.org/content/10.1101/2022.05.04.490536v1), we showed that it can be beneficial to actually model the scATAC-seq fragment counts.

**Describe the solution you'd like**
As previously discussed with @adamgayoso, it would be good to add the PoissonVI model to scvi, that users can also model the fragment count data.

adamgayoso:
Hi @lauradmartens, can you please describe again the difference between scVI with a poisson likelihood and the poissonVI model?

lauradmartens:
Hi Adam!

The model itself is similar but with some changes:
* AnnData setup checks that users input fragment counts and not binarized counts or read counts
* Function `reads_to_fragments` to approximately convert read counts to fragment counts
* Default architecture (number of layers etc.) and training procedure (learning rate etc.) is the one of PeakVI for ATAC data
* Function names as in PeakVI and adapted to ATAC terminology `get_accessibility_estimates`, `differential_accessibility`
* Possibility to output an accessibility probability from the Poisson rate parameter estimations in `get_accessibility_estimates`
*  And lastly but I have not implemented this yet, it might be interesting to exclude the region-specific bias (bias in the last layer) from the normalized accessibility estimates

But I am very open to also changing the structure or merging it with existing models if you think there is a more elegant solution.

marianogabitto:
Laura,
      I have already implemented but never push the commit. Let me know what you want to do about this.

Mariano

lauradmartens:
Hi Mariano, 

Cool to hear that! Sure, if this makes things easier! I am happy to help with input and documentation.

adamgayoso:
I need to think more about it, but perhaps we can add the poisson functionality directly into PeakVI, and then have the option to create a PoissonPeakVI model that just inherits peakvi and sets poisson to true.

The second part would increase expose for people to cite your publication, but isn't strictly necesary code wise

lauradmartens:
Thanks Adam! In this case, I think I would prefer the second option.

marianogabitto:
In my code, I included in MultiVI. To be sincere, I did not see a major difference compared to PeakVI. A couple of thoughts (caveat, I read the paper and implemented a while ago so I might be wrong in my comments):
[...]

PR #1946, opened 2023-03-07
Add reads to fragments function

lauradmartens:
Adds a function in `data._preprocessing` to convert 10x scATAC-seq read counts to approximate fragment counts.
#1926


scvi/data/_preprocessing.py
```diff
@@ -449,6 +449,39 @@ def add_dna_sequence(
         assert len(seqs) == len(chrom_df)
         seq_dfs.append(pd.DataFrame(seqs, index=chrom_df.index))
 
-    sequence_df = pd.concat(seq_dfs, axis=0).loc[adata.var_names]
-    adata.varm[sequence_varm_key] = sequence_df
-    adata.varm[code_varm_key] = sequence_df.applymap(_dna_to_code)
+    output_df = pd.concat(output_dfs, axis=0).loc[adata.var_names]
```
ivirshup, line 452:
```suggestion
    output_df = pd.concat(seq_dfs, axis=0).loc[adata.var_names]
```

Is this what was meant?

scvi/data/_preprocessing.py
```diff
@@ -452,3 +452,36 @@ def add_dna_sequence(
...
+def reads_to_fragments(
+    adata: anndata.AnnData, layer: Optional[str] = None, copy: bool = False
+) -> Optional[anndata.AnnData]:
+    """Convert read counts to appoximate fragment counts.
+
+    Parameters
+    ----------
+    adata
+        AnnData object that contains read counts.
+    layer
```
adamgayoso, line 466: I think this function should either:

1. just return the new scipy sparse matrix based on the input layer, then the user can put it wherever they want and we could remove the copy option.

or 

2. Have a read_layer argument for the input and a `fragment_layer` argument for the output layer, which should default to `fragments`

adamgayoso (changes requested):
Thanks for adding this util! Just a small comment

adamgayoso:
Ah and one more thing, would you be able to add the function to the docs? It should be added to this page:

https://github.com/scverse/scvi-tools/blob/main/docs/api/user.md#basic-preprocessing

scvi/data/_preprocessing.py
lauradmartens: I opted for the second option now.

scvi/data/_preprocessing.py
```diff
@@ -452,3 +452,32 @@ def add_dna_sequence(
...
+    -------
+    Adds layer with fragment counts in `.layers[fragment_layer]`.
+    """
+    if read_layer:
+        data = np.ceil(adata.layers[read_layer].data / 2)
+    else:
+        data = np.ceil(adata.X.data / 2)
+
+    adata.layers[fragment_layer] = adata.X.copy()
+    adata.layers[fragment_layer].data = data
```
adamgayoso, line 483:
```suggestion
    data = adata.layers[read_layer].data if read_layer else adata.X.data
    adata.layers[fragment_layer] =  adata.layers[read_layer].copy() is read_layer else adata.X.copy()
    adata.layers[fragment_layer].data = np.ceil(adata.layers[fragment_layer].data / 2)

```

adamgayoso: Would be great to also add a check with https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.issparse.html as this will give an uninformative error if the data is not in a sparse format.

lauradmartens: Thank you! I wrote it now so that is can handle both. Let me know if you'd prefer it differently.

scvi/data/_preprocessing.py
```diff
@@ -452,3 +453,34 @@ def add_dna_sequence(
     sequence_df = pd.concat(seq_dfs, axis=0).loc[adata.var_names]
     adata.varm[sequence_varm_key] = sequence_df
     adata.varm[code_varm_key] = sequence_df.applymap(_dna_to_code)
+
+
+def reads_to_fragments(
+    adata: anndata.AnnData,
+    read_layer: Optional[str] = None,
+    fragment_layer: Optional[str] = "fragments",
```
adamgayoso, line 461:
```suggestion
    fragment_layer: str = "fragments",
```

adamgayoso (approved):

adamgayoso (changes requested):
Actually @lauradmartens would you be able to add a test?

PR #1950, opened 2023-03-08
Deprecate `use_gpu` for `accelerator` and `devices`

martinkim0:


docs/release_notes/index.md
```diff
@@ -22,15 +22,16 @@ is available in the [commit logs](https://github.com/YosefLab/scvi-tools/commits
...
 
 #### Changed
 
+-   Setting up ST data in {class}`~scvi.external.SpatialStereoscope.from_rna_model` is deprecated. ST data should be set up using {class}`~scvi.external.SpatialStereoscope.setup_anndata` {pr}`1803`.
 -   Use ruff for fixing and linting {pr}`1921`, {pr}`1941`.
 -   Use sphinx autodoc instead of sphinx-autodoc-typehints {pr}`1941`.
 -   Remove .flake8 and .prospector files {pr}`1923`.
 -   Log individual loss terms in {meth}`scvi.module.MULTIVAE.loss` {pr}`1936`.
--   Setting up ST data in {class}`~scvi.external.SpatialStereoscope.from_rna_model` is deprecated. ST data should be set up using {class}`~scvi.external.SpatialStereoscope.setup_anndata` {pr}`1803`.
+-   Deprecate `use_gpu` in favor of `accelerator` and `devices` Lightning arguments {pr}`1950`.
```
adamgayoso, line 34: Let's add this to 0.20.3

scvi/model/_utils.py
```diff
@@ -1,22 +1,92 @@
...
 logger = logging.getLogger(__name__)
 
 
+def parse_device_args(
+    use_gpu: Optional[Union[str, int, bool]] = None,
+    accelerator: Optional[str] = None,
+    devices: Optional[Union[List[int], str, int]] = None,
+    return_device: Optional[Literal["torch", "jax"]] = None,
+) -> Tuple:
+    """Parses `accelerator` and `devices` arguments.
```
adamgayoso, line 27: can you document what happens by default when everything is none?

adamgayoso (changes requested):

scvi/dataloaders/_data_splitting.py
```diff
@@ -88,14 +89,21 @@ def __init__(
...
         self.train_size = float(train_size)
         self.validation_size = validation_size
         self.data_loader_kwargs = kwargs
+
+        if use_gpu is not None:
+            warnings.warn(
+                "The `use_gpu` argument is deprecated in v0.20 and will be removed in v1.0. "
+                "Please directly pass in `pin_memory` directly for device memory pinning.",
+                DeprecationWarning,
+            )
```
adamgayoso, line 106: this will cause a warning for anyone using a gpu right? I think most users don't go this deep into the code, nor do I think anyone is actively turning on this pin memory feature.

I think it's safe (although breaking) to remove this tiny feature altogether and expose pin memory explicitly here with default equal to False. We can then

1. deprecate `settings.dl_pin_memory_gpu_training` if someone tries to set it
2. and expose data_splitter_kwargs in the train method

scvi/model/_utils.py
```diff
@@ -1,22 +1,92 @@
...
+    valid = [None, "torch", "jax"]
+    if return_device not in valid:
+        raise InvalidParameterError(
+            param="return_device", value=return_device, valid=valid
+        )
+
+    if return_device == "torch":
+        pl_connector = AcceleratorConnector(accelerator=accelerator, devices=devices)
+        accelerator_ = pl_connector._accelerator_flag
+        devices_ = pl_connector._devices_flag
```
adamgayoso, line 69: is there a way to avoid using these private attrs? My experience with their package is that it will break sooner than later

scvi/model/_utils.py
```diff
@@ -1,22 +1,92 @@
...
+            device = torch.device(f"{accelerator_}:{devices_}")
+        elif isinstance(devices_, list):
+            device = torch.device(f"{accelerator_}:{devices_[0]}")
+        elif isinstance(devices_, str):
+            device = torch.device(devices_)
+        return accelerator, devices, device
+    elif return_device == "jax":
+        raise NotImplementedError("Jax device parsing not implemented.")
+    else:
+        return accelerator, devices
```
adamgayoso, line 83: one caveat here is that scvi will currently work with only one device. I'm wondering if we should alter this function to throw a warning if accelerator != cpu and devices is more than 1.

adamgayoso (changes requested):

PR #1957, opened 2023-03-14
Library size estimation: changing prior variance and n_hidden

vitkl:
Implementing changes proposed in https://github.com/scverse/scvi-tools/issues/1903#issuecomment-1466746644

1. the prior `l_sigma` is an overestimate of the total variance attributed to the technical effect. A potential fix (and an easy fix) would be adding a hyperparameter to allow the user to reduce variance using a simple weight.

2. Indeed it is possible that softplus would make the computation more stable. Is it easy to change to softplus exclusively for size factors 

3. I assume that the encoder network size, n_hidden is the same both for `z` and `l` which in my example is a pretty large number. This would mean that the network is massively overparameterised. In my trials of amortizing inference for cell2location I observed that such 1d parameters need to be amortized with a much smaller network to achieve numerical stability and avoid loss increase (n_hidden=10). Is it possible to change n_hidden exclusively for this parameter? If yes I would like to try this and report results.

4. [not use_observed_lib_size means scale_activation="softplus"](https://github.com/vitkl/scvi-tools/commit/ef3978ade7e02a5ea48c4ae8810d90c180fc9fd1)

### DIFF
--- a/scvi/module/_vae.py
+++ b/scvi/module/_vae.py
@@ -86,6 +86,8 @@ class VAE(BaseMinifiedModeModuleClass):
     library_log_vars_weight
         Weight that allows adjusting the expected variance in library sizes that can be attributed to technical 
         rather than biological effect. Set library_log_vars_weight < 1.0 to regularise technical effect.
+    library_n_hidden
+        Number of hidden layers to use for the encoder that learns library sizes. Default: use n_hidden.
     var_activation
         Callable used to ensure positivity of the variational distributions' variance.
         When `None`, defaults to `torch.exp`.
@@ -117,6 +119,7 @@ class VAE(BaseMinifiedModeModuleClass):
         library_log_means: Optional[np.ndarray] = None,
         library_log_vars: Optional[np.ndarray] = None,
         library_log_vars_weight: float = 1.0,
+        library_n_hidden: Optional[int] = None,
         var_activation: Optional[Callable] = None,
     ):
         super().__init__()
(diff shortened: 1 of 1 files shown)


vitkl:
Summary of the changes

1. Additive background term for every batch (batch_key):
        * Correcting for ambient RNA while not requiring the use of empty droplets & associated scalability/access issues
        * Combines cell2location batch correction features with scVI, but requires changing decoder transformation to softplus, encoder for y_c:
                   `d_cg ~ NB((mu_cg + s_eg) * y_c, alpha_eg)`              `d_cg ~ NB(mu_cg * y_c, alpha_eg)`
                   `mu_cg = softplus(DecoderNN(latent_cf))`                  `mu_cg = softmax_c(DecoderNN(latent_cf))`
                   `y_c = EncoderNN(d_cg, n_hidden=32)`
2. Regularising overdispersion alpha_eg to push NB towards Poisson:
        * Adding regularization term to loss
2. More constrained decoder not using batch_key only other categorical covariates.
        * Covariates provided to DecoderNN act like multiplicative effect in Combat, batch_key over-corrects.
        * Providing batch_key to EncoderNN still help inform latent space & normalization y_c.
4. Using much larger decoder and encoder to accommodate more diverse data:
                   `mu_cg = softplus(DecoderNN(latent_cf , n_hidden=3000, n_layers=1))`
                   `y_c = EncoderNN(d_cg, n_hidden=3000, n_latent=700, n_layers=1)`
5. Using LayerNorm in DecoderNN and EncoderNN, nn.Dropout as first not last operation in each layer

PR #1961, opened 2023-03-16
Upgrade to lightning 2.0

### LINKED ISSUE #1959: Test against lightning 2.0 release candidate

adamgayoso:
Fixes #1959 

Fixes autotune docs as well and makes hyperopt default searcher


adamgayoso:
@martinkim0 this is mostly ready to go, but we should wait for the accelerator/devices changes.

tests/models/test_models.py
```diff
@@ -192,7 +192,7 @@ def test_scvi(save_path):
         size_factor_key="size_factor",
     )
     model = SCVI(adata, n_latent=n_latent)
-    model.train(1, check_val_every_n_epoch=1, train_size=0.5)
+    model.train(12, check_val_every_n_epoch=1, train_size=0.5)
```
adamgayoso, line 195:
```suggestion
    model.train(1, check_val_every_n_epoch=1, train_size=0.5)
```

adamgayoso:
@martinkim0 the tuner test is failing now because validation loss is not being tracked, which seems to make sense to me as it's not tracked by default in SCVI. Do you know how it was being tracked before?

adamgayoso:
ah I think it's an issue with the ray tune callback not being compatible with lightning 2.0

adamgayoso:
https://github.com/ray-project/ray/issues/33426

PR #1970, opened 2023-03-17
sphinx autodoc typehints is back, use docrep for de docs

adamgayoso:
To align us back again with cookiecutter.

I removed it originally because it was giving us problems but now I fixed it by removing our custom reusable docs decorator and instead using docrep


docs/conf.py
```diff
@@ -16,7 +16,7 @@
 
 info = metadata("scvi-tools")
 project_name = info["Name"]
-author = info["Author"]
+author = "test"
```
adamgayoso, line 19:
```suggestion
author = info["Author"]
```

PR #1978, opened 2023-03-22
Deprecate `use_gpu` for `accelerator` and `devices`

### LINKED ISSUE #1860: Refactor `use_gpu`
We should deprecate `use_gpu` in exchange for using lightning arguments directly `accelerator` and `devices`. 

This will involve at least two breaking changes, after training we put the model back onto the one device it was trained on (lightning defaults to everything on the cpu after training). We will need to follow what lightning does here and then the user can use `model.to_device()` after training. Second, we will need to remove `use_gpu` from load and defer to the manual device placement workflow.

martinkim0:
Closes #1860

### DIFF
--- a/scvi/utils/_docstrings.py
+++ b/scvi/utils/_docstrings.py
@@ -189,13 +189,13 @@ devices
     The devices to use. Can be set to a non-negative index (`int` or `str`), a sequence
     of device indices (`list` or comma-separated `str`), the value `-1` to indicate all
     available devices, or `"auto"` for automatic selection based on the chosen
-    `accelerator`. If `"auto"` and `accelerator` is not determined to be `"cpu"`, then
-    `devices` will be set to the first available device."""
+    `accelerator`. If set to `"auto"` and `accelerator` is not determined to be `"cpu"`,
+    then `devices` will be set to the first available device."""
 
 param_device = """\
 device
     The device to use. Can be set to a non-negative index (`int` or `str`) or `"auto"`
-    for automatic selection based on the chosen accelerator. If `"auto"` and
+    for automatic selection based on the chosen accelerator. If set to `"auto"` and
     `accelerator` is not determined to be `"cpu"`, then `device` will be set to the
     first available device."""
 


scvi/utils/_docstrings.py
```diff
@@ -170,3 +170,42 @@
...
+param_accelerator = """\
+accelerator
+    Supports passing different accelerator types ("cpu", "gpu", "tpu", "ipu", "hpu",
+    "mps, "auto") as well as custom accelerator instances."""
+
+param_devices = """\
+devices
+    The devices to use. Can be set to a positive number (int or str), a sequence of
+    device indices (list or str), the value -1 to indicate all available devices should
+    be used, or "auto" for automatic selection based on the chosen accelerator."""
```
adamgayoso, line 191: let's state that auto will select only 1 device

"If "auto" and non-cpu accelerators are available, only 1 device will be used" or something similar.

scvi/model/base/_jaxmixin.py
```diff
@@ -96,7 +99,7 @@ def train(
                 training_plan=self.training_plan,
                 data_splitter=data_splitter,
                 max_epochs=max_epochs,
-                use_gpu=False,
+                use_gpu=False,  # TODO: remove this
```
adamgayoso, line 102: let's remove this and add `accelerator="cpu"` instead

docs/release_notes/index.md
```diff
@@ -9,13 +9,20 @@ is available in the [commit logs](https://github.com/YosefLab/scvi-tools/commits
 
 ## [Unreleased]
 
+### 1.1.0 (2023-MM-DD)
+
+#### Removed
+
+-   Remove deprecated `use_gpu` in favor of PyTorch Lightning arguments `accelerator` and `devices` {pr}`xxxx`.
+
```
adamgayoso, line 17:
```suggestion

## Version 1.0
```

adamgayoso (changes requested):
lgtm other than small stuff

scvi/model/base/_base_model.py
```diff
@@ -623,7 +627,12 @@ def load(
         >>> model.get_....
         """
         load_adata = adata is None
-        _, _, device = parse_use_gpu_arg(use_gpu)
+        _, _, device = parse_device_args(
+            use_gpu=use_gpu,
+            accelerator=accelerator,
+            devices=device,
+            return_device="torch",
```
adamgayoso, line 634: there are certain instances where we only expect one device. should we add a flag to the parse function like `validate_single_device` ? of course multiple devices would make sense always for cpu according to how lightning handles things

adamgayoso (approved):
this LGTM, but let's catch the single device case with the validation and wait to merge the lightning 2.0 pr. Then this PR will need to change one import statement for lightning

scvi/model/_utils.py
```diff
@@ -1,19 +1,102 @@
...
 import warnings
 from collections.abc import Iterable as IterableClass
-from typing import Dict, List, Optional, Sequence, Tuple, Union
+from typing import Dict, List, Literal, Optional, Sequence, Tuple, Union
 
+import jax
 import numpy as np
 import scipy.sparse as sp_sparse
 import torch
+from pytorch_lightning.trainer.connectors.accelerator_connector import (
```
adamgayoso, line 10: from lighting.pytorch

PR #2009, opened 2023-04-19
update totalvi/multivi protein loss to fix handling of missing proteins

### LINKED ISSUE #2008: TotalVI loss computation excluding unobserved proteins
How sure are you about this line here?
https://github.com/scverse/scvi-tools/blob/cb17e10b2a82dc95c0c8c261e1aef7781c3b0df9/scvi/module/_totalvae.py#L315

I guess it is supposed to do normal masking, but this mask_scatter_ method is filling the nonmasked entries just with the first entries of the loss matrix not taking into account the position of the masking matrix (https://pytorch.org/docs/stable/generated/torch.Tensor.masked_scatter_.html)

If I am correct, `temp_pro_loss_full = pro_batch_mask_minibatch.bool() * full_loss` should i think work.

adamgayoso:
Masked scatter was not actually filling in the right elements. Masked fill would do this but only works on 0d tensors. Here we just multiply with the mask to stop the gradients from backpropagating. 

Fixes #2008

### DIFF
--- a/docs/release_notes/index.md
+++ b/docs/release_notes/index.md
@@ -27,6 +27,7 @@ is available in the [commit logs](https://github.com/YosefLab/scvi-tools/commits
 #### Fixed
 
 -   Fix creation of minified adata by copying original uns dict {pr}`2000`. This issue arises with anndata>=0.9.0.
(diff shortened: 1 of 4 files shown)


adamgayoso:
Tutorials were updated in https://github.com/scverse/scvi-tutorials/commit/c2fc6d100ecc28e716f9ffc96bc68af48a7733b4

PR #2016, opened 2023-04-28
Add matrix job to test against pre-release version of dependencies

grst:


grst:
Looks like this already caught a couple of issues :)

PR #2017, opened 2023-05-01
remove seqfish and seqfish plus datasets

adamgayoso:
this also allows us to drop openpyxl as a dependency

These were legacy datasets and not really used.


martinkim0 (approved):
LGTM

PR #2023, opened 2023-05-04
Track more classification metrics in training plans

### LINKED ISSUE #1958: Track more classification metrics in `SemiSupervisedTrainingPlan` and `ClassifierTrainingPlan`

martinkim0:
Closes #1958

### DIFF
--- a/scvi/module/base/_base_module.py
+++ b/scvi/module/base/_base_module.py
@@ -52,6 +52,12 @@ class LossOutput:
     kl_global
         Global KL divergence term. Should be one dimensional with one value. If a tensor, converted to
         a dictionary with key "kl_global" and value as tensor.
+    classification_loss
+        Classification loss.
+    logits
+        Logits for classification.
+    true_labels
+        True labels for classification.
     extra_metrics
         Additional metrics can be passed as arrays/tensors or dictionaries of
         arrays/tensors.
@@ -74,7 +80,9 @@ class LossOutput:
     reconstruction_loss: LossRecord | None = None
     kl_local: LossRecord | None = None
     kl_global: LossRecord | None = None
-    classification: dict[str, Tensor] | None = field(default_factory=dict)
+    classification_loss: LossRecord | None = None
+    logits: Tensor | None = None
+    true_labels: Tensor | None = None
     extra_metrics: dict[str, Tensor] | None = field(default_factory=dict)
     n_obs_minibatch: int | None = None
     reconstruction_loss_sum: Tensor = field(default=None, init=False)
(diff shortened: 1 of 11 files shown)


scvi/train/_trainingplans.py
```diff
@@ -708,6 +709,37 @@ def __init__(
...
+
+        accuracy = Accuracy(task="multiclass", num_classes=n_classes).to(logits.device)(
+            predicted_labels, true_labels
+        )
+        f1 = F1Score(task="multiclass", num_classes=n_classes).to(logits.device)(
+            predicted_labels, true_labels
+        )
+        auroc = AUROC(task="multiclass", num_classes=n_classes).to(logits.device)(
+            logits, true_labels.view(-1).long()
+        )
```
adamgayoso, line 736: is there a way to avoid `.to`? In general this can be problematic on multi device training. I see it might be because you don't know n_classes a priori.

adamgayoso (changes requested):

scvi/train/_trainingplans.py
martinkim0: I think it should be possible by moving the tensors to CPU before logging and then the metric will be on the same device by default

adamgayoso: I though if you decalre the metrics in the init, lightning will handle all of this? Maybe we could detect n_classes in a different way

martinkim0: Makes sense. n_classes can be passed into the training plan init from scanvi

scvi/train/_trainingplans.py
```diff
@@ -1050,6 +1086,11 @@ def forward(self, *args, **kwargs):
         """Passthrough to the module's forward function."""
         return self.module(*args, **kwargs)
 
+    def compute_and_log_metrics(
+        self, loss_output: LossOutput, metrics: Dict[str, ElboMetric], mode: str
+    ):
+        assert loss_output.classification != {}
```
adamgayoso, line 1092: let's raise a proper error here

adamgayoso (changes requested):

scvi/module/base/_base_module.py
```diff
@@ -74,6 +74,7 @@ class LossOutput:
     reconstruction_loss: LossRecord | None = None
     kl_local: LossRecord | None = None
     kl_global: LossRecord | None = None
+    classification: dict[str, Tensor] | None = field(default_factory=dict)
```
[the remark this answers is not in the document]
martinkim0: Still not sure whether the keys in this argument should be strict or not. For now, I'm thinking that it should at least contain a classification loss, the ground truth labels, and either logits or predicted labels. logits are probably better since auroc does not accept predicted labels.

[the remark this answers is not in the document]
adamgayoso: my concern lies in the fact that all other arguments here are used in an explicit way, even `extra_metrics`, which just get logged. So I'm wondering if classification should also be a dataclass or more structured?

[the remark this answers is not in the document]
martinkim0: I see your point, since all other args do get logged, but this isn't necessarily the case for `classification`. In that case, I feel like separating it out into a new dataclass would make sense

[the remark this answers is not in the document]
adamgayoso: alternatively, add labels and logits here just like kl_local and global

martinkim0:
There's some repeated code here such as `log_with_mode`. I'm planning on a separate PR to refactor these more cleanly.

adamgayoso (approved):

PR #2025, opened 2023-05-05
Move `__repr__` to `BaseMinifiedModeModelClass`

### LINKED ISSUE #1781: Move latent mode init code to latent mode base model
https://github.com/scverse/scvi-tools/blob/0b802762869c43c9f49e69fe62b1a5a9b5c4dae6/scvi/model/base/_base_model.py#L68-L76

### LINKED ISSUE #1873: Base model repr has strange inheritance
Here it sort of depends on a child class, the base

https://github.com/scverse/scvi-tools/blob/c2c7996118b4ae6b91db6c04606809f2fa676542/scvi/model/base/_base_model.py#L735-L746

martinkim0:
Closes #1873

### Changes

Currently, regardless of whether the model class supports minification or not, the `__repr__` prints out the minification status of the model. This might be confusing for users - instead, we shouldn't print out that part of the model summary at all if it is not supported.


scvi/model/base/_base_model.py
```diff
@@ -834,6 +820,16 @@ def view_anndata_setup(
...
+    def __init__(self, adata: Optional[AnnOrMuData] = None):
+        data_is_minified = (
+            adata is not None and _get_adata_minify_type(adata) is not None
+        )
+        if data_is_minified and not issubclass(type(self), BaseMinifiedModeModelClass):
+            raise NotImplementedError(
+                f"The {type(self).__name__} model currently does not support minified data."
+            )
+        super().__init__(adata)
+
```
adamgayoso, line 832: If I recall, we needed this in the main __init__ in case a user passes a minified anndata to a non-supported model.

martinkim0: Ah I see. I'll revert those changes then and close the issue

martinkim0:
Do we want model reprs to print out the minified mode status regardless of whether they support the mode or not?

Issue #1682, opened 2022-09-15
Solo model KeyError when using continuous_covariate_keys


Nicholas-Everetts:
I'm having an issue creating a Solo model from an scVI model that was supplied continuous covariate data (percent mitochondrial reads, via "continuous_covariate_keys" parameter). The scVI model itself trains without an issue and looks reasonable, but it seems that the percent.mito data can't be found when trying to validate the AnnData object.

I haven't tested if this bug occurs when using categorical_covariate_keys.

```python
contin_keys = ["percent.mito"]
scvi.model.SCVI.setup_anndata(adata,
                              batch_key = "batch",
                              categorical_covariate_keys = None,
                              continuous_covariate_keys = contin_keys)
model = scvi.model.SCVI(adata,
                        n_latent = 20,
                        n_layers = 2,
                        gene_likelihood = "nb")
model.train(max_epochs = 400, train_size = 0.9)
print(adata.obs)
```
```pytb
#Output of the above print statement
                          batch  percent.mito  _scvi_batch  _scvi_labels
AAACCCACAAACAGGC-1        0.0      5.254359            0             0
AAACCCACAATTGCTG-1        0.0      3.539685            0             0
AAACCCACAGTTGCGC-1        0.0      5.130513            0             0
AAACCCAGTCCACTTC-1        0.0      6.404687            0             0
AAACGAAAGAGTTCGG-1        0.0      6.707605            0             0
...                       ...           ...          ...           ...
TTTGTTGGTATGGTAA-1        5.0      4.456618            5             0
TTTGTTGGTTCAAAGA-1        5.0      3.795428            5             0
TTTGTTGGTTGCACGC-1        5.0      4.157337            5             0
TTTGTTGTCATGAAAG-1        5.0      2.413516            5             0
TTTGTTGTCCGAGAAG-1        5.0      8.514687            5             0
```
```python
#Error occurs here
solo_model = scvi.external.SOLO.from_scvi_model(model, restrict_to_batch =  0)
print(solo_model)
```

```pytb
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Input In [13], in <cell line: 1>()
----> 1 solo_model = scvi.external.SOLO.from_scvi_model(model, restrict_to_batch =  0)
      2 print(solo_model)

[...]

PR #2027, opened 2023-05-08
Throw error when initializing SOLO from SCVI with extra covariates

martinkim0:
Addresses issue in #1682 with a more descriptive error message. Also fixes a case where SCVI is trained with batch labels but `restrict_to_batch` is not provided that errored out.


adamgayoso (approved):
just need a release note

PR #2032, opened 2023-05-12
Fix subsampling in SemiSupervisedDataLoader

martinkim0:


martinkim0:
Not working in its current state, for some reason the dataloader is only loading one observation.

martinkim0:
Current code on main has a bug when passing in `n_samples_per_label`

martinkim0:
[This](https://github.com/scverse/scvi-tools/files/11547154/scanvi_subsampling.pdf) notebook has a small comparison between the old and new subsampling code. The new version is achieving a lower final classification loss and higher accuracy and F1. Confusion matrix seems to be the same.

martinkim0:
[This](https://github.com/scverse/scvi-tools/files/11547322/scanvi_subsampling.pdf) benchmark shows that the new subsampling version leads to a lower final classification loss as well as higher accuracy and F1 scores.

martinkim0:
[Updated benchmark](https://github.com/scverse/scvi-tools/files/11559073/scanvi.pdf) with a model trained without subsampling

adamgayoso (approved):

PR #2035, opened 2023-05-15
Allow for custom samplers in dataloaders

### LINKED ISSUE #1995: Refactor AnnDataLoader to accept custom samplers
It looks like the way https://github.com/scverse/scvi-tools/blob/main/scvi/dataloaders/_ann_dataloader.py is written it'll always overwrite any custom samplers supplied to `**data_loader_kwargs` with a default torch sampler. For some models like CPA, it would help to have a sampler that takes into account cell labels since it implicitly assumes there are control cells in each batch. 

Would it be possible to only add the default sampler if the argument `sampler` isn't supplied to `**data_loader_kwargs`?

Thanks!
Yan

martinkim0:
Closes #1995

### DIFF
--- a/scvi/dataloaders/_ann_dataloader.py
+++ b/scvi/dataloaders/_ann_dataloader.py
@@ -6,7 +6,9 @@ import numpy as np
 from torch.utils.data import (
     BatchSampler,
     DataLoader,
+    Dataset,
     RandomSampler,
+    Sampler,
     SequentialSampler,
     Subset,
 )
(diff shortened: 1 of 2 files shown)


scvi/dataloaders/_ann_dataloader.py
```diff
@@ -24,38 +26,58 @@ class AnnDataLoader(DataLoader):
...
     data_loader_kwargs
-        Keyword arguments for :class:`~torch.utils.data.DataLoader`
+        Keyword arguments for :class:`~torch.utils.data.DataLoader`.
     """
 
     def __init__(
         self,
-        adata_manager: AnnDataManager,
+        adata_manager: Optional[AnnDataManager] = None,
+        dataset: Optional[Dataset] = None,
```
adamgayoso, line 57: let's not expose this

scvi/dataloaders/_ann_dataloader.py
```diff
@@ -66,21 +88,32 @@ def __init__(
...
+            sampler_cls = SequentialSampler if not shuffle else RandomSampler
+            sampler = BatchSampler(
+                sampler=sampler_cls(self.dataset),
+                batch_size=batch_size,
+                drop_last=drop_last,
+            )
+        if isinstance(sampler, BatchSampler):
+            batch_size = None  # disables PyTorch automatic batching
+            shuffle = False
+            drop_last = False
```
adamgayoso, line 102: what is going on here?

scvi/dataloaders/_ann_dataloader.py
```diff
@@ -66,21 +88,32 @@ def __init__(
...
+
+        if sampler is None:
+            sampler_cls = SequentialSampler if not shuffle else RandomSampler
+            sampler = BatchSampler(
+                sampler=sampler_cls(self.dataset),
+                batch_size=batch_size,
+                drop_last=drop_last,
+            )
+        if isinstance(sampler, BatchSampler):
+            batch_size = None  # disables PyTorch automatic batching
```
adamgayoso, line 100: this is important to add as a data loader kwarg when using the batchsampler

adamgayoso (changes requested):

scvi/dataloaders/_ann_dataloader.py
```diff
@@ -24,38 +26,58 @@ class AnnDataLoader(DataLoader):
...
+        `adata_manager` is not or if `sampler` is provided.
     shuffle
-        Whether the data should be shuffled
+        Whether the data should be shuffled.
     indices
-        The indices of the observations in the adata to load
+        The indices of the observations in the adata to load.
+    sampler
+        The sampler used for the dataloader. If `None`, a
+        :class:`~torch.utils.data.BatchSampler` will be used.
```
adamgayoso, line 40:
```suggestion
        :class:`~torch.utils.data.BatchSampler` will be used.
        This enables access anndata with a minibatch of indices,
        and makes data loading much faster for sparse arrays.
```

scvi/dataloaders/_ann_dataloader.py
adamgayoso: it should be forced as a data loader kwarg under our default sampler

scvi/dataloaders/_ann_dataloader.py
martinkim0: We need to force these kwargs if the custom sampler they are passing in is a batch sampler, otherwise torch throws an error

adamgayoso:
Close due to #2036

PR #2036, opened 2023-05-16
Allow manager to create torch dataset and expose option to use custom sampler

### LINKED ISSUE #1995: Refactor AnnDataLoader to accept custom samplers

adamgayoso:
Closes #1995


martinkim0 (approved):
LGTM

PR #2037, opened 2023-05-16
Add option to disable shuffling in datasplitter

martinkim0:


scvi/dataloaders/_data_splitting.py
```diff
@@ -72,6 +72,8 @@ class DataSplitter(pl.LightningDataModule):
         float, or None (default is 0.9)
     validation_size
         float, or None (default is None)
+    shuffle
+        Whether to shuffle indices before splitting.
```
adamgayoso, line 76:
```suggestion
        Whether to shuffle indices before splitting. If `False`, the val, train, and test set are split in the
        sequential order of the data according to `validation_size` and `train_size` percentages.
```

adamgayoso (approved):
LGTM after that one small change

scvi/dataloaders/_data_splitting.py
```diff
@@ -95,13 +98,15 @@ def __init__(
         adata_manager: AnnDataManager,
         train_size: float = 0.9,
         validation_size: Optional[float] = None,
+        shuffle: bool = True,
```
adamgayoso, line 101: actually can you expose this in unsupervisedtrainingmixn?

adamgayoso (changes requested):

scvi/dataloaders/_data_splitting.py
adamgayoso: and totalvi.train, peakvi.train, multivi.train, JaxTrainingMixin.train, destvi.train, condscvi.train,

adamgayoso: cellassign.train, scbasset.train, and then check solo and stereoscope

adamgayoso: also could rename to `shuffle_set_split` as `shuffle` typically refers to minibatching

adamgayoso (changes requested):
just need some tests then this LGTM

scvi/external/solo/_model.py
```diff
@@ -334,6 +338,7 @@ def train(
             self.adata_manager,
             train_size=train_size,
             validation_size=validation_size,
+            shuffle=shuffle_set_split,
```
martinkim0, line 341:
```suggestion
            shuffle_set_split=shuffle_set_split,
```

adamgayoso (approved):

PR #2060, opened 2023-05-24
Add `deregister_manager` function to base model class

### LINKED ISSUE #2053: Un-registering an anndata object from an scvi-tools BaseModelClass model
It seems like in order to create an scvi dataloader from an anndata object, that object must be "[registered](https://github.com/scverse/scvi-tools/blob/864afefa317ec6ae72fd5824d02d6bb3a19959c7/scvi/model/base/_base_model.py#L401)" with the model which then seems to create a [pointer](https://github.com/scverse/scvi-tools/blob/864afefa317ec6ae72fd5824d02d6bb3a19959c7/scvi/model/base/_base_model.py#L214) to that object in the model itself. This creates an issue for perturbational prediction models like CPA where you might want to run inference on many new datasets, which will then accumulate in memory due the pointers to all of these datasets in the model itself

Would it be possible to create an easy way to "unregister" an anndata object so that it doesn't persist in the model?

martinkim0:
Thank you for your contribution to scvi-tools. Feel free to modify or omit any
sections below based on your needs.

---

## General

-   [x] I have read the [contributing guide](https://docs.scvi-tools.org/en/stable/contributing/index.html).
-   [x] I have verified that there are no other pull requests open for the same fix or feature.

## Description

Closes #2053 

Adds `register_manager` calls wherever `_register_manager_for_instance` is called to ensure that the instance manager store is always a subset of the class manager store.

-   [x] I have linked the relevant issue(s) and/or included a description of the changes.
-   [x] If the changes fully address an issue, I have added `Fixes #<issue number>` to the PR description.

## Documentation

-   [x] I have added appropriate comments and documentation where necessary.
-   [x] I have added a description of the changes to the most recent entry in `docs/release_notes/index.md`.

## Testing

-   [x] I have added relevant tests for my changes.
-   [x] All new and existing tests pass locally.

## Development

### DIFF
--- a/scvi/model/base/_base_model.py
+++ b/scvi/model/base/_base_model.py
@@ -225,7 +225,7 @@ class BaseModelClass(TunableMixin, metaclass=BaseModelMetaClass):
         instance_manager_store[adata_id] = adata_manager
 
     def deregister_manager(self, adata_manager: Optional[AnnDataManager] = None):
-        """Deregisters an :class:`~scvi.data.AnnDataManager` instance with this model instance.
+        """Deregisters an :class:`~scvi.data.AnnDataManager` instance with this model instance and class.
 
         If `adata_manager` is `None`, deregisters all :class:`~scvi.data.AnnDataManager` instances
         in both the class and instance-specific manager stores, except for the current
@@ -235,16 +235,25 @@ class BaseModelClass(TunableMixin, metaclass=BaseModelMetaClass):
         instance_manager_store = self._per_instance_manager_store[self.id]
 
         if adata_manager is None:
-            manager_ids_to_clear = list(cls_manager_store.keys())
+            instance_managers_to_clear = list(instance_manager_store.keys())
+            cls_managers_to_clear = list(cls_manager_store.keys())
         else:
-            manager_ids_to_clear = [adata_manager.adata_uuid]
+            instance_managers_to_clear = [adata_manager.adata_uuid]
+            if adata_manager.adata_uuid in cls_manager_store:
+                cls_managers_to_clear = [adata_manager.adata_uuid]
+            else:
+                cls_managers_to_clear = []
 
-        for manager_id in manager_ids_to_clear:
+        for manager_id in instance_managers_to_clear:
             if adata_manager is None and manager_id == self.adata_manager.adata_uuid:
                 # don't clear the current manager by default
                 continue
-            if manager_id in instance_manager_store:
-                del instance_manager_store[manager_id]
+            del instance_manager_store[manager_id]
+
+        for manager_id in cls_managers_to_clear:
+            if adata_manager is None and manager_id == self.adata_manager.adata_uuid:
+                # don't clear the current manager by default
+                continue
             del cls_manager_store[manager_id]
 
     @classmethod
@@ -336,7 +345,6 @@ class BaseModelClass(TunableMixin, metaclass=BaseModelMetaClass):
             _assign_adata_uuid(adata, overwrite=True)
             adata_manager = self.adata_manager.transfer_fields(adata)
             self._register_manager_for_instance(adata_manager)
-            self.register_manager(adata_manager)
 
         return adata_manager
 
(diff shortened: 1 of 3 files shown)


scvi/model/base/_base_model.py
```diff
@@ -224,6 +224,29 @@ def _register_manager_for_instance(self, adata_manager: AnnDataManager):
...
+            manager_ids_to_clear = [adata_manager.adata_uuid]
+
+        for manager_id in manager_ids_to_clear:
+            if adata_manager is None and manager_id == self.adata_manager.adata_uuid:
+                # don't clear the current manager by default
+                continue
+            if manager_id in instance_manager_store:
+                del instance_manager_store[manager_id]
+            del cls_manager_store[manager_id]
+
```
adamgayoso, line 249: this doesn't capture the case where manager is in the instance but not the cls manager store. 

This arises commonly if a new anndata is passed to a `get...` method.

Therefore we should independently clear the cls and instance stores but always skip over the current.

adamgayoso (changes requested):

scvi/model/base/_base_model.py
martinkim0: The current changes add calls to `register_manager` wherever the instance manager store is updated, such as in `get_anndata_manager`. This ensures that both manager stores are in sync and that the instance is always a subset of the class one. I guess my question now would be whether this is necessary or something we want to do at all.

adamgayoso: hmmm I'm not sure if we should add this change here. We can discuss with @justjhong later but I think just for this PR we should deregister without changing how registration works. Wdyt?

martinkim0: Sounds good, I'll revert the changes then

adamgayoso: this would be easier to read if there were two separate for loops, one for the cls manager store and one for the instance manager store

scvi/model/base/_base_model.py
```diff
@@ -413,9 +437,9 @@ def _validate_anndata(
                 "Input AnnData not setup with scvi-tools. "
                 + "attempting to transfer AnnData setup"
             )
-            self._register_manager_for_instance(
-                self.adata_manager.transfer_fields(adata)
-            )
+            adata_manager = self.adata_manager.transfer_fields(adata)
+            self._register_manager_for_instance(adata_manager)
+            self.register_manager(adata_manager)
```
adamgayoso, line 442: no need for these changes

adamgayoso (changes requested):

scvi/model/base/_base_model.py
```diff
@@ -224,6 +224,39 @@ def _register_manager_for_instance(self, adata_manager: AnnDataManager):
         instance_manager_store = self._per_instance_manager_store[self.id]
         instance_manager_store[adata_id] = adata_manager
 
+    def deregister_manager(self, adata_manager: Optional[AnnDataManager] = None):
```
justjhong, line 227: should this take the manager or the anndata, or both? I feel like most users aren't aware of how to get the manager object, and you could do with just the Anndata object by checking and fetching it's uuid

scvi/model/base/_base_model.py
```diff
@@ -224,6 +224,39 @@ def _register_manager_for_instance(self, adata_manager: AnnDataManager):
...
+        instance_manager_store = self._per_instance_manager_store[self.id]
+
+        if adata_manager is None:
+            instance_managers_to_clear = list(instance_manager_store.keys())
+            cls_managers_to_clear = list(cls_manager_store.keys())
+        else:
+            cls_managers_to_clear = [adata_manager.adata_uuid]
+            instance_managers_to_clear = [adata_manager.adata_uuid]
+
+        for manager_id in cls_managers_to_clear:
```
justjhong, line 244: call this adata_id still because manager.id is a separate thing

justjhong (commented):
thanks for doing this, left a couple comments!

justjhong (commented):
thanks for doing this, left a couple comments!

scvi/model/base/_base_model.py
martinkim0: hmm yeah since this is user-facing, I feel like this should just take the anndata

scvi/model/base/_base_model.py
martinkim0: makes sense

justjhong (approved):
lgtm!

adamgayoso (approved):

PR #2158, opened 2023-07-05
Add experimental sparse tensor functionality

martinkim0:
-   [ ] Closes #xxxx (Replace xxxx with the GitHub issue number)
-   [ ] Tests added and passed if fixing a bug or adding a new feature
-   [ ] All code checks passed.
-   [ ] Added type annotations to new arguments/methods/functions.
-   [ ] Added an entry in the latest `docs/release_notes/index.md` file if fixing a bug or adding a new feature.
-   [ ] If the changes are patches for a version, I have added the `on-merge: backport to x.x.x` label.

### DIFF
--- a/scvi/data/_anntorchdataset.py
+++ b/scvi/data/_anntorchdataset.py
@@ -6,6 +6,7 @@ import numpy as np
 import pandas as pd
 from anndata._core.sparse_dataset import SparseDataset
 from scipy.sparse import issparse
+from torch import as_tensor, sparse_csr_tensor
 from torch.utils.data import Dataset
 
 from scvi._constants import REGISTRY_KEYS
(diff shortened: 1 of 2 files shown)


martinkim0:
@valentinwust This PR contains an initial implementation of your idea. I can tag you as co-author when it merges if that's ok with you!

valentinwust:
> @valentinwust This PR contains an initial implementation of your idea. I can tag you as co-author when it merges if that's ok with you!

Sounds good, the code looks fine. You are right that it only works for csr, but this is anyway the only sensible choice if you care about speed. Currently using something else anyway throws a warning, I don't really see the use case so it might be worth adding a forced conversion for the people who ignore warnings.

martinkim0:
`AnnData` currently only supports csr and csc as sparse formats, so I think it's sensible to assume this in `AnnTorchDataset` and not force conversion.

valentinwust:
Fair enough. The interfaces seem nicely consistent, so something like this should be sufficient to deal with both csr and csc properly:
```python
to_sparse_tensor = sparse_csr_tensor if scipy.sparse.isspmatrix_csr(sliced_sparse) else sparse_csc_tensor
sliced_data = to_sparse_tensor(
                      as_tensor(sliced_sparse.indptr),
                      as_tensor(sliced_sparse.indices),
                      as_tensor(sliced_sparse.data),
                      size=sliced_sparse.shape,
                  )
```

PR #2224, opened 2023-08-02
Add `use_posterior_mean` option to scanvi `predict`

### LINKED ISSUE #1587: Stochastic/Non-deterministic prediction for scANVI
I am trying to run a procedure called self projection with a scANVI model. The procedure entails taking a trained scANVI model and repeatedly generating predictions on the dataset used to train it, hence the name 'self projection.' Because scANVI employs a probabilistic framework and not a deterministic one, I would think that each time scANVI generates predictions on some dataset (i.e. scANVI.predict()), the predictions would be slightly different each time. However, when I repeatedly use scANVI.predict(), it generates the same predictions each time.

I looked through the documentation for scANVI, but I do not see any option/parameter for .predict() function that allows this feature. I tried changing the seed but no change. Having such an option would be extremely helpful! Please let me know if there is a current work-around.

Thank you.

martinkim0:
-   [x] Closes #1587 (Replace xxxx with the GitHub issue number)

### Summary by CodeRabbit

**New Feature**
- Added a new argument `use_posterior_mean` to the `predict` method in the `SCANVI` class. This allows for stochastic prediction of cell type labels.

**Bug Fix**
- Replaced `torch.log` with `torch.log1p` in `_scanvae.py` to avoid numerical instability.

**Test**
- Introduced a new test function `test_scanvi_predict_use_posterior_mean()` to validate the functionality of the new feature.

> 🎉 Here's to the code that's ever so lean,  
> With `use_posterior_mean`, predictions are keen.  
> No more instability, thanks to `log1p`,  
> Our tests ensure we don't miss a blip! 🚀

### DIFF
--- a/docs/release_notes/index.md
+++ b/docs/release_notes/index.md
@@ -30,6 +30,8 @@ is available in the [commit logs](https://github.com/YosefLab/scvi-tools/commits
     and {class}`scvi.external.GIMVI.save` {pr}`2200`.
 -   Add `model_kwargs` and `train_kwargs` arguments to {meth}`scvi.autotune.ModelTuner.fit` {pr}`2203`.
 -   Add `datasplitter_kwargs` to model `train` methods {pr}`2204`.
+-   Add `use_posterior_mean` argument to {meth}`scvi.model.SCANVI.predict` for
+    stochastic prediction of celltype labels {pr}`2224`.
 
 #### Changed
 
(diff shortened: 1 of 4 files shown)


cane11:
The code change makes sense and I was also puzzled by this decision in the past. However, are there any tasks where the non-default value has shown added value?

PR #2240, opened 2023-08-15
Implement batch embeddings for `SCVI`

martinkim0:
-   [x] Closes #xxxx (Replace xxxx with the GitHub issue number)

### DIFF
--- a/scvi/model/base/_vaemixin.py
+++ b/scvi/model/base/_vaemixin.py
@@ -210,3 +210,15 @@ class VAEMixin:
             if return_dist
             else torch.cat(latent).numpy()
         )
+
+    @torch.inference_mode()
+    def get_batch_representation(self) -> np.ndarray:
+        """Return the batch representations."""
+        self._check_if_trained(warn=False)
+
+        if not hasattr(self.module, "batch_embedding"):
+            raise NotImplementedError(
+                "Module does not have attribute `batch_embedding`."
+            )
+
+        return self.module.batch_embedding.weight.detach().cpu().numpy()
--- a/scvi/model/base/_vaemixin.py
+++ b/scvi/model/base/_vaemixin.py
@@ -217,16 +217,15 @@ class VAEMixin:
     @torch.inference_mode()
     def get_batch_representation(
         self,
-        adata: AnnData | None = None,
         batch_keys: list[str] | None = None,
     ) -> np.ndarray:
         """Return batch representations.
 
+        The model be initialized with ``use_batch_embeddings=True`` and trained
+        in order to use this method.
+
         Parameters
         ----------
-        adata
-            AnnData object with equivalent structure to initial AnnData. If ``None``, defaults to
-            the AnnData object used to initialize the model.
         batch_keys
             Keys for batches to return representations for. If ``None``, all batches are used.
 
@@ -236,13 +235,15 @@ class VAEMixin:
         """
         self._check_if_trained(warn=False)
 
-        if not hasattr(self.module, "batch_embedding"):
-            raise NotImplementedError("Model does not support batch embeddings.")
-        elif self.module.batch_embedding is None:
-            raise ValueError("Model was not trained with batch embeddings.")
+        if (
+            not hasattr(self.module, "batch_embedding")
+            or self.module.batch_embedding is None
+        ):
+            raise NotImplementedError(
+                "Model does not support or was not trained with batch embeddings."
+            )
 
-        adata = self._validate_anndata(adata)
-        manager = self.get_anndata_manager(adata)
+        manager = self.get_anndata_manager(self.adata)
         batch_state_registry = manager.get_state_registry(REGISTRY_KEYS.BATCH_KEY)
         cat_mapping = batch_state_registry.categorical_mapping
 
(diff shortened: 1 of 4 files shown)


scvi/module/_vae.py
```diff
@@ -166,6 +173,18 @@ def __init__(
...
+        if use_batch_embedding:
+            if n_batch < 2:
+                raise ValueError(
+                    "`adata` must have more than 1 batch when passing in "
+                    "`use_batch_embedding=True`."
+                )
+
+            self.batch_embedding = nn.Embedding(n_batch, batch_embedding_dim)
+            # fclayers expects size of batch input == n_batch
+            n_batch = batch_embedding_dim
```
canergen, line 186: Does the order make sense here?

scvi/model/base/_vaemixin.py
```diff
@@ -210,3 +213,45 @@ def get_latent_representation(
...
+        Array of batch representations of shape ``(n_batches, batch_embedding_dim)``.
+        """
+        self._check_if_trained(warn=False)
+
+        if not hasattr(self.module, "batch_embedding"):
+            raise NotImplementedError("Model does not support batch embeddings.")
+        if self.module.batch_embedding is None:
+            raise ValueError("Model was not trained with batch embeddings.")
+
+        adata = self._validate_anndata(adata)
```
canergen, line 244: Add a warning if batch_key has similar name but actually describes different batches the result will be non-sense.

scvi/model/base/_vaemixin.py
```diff
@@ -210,3 +213,45 @@ def get_latent_representation(
...
+        cat_mapping = batch_state_registry.categorical_mapping
+
+        if batch_keys is None:
+            batch_indexes = np.arange(len(cat_mapping))
+        elif not all((key in cat_mapping) for key in batch_keys):
+            raise ValueError("``batch_keys`` contains keys not present in ``adata``.")
+        else:
+            batch_indexes = np.where(cat_mapping == batch_keys)[0]
+
+        batch_embeddings = self.module.batch_embedding.weight.detach().cpu().numpy()
```
canergen, line 256: Potentially leave this on GPU. Run embedding and retrieve the embeddings. This object can be rather large if you have a large number of batches.

scvi/module/_vae.py
canergen, line 186: Currently, categorical_covariates are not batch_embedded. I guess it's fine but depending on the use case it might be useful.

scvi/module/_vae.py
```diff
@@ -279,6 +302,10 @@ def _get_generative_input(self, tensors, inference_outputs):
             else None
         )
 
+        if self.batch_embedding is not None:
```
canergen, line 305: Out of curiosity: Is the embedding optimized to be good for the encoder or the decoder?

canergen (commented):
We should add a tutorial. Code changes LGTM. Have you done some benchmarking of it?

scvi/model/base/_vaemixin.py
martinkim0: Thanks for taking a look! I'm not following, could you clarify what you mean?

scvi/model/base/_vaemixin.py
martinkim0: Yeah it could potentially be large depending on the number of batches. Two reasons why I support moving it to CPU: (1) it would be more consistent with `get_latent_representation`, which returns numpy arrays, (2) and by default the batch embeddings are guaranteed to be smaller than the latent representation since `batch_embedding_dim=5`.

scvi/model/base/_vaemixin.py
canergen: If a user provides a new adata object with different batches, in which batches are e.g. just numbered, it will output batch embeddings for those that have no meaning. I think it's important that we allow using a new adata object but we could issue a warning.

scvi/module/_vae.py
martinkim0: Good point! Do you think supporting other categorical covariates would be an interesting use case? In terms of the implementation, it would make sense if we supported embedding other covariates, we would just have to think a bit more to make the API non-redundant.

scvi/model/base/_vaemixin.py
canergen: We are not copying a 5-dimensional batch_embedding but a n_batchxbatch_embedding_dim dimensional tensor here. However, copying batch_indices to GPU will not reduce the amount that needs to be transferred.

scvi/model/base/_vaemixin.py
canergen: I guess the same thing would happen without batch_embeddings as well. Marking it as resolved.

canergen (approved):
We should add a tutorial and see that the results are fine there. Code LGTM.

scvi/model/base/_vaemixin.py
martinkim0: Ah I see what you mean now. I have to think more about this.

scvi/module/_vae.py
martinkim0: The embeddings are by default fed into the decoder only

scvi/module/_vae.py
canergen: I think batch_key and categorical_covariate in the same low-dimensional batch_embedding are fine (simply adding up both embeddings). It would be good to keep the names separate though. I have to think more about the use case. It requires cases with a large number of categorical covariate values and I haven't used those cases.

scvi/model/base/_vaemixin.py
martinkim0: I feel like the simplest solution to this would be removing the `adata` argument in `get_batch_representation`.

PR #2249, opened 2023-08-23
PoissonVI to model quantitative scATAC-seq fragment counts

### LINKED ISSUE #1926: Implement PoissonVI to model scATAC-seq fragment counts

lauradmartens:
-   [x] Closes #1926 
      - Where should I add the test for the Layer field?
      -  Fails with Python 3.9 because of typing error

## New things added with PoissonVI:
- New LayerField `FragmentCountLayerField` that checks that the provided matrix does not contain binarized scATAC counts or read counts (which normally have more even than uneven counts)
- PoissonVI model that models fragment counts with a Poisson distribution. This used the default `VAE` model but differs from scvi because it: 
     - uses the default PeakVI architecture and training procedure
     - Implements the functions `get_accessibility_estimates` and `differential_accessibility`
     - In `get_accessibility_estimates` one has the option to compute normalized accessibility estimates without the region specific bias which is essentially the bias parameter of the last layer of the model (similar to PeakVi's region parameter)


scvi/external/poissonvi/_model.py
```diff
@@ -0,0 +1,541 @@
+import logging
```
martinkim0, line 1:
```suggestion
from __future__ import annotations

import logging
```
I think this will fix the typing error in Python 3.9. I'll take a look at the rest of the code later. Thanks!

lauradmartens: Thanks!

scvi/data/_utils.py
```diff
@@ -317,3 +317,37 @@ def _is_minified(adata: Union[AnnData, str]) -> bool:
...
+        non_binary, non_reads = _is_not_fragment_val(check)
+        ret = not (non_reads or non_binary)
+    return ret
+
+
+@jax.jit
+def _is_not_fragment_val(data: jnp.ndarray):
+    non_binary = jnp.logical_not(jnp.any(data > 1))
+    non_reads = jnp.count_nonzero(data == 1) > jnp.count_nonzero(data == 2)
+    return non_binary, non_reads
```
martinkim0, line 353: I think it would be preferable to use pure numpy here instead of jax because, for most use cases, this function will only be called once. Jax jit is generally not worth it unless the function is called multiple times.

scvi/data/_utils.py
```diff
@@ -317,3 +317,37 @@ def _is_minified(adata: Union[AnnData, str]) -> bool:
...
+    n_to_check: int = 100,
+):
+    """Approximately checks values of data to ensure it is fragment count data."""
+    # for backed anndata
+    if isinstance(data, h5py.Dataset) or isinstance(data, SparseDataset):
+        data = data[:100]
+
+    if isinstance(data, np.ndarray):
+        data = data
+    elif issubclass(type(data), sp_sparse.spmatrix):
```
martinkim0, line 333: Cleaner to use `scipy.sparse.issparse`

scvi/data/_utils.py
```diff
@@ -317,3 +317,37 @@ def _is_minified(adata: Union[AnnData, str]) -> bool:
             return uns_key in read_elem(fp["uns"]).keys()
     else:
         raise TypeError(f"Unsupported type: {type(adata)}")
+
+
+def _check_fragment_counts(
+    data: Union[pd.DataFrame, np.ndarray, sp_sparse.spmatrix, h5py.Dataset],
+    n_to_check: int = 100,
```
martinkim0, line 324: Need to check for the unlikely edge case where `adata.n_obs < n_to_check`.

scvi/data/_utils.py
```diff
@@ -317,3 +317,37 @@ def _is_minified(adata: Union[AnnData, str]) -> bool:
...
+):
+    """Approximately checks values of data to ensure it is fragment count data."""
+    # for backed anndata
+    if isinstance(data, h5py.Dataset) or isinstance(data, SparseDataset):
+        data = data[:100]
```
martinkim0, line 329: Same here

scvi/data/_utils.py
```diff
@@ -317,3 +317,37 @@ def _is_minified(adata: Union[AnnData, str]) -> bool:
...
+
+    if isinstance(data, np.ndarray):
+        data = data
+    elif issubclass(type(data), sp_sparse.spmatrix):
+        data = data.data
+    elif isinstance(data, pd.DataFrame):
+        data = data.to_numpy()
```
martinkim0, line 336: This line could potentially be copying the entire dataframe into a numpy. In terms of performance, it would be preferable to subset the dataframe first and then convert to numpy.

scvi/data/_utils.py
martinkim0: I'd also just refactor so that the logic in `_is_not_fragment_val` is inside `_check_fragment_counts`

scvi/data/fields/_layer_field.py
```diff
@@ -133,3 +134,51 @@ def view_state_registry(self, _state_registry: dict) -> Optional[rich.table.Tabl
...
+                "adata.X" if self.attr_key is None else f"adata.layers[{self.attr_key}]"
+            )
+            warnings.warn(
+                f"{logger_data_loc} does not contain fragment count data. "
+                "Are you sure this is what you want?. "
+                "Check that your data is not binarized and does not contain read counts. "
+                "You can approximate read counts to fragment counts using scvi.data.reads_to_fragments",
+                UserWarning,
+                stacklevel=settings.warnings_stacklevel,
+            )
```
martinkim0, line 184: Since this new class only overwrites `validate_field`, it would be cleaner to instead include an argument in `LayerField.__init__` (_e.g._ `is_fragment_count_data`) and then add the appropriate check into `LayerField.validate_field`

tests/external/test_poissonvi.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def _get_adata(sparse_format: Optional[str] = None):
+    dataset1 = synthetic_iid(batch_size=100, sparse_format=sparse_format)
+    return dataset1
+
+
+def test_poissonvi():
+    adata = _get_adata()
+    print(adata.obs)
```
martinkim0, line 14:
```suggestion
```

tests/external/test_poissonvi.py
```diff
@@ -0,0 +1,21 @@
...
+from scvi.external import POISSONVI
+
+
+def _get_adata(sparse_format: Optional[str] = None):
+    dataset1 = synthetic_iid(batch_size=100, sparse_format=sparse_format)
+    return dataset1
+
+
+def test_poissonvi():
+    adata = _get_adata()
```
martinkim0, line 13:
```suggestion
    adata = synthetic_iid(batch_size=100)
```

tests/external/test_poissonvi.py
```diff
@@ -0,0 +1,21 @@
...
+
+from scvi.data import synthetic_iid
+from scvi.external import POISSONVI
+
+
+def _get_adata(sparse_format: Optional[str] = None):
+    dataset1 = synthetic_iid(batch_size=100, sparse_format=sparse_format)
+    return dataset1
+
+
```
martinkim0, line 11:
```suggestion
```

martinkim0 (commented):
Thanks for the contribution! One comment that I wasn't able to include: since `POISSONVI` needs to reproduce the default parameters of `PEAKVI`, it would be preferable if the latter is a subclass of the former so there is less repeated code. And then you could add a parameter `likelihood` to the model init that is defaulted to `"poisson"`.

lauradmartens:
Thanks a lot for the comments! I addressed the minor changes on the `_check_fragment_counts` and will rewrite the model as a subclass of `PEAKVI` by the end of the week.

lauradmartens:
Thanks again for your comments! I thought again of the suggestions of making `PEAKVI` a subclass of `POISSONVI` to have less repeated code for the model architecture. If I understood it correctly, you would call `PEAKVI` init with an additional likelihood parameter set to `"poisson"`? I am unsure how to make this work with `VAE` as the module class, because `PEAKVI/PEAKVAE` has many arguments that are not in `VAE` (`n_input_regions`, `region_factors`, `model_depth`).

martinkim0:
Hi @lauradmartens, sorry for the late reply. My understanding is that `POISSONVI` uses the default model architecture of `PEAKVI`, so maybe it makes more sense to use `PEAKVAE` as the module class instead of `VAE`? This would make it easier to subclass from `PEAKVI` instead. Let me know what you think!

lauradmartens:
Hi @martinkim0! I think it's actually easier to use the `VAE` model because it implements all the functions to deal with Poisson/negative binomial loss whereas the `PEAKVAE` model has the binary decoder and the additional region and cell-specific factors. When I said the same architecture, I just meant the same number of layers, etc. to deal with the much larger feature space in scATAC-seq compared to scRNA-seq. I now wrote it in a way that `POISSONVI` inherits from `PEAKVI`, which reduces the doubled code at least for the training. Let me know what you think :)

scvi/external/poissonvi/_model.py
```diff
@@ -0,0 +1,441 @@
...
+
+    Notes
+    -----
+    See further usage examples in the following tutorials:
+
+    1. :doc:``
+    """
+
+    # TODO: change tutorial link
+    # TODO: change citation
```
martinkim0, line 77: Needs to be addressed

scvi/external/poissonvi/_model.py
```diff
@@ -0,0 +1,441 @@
...
+                region_factors
+            )
+        return accs
+
+    def get_normalized_expression(
+        self,
+    ):
+        # Refer to function get_accessibility_estimates
+        msg = f"differential_expression is not implemented for {self.__class__.__name__}, please use {self.__class__.__name__}.get_accessibility_estimates"
+        raise ValueError(msg)
```
martinkim0, line 246: Preferable to raise a `NotImplementedError` here

scvi/data/fields/_layer_field.py
martinkim0, line 184: Taking a second look at this, I think it would be cleaner if we're able to directly modify `LayerField` instead. In other words, we can add an argument `check_fragment_counts` to `LayerField.__init__` and then have 
```
if self.check_fragment_counts and not _check_fragment_counts(x):
    ....
```
in `LayerField.validate_field`. What do you think?

scvi/external/poissonvi/_model.py
```diff
@@ -0,0 +1,441 @@
...
+            }
+        )
+        return result
+
+    def differential_expression(
+        self,
+    ):
+        # Refer to function differential_accessibility
+        msg = f"differential_expression is not implemented for {self.__class__.__name__}, please use {self.__class__.__name__}.differential_accessibility"
+        raise ValueError(msg)
```
martinkim0, line 393: `NotImplementedError` here too

martinkim0 (commented):
Looks great overall! Left some minor comments and it should be good to merge after that. Let me know if you're also interested in contributing a tutorial

lauradmartens:
Hi Martin! I added the changes as you suggested and used a provisional link for the tutorial. I am of course happy to also contribute a tutorial. Would you also like me to do a `user_guide` section of what is the difference?

martinkim0:
Thanks for addressing the comments! The user guide is optional - it should just be a condensed explanation of how the model works. Since we already have a PeakVI [one](https://docs.scvi-tools.org/en/stable/user_guide/models/peakvi.html), you could just copy paste and make the necessary changes. For tutorials, you can put in a PR in our [tutorials repo](https://github.com/scverse/scvi-tutorials).

martinkim0 (approved):
LGTM

martinkim0:
Hi @lauradmartens, what's the status on this? Are there any changes you wanted to add before merging?

lauradmartens:
No it looks good to me. I will add the tutorial by the end of the week.

martinkim0:
Hey @lauradmartens! Thanks for your work so far - I'm planning on releasing this with scvi-tools 1.1.0 within a couple of weeks. I will probably post a Twitter thread with release updates. I was wondering if you'd like me to tag you in the thread (if so, please reply with your handle) and/or if you'd like to write the tweet describing the model (if so, please reply with the summary). Let me know!

lauradmartens:
Thanks @martinkim0! My twitter handle is `lauradmartens`.

For the description maybe something like:
Modeling quantitative scATAC-seq fragment counts is now possible using the [PoissonVI model](https://www.nature.com/articles/s41592-023-02112-6). Also check out the tutorial on how you can convert your 10x matrices to fragment count matrices.

PR #2297, opened 2023-10-23
Fix a bug in the imputation process of Gimvi

HelloWorldLTY:
Hi, this is my current solution for this bug:

https://github.com/scverse/scvi-tools/issues/1634

I think we need to add a key about batch_size in the model.py part  to ensure that the user can adjust the batch_size they used for inference. Moreover, we also need a warning if one modality has data size which will lead to one example remining problem.


HelloWorldLTY:
Any feedback?

martinkim0:
Hi, thank you for putting in this PR. Regarding adding `batch_size`, I believe we already expose that argument in both the `train` method as well as all inference method. Were you referring to a different method? 

I agree that it's probably best to include some sort of warning if there ends up being a single observation in a batch since this leads to issues with batch norm. However, it's not clear to me why this is occurring during inference. Could you provide a reproducible example that leads to this error? Thank you.

kamilkrukowski:
There's code dependent on ```_make_scvi_dls``` that expects the ```batch_size``` parameter to do something instead of nothing.

```get_latent_representation``` and ```get_imputed_values```  iterate over the data loaders created by ```_make_scvi_dls``` and the batch_size parameter of ```get_latent_representation``` is only passed to ```_make_scvi_dls``` and otherwise unused.

```
344        imputed_values = []
345        for mode, scdl in enumerate(scdls):
346            imputed_value = []
```

Thus, currently the parameter ```batch_size``` is completely ignored for ```get_latent_representation```, which seems like unintentional buggy behavior resolved by the PR.

About the "batch_size of 1" issue --> If we have 129 rows and the dataloader doesn't drop-incomplete minibatches then a batch_size of 128 would have mini batches of size [128, 1] and cause issues with batch norm.

martinkim0:
Thanks, I didn't catch that `batch_size` was not being passed into `_make_scvi_dls`. Regarding the second point, I'm aware of this but afaik it should only be an issue during training (and not inference). I'll see if I can reproduce the error.

HelloWorldLTY:
Hi, thanks for your catch up. Happy to offer more help if you need.

martinkim0:
Closing in favor of #2366

PR #2362, opened 2023-12-08
Fixes error in get losses functions

canergen:
-   [ ] Closes #xxxx (Replace xxxx with the GitHub issue number)
-   [X] Tests added and passed if fixing a bug or adding a new feature
-   [X] All code checks passed
-   [x] Added type annotations to new arguments/methods/functions
-   [x] Added an entry in the latest `docs/release_notes/index.md` file if fixing a bug or adding a new feature
-   [x] If the changes are patches for a version, I have added the `on-merge: backport to x.x.x` label


scvi/module/_vae.py
```diff
@@ -609,6 +609,8 @@ def marginal_ll(
                 q_l_x = ql.log_prob(library).sum(dim=-1)
 
                 log_prob_sum += p_l - q_l_x
+            if n_mc_samples_per_pass == 1:
+                log_prob_sum = log_prob_sum.unsqueeze(0)
```
martinkim0, line 613: Wait so does this mean that this method was not working properly before? Since the default is `n_mc_samples_per_pass=1`

tests/model/test_scanvi.py
```diff
@@ -237,6 +237,17 @@ def test_multiple_covariates_scanvi():
     m.get_latent_representation()
     m.get_elbo()
     m.get_marginal_ll(n_mc_samples=3)
+    # m.get_marginal_ll(adata, return_mean=True, n_mc_samples=6, n_mc_samples_per_pass=1)
```
martinkim0, line 240: Uncomment this line?

scvi/model/base/_log_likelihood.py
```diff
@@ -13,22 +14,34 @@ def compute_elbo(vae, data_loader, feed_labels=True, **kwargs):
...
+
+    kl_global = np.sum([np.array(i) for i in scvi_loss.kl_global.values()], axis=0)
     n_samples = len(data_loader.indices)
-    elbo += kl_global
-    return elbo / n_samples
+    if return_mean:
+        elbo += kl_global
+        return elbo / n_samples
+    else:
+        return elbo + kl_global / n_samples
```
martinkim0, line 40: Probably best to also use torch operations here since `compute_reconstruction_error` doesn't use numpy. Let me know what you think

canergen:
Yes but we only use it for DE genes where it's called with n_samples_per_mc of 100 (?). ScANVI wasn't supporting importance weighting beforehand. It doesn't work with multiple samples and this might be an issue for multi-GPU support (see bug report).

scvi/module/_vae.py
canergen: Yes but we only use it for DE genes where it's called with n_samples_per_mc of 100 (?). ScANVI wasn't supporting importance weighting in DE beforehand.
ScANVI doesn't work with multiple samples (major work to fix this) and this might be an issue for multi-GPU support (see bug report). If you want to fix this, I have a fix for broadcast labels but then dropped it as also the encoder doesn't support it.

scvi/model/base/_log_likelihood.py
canergen: Fine with me. Can you puh those changes?
```suggestion
        return elbo + kl_global / n_samples
```

martinkim0:
I'll get this in by today @canergen

martinkim0:
#2429 needs to be addressed before this

canergen:
@ori-kron-wis We can merge this one after fixing the n_samples issue.

PR #2373, opened 2023-12-18
Add seed argument for pymde

### LINKED ISSUE #2372: PyMDE seed fixation suggestion
Hello,

I've been trying out different dimension reduction methods mostly for visualization purpose.
scvi.model.utils.mde seems to be a very interesting and useful feature added in the ecosystem, but I realized while I was playing around few parameters like neighbors and repulsive_fraction that the random seed is not fixed nor it could be fixed with a global seed in my script.
Could we just add a quick feature like scvi.model.utils.mde(seed=42) and add the pymde.seed(seed) calling right around there?

in _mde.py
 ```
    try:
         import pymde
         pymde.seed(seed)
```

Thank you for all your work!
Seyoon

martinkim0:
Closes #2372


martinkim0:
meeseeksdev backport to 1.1.x

PR #2377, opened 2024-01-04
Clean up posterior predictive sample methods

martinkim0:
- Remove unused `library_size` argument (could add this back later once it is implemented) in `VAE.sample`
- Add typing and additional documentation in `VAE.sample`
- Clean up tensor shape annotations in `VAE.sample` and `RNASeqMixin.posterior_predictive_sample`
- Use Python 3.9+ annotations in `scvi.model.base._rnamixin`
- Clean up code in `RNASeqMixin.posterior_predictive_sample`


martinkim0:
meeseeksdev backport to 1.1.x

PR #2379, opened 2024-01-05
Can hub paper

canergen:
Adds minified mode to CondSCVI. Simplifies minification for new models. Adds add_latent_posterior mode which keeps count data. Fixes glitches in DE function.
-   [X] Tests added and passed if fixing a bug or adding a new feature
-   [X] All code checks passed
-   [X] Added type annotations to new arguments/methods/functions
-   [ ] Added an entry in the latest `docs/release_notes/index.md` file if fixing a bug or adding a new feature


martinkim0:
If possible, could you separate this out into 4 PRs:

- Minified mode for CondSCVI
- Simplification of minification for new models
- `add_latent_posterior` mode
- Fixes in differential expression

Thanks

scvi/model/base/_base_model.py
```diff
@@ -481,6 +490,103 @@ def _check_if_trained(
             else:
                 raise RuntimeError(message)
 
+    def minify_adata(
```
martinkim0, line 493: If the plan is to put this method under `BaseModelClass`, we should also move over the other methods under `BaseMinifiedModeModelClass` and then delete it. What do you think?

scvi/model/utils/_minification.py
```diff
@@ -21,12 +22,18 @@ def get_minified_adata_scrna(
...
+    if not ADATA_MINIFY_TYPE.__contains__(minified_data_type):
         raise NotImplementedError(f"Unknown MinifiedDataType: {minified_data_type}")
 
-    all_zeros = csr_matrix(adata.X.shape)
-    layers = {layer: all_zeros.copy() for layer in adata.layers}
+    if not keep_count_data:
+        all_zeros = csr_matrix(adata.X.shape)
+        layers = {layer: all_zeros.copy() for layer in adata.layers}
+    else:
+        all_zeros = adata.X
```
martinkim0, line 35: Thoughts on checking whether `adata.X` is a sparse matrix and if not, sparsify it (in the spirit of minification)?

scvi/model/utils/_minification.py
```diff
@@ -21,12 +22,18 @@ def get_minified_adata_scrna(
...
+        layers = adata.layers
```
martinkim0, line 36: It may also make sense to only copy the count layer registered with the manager so we don't have to carry around the unused layers as part of the minified data.

martinkim0 (commented):
Thanks, LGTM overall! Just one major comment about switching the new minified mode to a new argument and some minor comments about performance/storage.

Since there's a lot of conflicts w/ the main branch, I can take care of transferring this to another PR and tagging you as co-author.

PR #2467, opened 2024-02-05
Add API for custom dataloaders in `SCVI`

martinkim0:
This commit enables users to bypass the default workflow of setting up an `AnnData` and initializing a model with it in order to train, and instead being able to train a model directly with a custom dataloader.

- Make `adata` argument optional in `SCVI`, which sets a `_module_init_on_train` flag (through `BaseModelClass`) if no `adata` is passed in
- This causes the module init to be delayed until train time since we won't have access to certain arguments (*e.g.* `n_input` and `n_batch`) that must be inferred from the data
- Add an optional `data_module` argument to `UnsupervisedTrainingMixin.train` that must be a `LightningDataModule` and must be passed in if the model was not initialized with `adata`

Set as an experimental feature for now. Will expand to models other than `SCVI` when the implementation and API stabilize.

### DIFF
--- a/docs/api/developer.md
+++ b/docs/api/developer.md
@@ -287,6 +287,7 @@ Utility functions used by scvi-tools.
    utils.track
    utils.setup_anndata_dsp
    utils.attrdict
+   model.get_max_epochs_heuristic
 ```
 
 [ray tune]: https://docs.ray.io/en/latest/tune/index.html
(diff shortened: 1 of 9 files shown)


martinkim0:
To-do:

- Add test with CELLxGENE data pipe 
- Add dev tutorial and tutorial with CELLxGENE

PR #2472, opened 2024-02-07
Make xarray and sparse optional deps, remove chex dep, remove Pandas pin, support newest Ray, temp remove cellxgene dep

martinkim0:
- Change `xarray` and `sparse` to optional dependencies as they're only used for `RNAMixin.posterior_predictive_sample` and `PPC`
- Remove `chex` dependency (planning on adding this back as a test dependency once we have a better JAX testing framework) by changing `TuneAnalysis` to the standard dataclass and `LossOutput` as a `flax.struct.dataclass`
- Remove pin on Pandas as #2308 is resolved now
- Temporarily remove cellxgene-census dep
- Update autotune for new ray API

Need to find a cleaner way to modify `LossOutput` attributes post init.


martinkim0:
It's possible that cellxgene-census's upperbound of numpy < 1.25 is causing a lot of dependency issues https://github.com/chanzuckerberg/cellxgene-census/issues/979

PR #2567, opened 2024-02-27
deprecate `SaveBestState`

martinkim0:
Users should now use `scvi.train.SaveCheckpoint` or the `enable_checkpointing` Lightning arg instead.


martinkim0:
closing in favor of #2673

Issue #2393, opened 2024-01-17
Multimodal reference mapping with MuData objects


ethanweinberger:
In the current (1.0.4) version of scvi-tools, multimodal models (e.g. totalVI) can be set up and trained with either AnnData or MuData objects. Thus, it would be nice to be able to perform reference atlas mapping using either workflow (i.e., using an AnnData-based totalVI or MuData-based totalVI model). However, it appears that the [ArchesMixin API](https://docs.scvi-tools.org/en/1.0.2/api/reference/scvi.model.base.ArchesMixin.html) was designed only to handle AnnData objects. Indeed, when attempting to naively pass MuData objects into the ArchesMixin functions to perform reference mapping with totalVI I quickly ran into errors (notebook reproducing my results available here: https://colab.research.google.com/drive/1lL1JJ3bdG6UU3XVjA0tLvp1V2QLWihuI?usp=sharing).

@martinkim0 I'd be happy to take this on since I don't think it should be too much work to handle the MuData case- would you prefer to keep the current two function user-facing API (i.e., `load_query_data` and `prepare_query_anndata`) and handle the MuData case internally or explicitly have separate functions for MuData objects (e.g. `prepare_query_mudata`)?

martinkim0:
Thanks for bringing up this issue! I think it makes sense based on how we've named our `ArchesMixin` API to implement `prepare_query_mudata`, which I anticipate will have much different code from `prepare_query_anndata`, and add MuData support to the existing `load_query_data`, since this method mostly deals with freezing network layers.

Let me know what you think, and we can go forward from there.

ethanweinberger:
Makes sense to me! I've added this to my to-do list, and should hopefully have an initial implementation ready for review within the next week or two.

ethanweinberger:
Hey @martinkim0! Apologies for the (large) delay here. I just opened an initial PR addressing this (#2578) though I think my current way of accomplishing this is too hacky and would love to hear your thoughts on a more elegant way.

tomsing1:
@martinkim0  and @ethanweinberger Thanks a lot for working on this issue. I just tried to run the [example notebook](https://colab.research.google.com/drive/1lL1JJ3bdG6UU3XVjA0tLvp1V2QLWihuI?usp=sharing) with version with `scvi-tools` version 1.1.6 and still got the same error @ethanweinberger originally reported:

```
query_model = scvi.model.TOTALVI.load_query_data(
    mdata_query,
    ref_model,
)

[...]

PR #2578, opened 2024-02-29
Reference mapping with MuData-based models

ethanweinberger:
(Addresses #2393) While certain multimodal models (e.g. totalVI) can be used with either `AnnData` or `MuData` objects, currently the `ArchesMixin` class only supports `AnnData`-based models and so reference mapping cannot be performed with `MuData`-based models. This PR thus extends `ArchesMixin` class to handle MuData-based models by making appropriate changes to the `load_query_data` function and adding a new `prepare_query_mudata` function.

@martinkim0 my initial PR is still a work-in-progress (e.g. no unit tests yet), but I thought it would be helpful to get some initial feedback from you. To avoid duplicating code, my instinct for the `prepare_query_mudata` function was to essentially be a lightweight wrapper that calls `prepare_query_anndata` for each modality in the `MuData` object. However, when attempting to do this naively I quickly ran into issues with `var_names` for `MuData` objects being saved as a single list for all modalities. In particular, because variable names don't have identifiers for their modality of origin, I ran into issues where e.g. a protein-modality-`AnnData` would erroneously be padded with zeros for "missing" RNA-modality features. To work around this, I modified the base `save` function such that variable names are saved as a dictionary (with modalities as keys) for `MuData` objects.

This worked (as in, my code ran without errors and current tests still pass), but I imagine this would lead to backwards incompatibility issues for previously trained `MuData`-based models. Moreover, I still needed a way to tell `setup_anndata` which specific variable names to consider ahead of time, which seemed a bit ugly to me. Do you have ideas for a cleaner way to associate features with modalities?


martinkim0:
Thanks for taking a stab at this! You're right - it looks like we should save var_names separately for each modality, so it's unfortunate that we don't already do this. 

This would actually be a great feature to include in our v2 release, since our plan is to rework large parts of the codebase. So there would be no issue with backwards compatibility. 

The other alternative is to use this new save method from now on, with the caveat that we'd have to implement a method to convert "legacy" MuData saves (similar to [`convert_legacy_save`](https://docs.scvi-tools.org/en/stable/api/reference/scvi.model.base.BaseModelClass.html#scvi.model.base.BaseModelClass.convert_legacy_save)). 

I'm personally leaning more towards the first option since there would be less room for user confusion and maintenance issues. Let me know what you think! Is this a feature you're looking to use in a new model?

martinkim0 (commented):
Thanks for taking a stab at this! You're right - it looks like we should save var_names separately for each modality, so it's unfortunate that we don't already do this. 

This would actually be a great feature to include in our v2 release, since our plan is to rework large parts of the codebase. So there would be no issue with backwards compatibility. 

The other alternative is to use this new save method from now on, with the caveat that we'd have to implement a method to convert "legacy" MuData saves (similar to [`convert_legacy_save`](https://docs.scvi-tools.org/en/stable/api/reference/scvi.model.base.BaseModelClass.html#scvi.model.base.BaseModelClass.convert_legacy_save)). 

I'm personally leaning more towards the first option since there would be less room for user confusion and maintenance issues. Let me know what you think! Is this a feature you're looking to use in a new model?

scvi/model/base/_archesmixin.py
```diff
@@ -8,11 +8,13 @@
 import pandas as pd
 import torch
 from anndata import AnnData
+from muon import MuData
```
martinkim0, line 11: nit: need to import from `mudata` since we don't directly depend on `muon`

ethanweinberger:
@martinkim0 with the closing of #2769, I've gone ahead and taken another stab at this PR.

The biggest change now compared to main is that I factored out most of the old code in `prepare_query_anndata` into a new function `_pad_and_sort_query_anndata` that's reused across `prepare_query_anndata` and a new `prepare_query_mudata` method.

Let me know what you think/if there's anything else you'd like in this PR in terms of testing/documentation/etc.

ethanweinberger:
Hi @martinkim0. Just wanted to check in to see if you had any feedback for the latest version of this PR. Thanks!

src/scvi/model/base/_archesmixin.py
```diff
@@ -326,7 +359,68 @@ def _get_loaded_data(reference_model, device=None):
...
+            var_names = reference_model.adata.var_names
         load_state_dict = deepcopy(reference_model.module.state_dict())
 
     return attr_dict, var_names, load_state_dict
+
+
+def _pad_and_sort_query_anndata(
+    adata,
+    reference_var_names,
+    inplace,
```
martinkim0, line 377: typing here would be great!

martinkim0, line 377: typing here would be great!

src/scvi/model/base/_archesmixin.py
```diff
@@ -326,7 +359,68 @@ def _get_loaded_data(reference_model, device=None):
...
         attr_dict = reference_model._get_user_attributes()
         attr_dict = {a[0]: a[1] for a in attr_dict if a[0][-1] == "_"}
-        var_names = reference_model.adata.var_names
+        if isinstance(reference_model.adata, MuData):
+            var_names = {
+                mod: reference_model.adata[mod].var_names
+                for mod in reference_model.adata.mod.keys()
+            }
+        else:
+            var_names = reference_model.adata.var_names
```
martinkim0, line 368: this could be replaced by a call to `scvi.model.base._save_load._get_var_names`

martinkim0 (commented):
Thanks a lot! This looks great - I think refactoring out `_pad_and_sort_query_anndata` makes a lot of sense. Just left some minor comments regarding typing and getting var names. 

One last thing: could you add a test for `prepare_query_mudata` as well as a release note in `CHANGELOG.md` (this can go under the v1.2 release).

src/scvi/model/base/_archesmixin.py
ethanweinberger: Done. Thanks for the pointer!

src/scvi/model/base/_archesmixin.py
ethanweinberger: Done

ethanweinberger:
> Thanks a lot! This looks great - I think refactoring out `_pad_and_sort_query_anndata` makes a lot of sense. Just left some minor comments regarding typing and getting var names.
> 
> One last thing: could you add a test for `prepare_query_mudata` as well as a release note in `CHANGELOG.md` (this can go under the v1.2 release).

Added the requested typing + incorporated your suggestion on `._get_var_names`. For testing I added a `test_scarches_mudata_prep_layer` in `test_totalvi.py` analogous to `test_scarches_data_prep_layer` in `test_scvi.py`.

martinkim0 (approved):
LGTM! thanks a lot for addressing the comments

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ to [Semantic Versioning]. Full commit history is available in the
     passing in `gene_likelihood="normal"` {pr}`2780`.
 - Add {class}`scvi.external.MRVI` for modeling sample-level heterogeneity in single-cell RNA-seq
     data {pr}`2756`.
+- Add support for reference mapping with `MuData` models to {class}`scvi.model.base.ArchesMixin` {pr}`2578`.
```
martinkim0, line 43:
```suggestion
- Add support for reference mapping with {class}`mudata.MuData` models to {class}`scvi.model.base.ArchesMixin` {pr}`2578`.
```

CHANGELOG.md
```diff
@@ -40,6 +40,7 @@ to [Semantic Versioning]. Full commit history is available in the
     passing in `gene_likelihood="normal"` {pr}`2780`.
 - Add {class}`scvi.external.MRVI` for modeling sample-level heterogeneity in single-cell RNA-seq
     data {pr}`2756`.
+- Add support for reference mapping with {class}`mudata.MuData` models to {class}`scvi.model.base.ArchesMixin` {pr}`2578`.
```
martinkim0, line 43:
```suggestion
- Add support for reference mapping with {class}`mudata.MuData` models to
    {class}`scvi.model.base.ArchesMixin` {pr}`2578`.
```

PR #2608, opened 2024-03-11
Deprecation warning for scvi.nn._utils.one_hot, in favour of PyTorch version

### LINKED ISSUE #2430: Replace `scvi.nn._utils.one_hot` in favor of the PyTorch version
No deprecation as it is part of the internal API

erleholgersen:
Closes #2430 .

Somewhat cowardly approach as it still uses an internal helper function that respects the same API – let me know if you had something else in mind!

### DIFF
--- a/scvi/external/cellassign/_module.py
+++ b/scvi/external/cellassign/_module.py
@@ -7,8 +7,8 @@ from torch.distributions import Dirichlet, Normal
 
 from scvi import REGISTRY_KEYS
 from scvi.distributions import NegativeBinomial
-from scvi.module._utils import one_hot
 from scvi.module.base import BaseModuleClass, LossOutput, auto_move_data
+from scvi.nn._utils import _one_hot
 
 LOWER_BOUND = 1e-10
 THETA_LOWER_BOUND = 1e-20
@@ -114,7 +114,7 @@ class CellAssignModule(BaseModuleClass):
 
         to_cat = []
         if self.n_batch > 0:
-            to_cat.append(one_hot(tensors[REGISTRY_KEYS.BATCH_KEY], self.n_batch))
+            to_cat.append(_one_hot(tensors[REGISTRY_KEYS.BATCH_KEY], self.n_batch))
 
         cont_key = REGISTRY_KEYS.CONT_COVS_KEY
         if cont_key in tensors.keys():
@@ -125,7 +125,7 @@ class CellAssignModule(BaseModuleClass):
             for cat_input, n_cat in zip(
                 torch.split(tensors[cat_key], 1, dim=1), self.n_cats_per_cov
             ):
-                to_cat.append(one_hot(cat_input, n_cat))
+                to_cat.append(_one_hot(cat_input, n_cat))
 
         design_matrix = torch.cat(to_cat, dim=1) if len(to_cat) > 0 else None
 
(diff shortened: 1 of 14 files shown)


scvi/nn/_utils.py
```diff
@@ -1,8 +1,32 @@
...
+
+    onehot = torch.nn.functional.one_hot(index, n_cat).squeeze(-2)
+    return onehot.type(torch.float32)
+
 
 def one_hot(index: torch.Tensor, n_cat: int) -> torch.Tensor:
     """One hot a tensor of categories."""
+    # TODO: remove in 1.3.0
+    warnings.warn(
+        "The `one_hot` function is deprecated and will be removed in scvi-tools 1.3. "
```
martinkim0, line 25:
```suggestion
        "The `one_hot` function is deprecated in v1.2 and will be removed in v1.3. "
```

martinkim0 (commented):
Hi, thanks so much for putting in this PR! 

Would it be possible to refactor so that we call the torch one_hot function directly instead of having a wrapper around it? Checking the index shape is nice, but I think we should be fine without it. 

Also we can probably take out the explicit cast to float32 since I believe PyTorch should do it automatically in later operations.

erleholgersen:
Took a stab at refactoring to call torch `one_hot` directly. 

I did run into errors from the `torch.einsum` and `torch.nn.functional.linear` functions on the data types, but I have solved that by doing the conversion for those function calls. I'm by no means a PyTorch data type expert, so let me know if there are better ways you had in mind!

scvi/external/cellassign/_module.py
```diff
@@ -114,7 +113,7 @@ def _get_generative_input(self, tensors, inference_outputs):
 
         to_cat = []
         if self.n_batch > 0:
-            to_cat.append(one_hot(tensors[REGISTRY_KEYS.BATCH_KEY], self.n_batch))
+            to_cat.append(F.one_hot(tensors[REGISTRY_KEYS.BATCH_KEY], self.n_batch).squeeze(-2))
```
martinkim0, line 116:
```suggestion
            to_cat.append(F.one_hot(tensors[REGISTRY_KEYS.BATCH_KEY].squeeze(-1), self.n_batch)
```
Could we actually call `.squeeze(-1)` on `tensors[REGISTRY_KEYS.BATCH_KEY]` instead of `squeeze(-2)` (for all instances of this)? I think it makes it a bit more clear.

martinkim0 (commented):
Thanks for addressing the previous comments! Just one minor comment on when squeeze is being called. Also sorry, didn't realize einsum and linear need explicit casts to float - seems reasonable to me to just call `float()` for these instances.

scvi/external/cellassign/_module.py
erleholgersen: Oh this is a much better idea – thanks! Updated.

scvi/module/_multivae.py
```diff
@@ -225,7 +225,8 @@ class MULTIVAE(BaseModuleClass):
     n_latent
         Dimensionality of the latent space. If `None`, defaults to square root
         of `n_hidden`.
-    n_layers_encoder
+    n_layers_en
```
martinkim0, line 228:
```suggestion
    n_layers_encoder
```

scvi/module/_multivae.py
```diff
@@ -225,7 +225,8 @@ class MULTIVAE(BaseModuleClass):
...
+    coder
```
martinkim0, line 229:
```suggestion
```

martinkim0 (approved):
Thanks once again! LGTM

erleholgersen:
Thanks @martinkim0! Not sure what's going on with the tests here. Do you usually handle merging the PRs, or is there anything more I should be doing?

martinkim0:
> Thanks @martinkim0! Not sure what's going on with the tests here. Do you usually handle merging the PRs, or is there anything more I should be doing?

I can take care of merging the PR. Some of our tests are broken (not because of this PR, though).

Issue #2610, opened 2024-03-11
Transfer veloVI to external


martinkim0:
For users transferring from the original [`velovi`](https://github.com/YosefLab/velovi) package, the public API has been kept identical in the following locations:
- `velovi.VELOVI` -> `scvi.external.VELOVI`
- `velovi.VELOVAE` -> `scvi.external.velovi.VELOVAE`
- `velovi.get_permutation_scores` -> `scvi.external.velovi.get_permutation_scores`
- `velovi.preprocess_data` -> `scvi.external.velovi.preprocess_data`

PR #2611, opened 2024-03-11
feat(external): add velovi to external

martinkim0:
closes feature request #2610

### DIFF
--- a/docs/api/developer.md
+++ b/docs/api/developer.md
@@ -178,6 +178,7 @@ Module classes in the external API with respective generative and inference proc
    external.tangram.TangramMapper
    external.scbasset.ScBassetModule
    external.contrastivevi.ContrastiveVAE
+   external.velovi.VELOVAE
 
 ```
 
--- a/docs/api/user.md
+++ b/docs/api/user.md
@@ -59,6 +59,7 @@ import scvi
    external.SCBASSET
    external.ContrastiveVI
    external.POISSONVI
+   external.VELOVI
 
 ```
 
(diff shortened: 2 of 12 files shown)


adamgayoso:
cc @WeilerP

WeilerP:
@martinkim0, the idea was to have velovi as part of the `scvelo` package. Either way is fine for me, we should just only have one option, IMO. What do you think, @adamgayoso?

adamgayoso:
So I'm mostly indifferent, I had forgotten about this. But a few thoughts:

1. I think it will be easier to maintain if it's in scvi-tools because then it's easier to integrate any API changes (so that @martinkim0 doesn't have to make PRs into scvelo).
2. In that scenario, scvelo would export the model here, special utilities could be removed from here and kept only in scvelo, and the docs here can point to the scvelo docs for complete usage.

But really I think either is fine. @martinkim0 perhaps you should revert this as we make a decision.

WeilerP:
I think that's totally reasonable and the better option! Let's keep it here and I'll remove it from scvelo.

martinkim0:
Great! So I'm going to keep this commit then but remove the utility functions (`get_permutation_scores` and `preprocess_data`) and open a PR to add them to scvelo. Does that sound good?

WeilerP:
> Great! So I'm going to keep this commit then but remove the utility functions (`get_permutation_scores` and `preprocess_data`) and open a PR to add them to scvelo. Does that sound good?

Sounds good to me! Let's discuss everything else in the corresponding PR. Thanks in advance!

PR #2626, opened 2024-03-22
feat(model): support batch key in condscvi

### LINKED ISSUE #2623: Add support for `batch_key` in `CondSCVI`
Need to ensure that not passing in a `batch_key` replicates the old code path.

martinkim0:
closes #2623


martinkim0:
@canergen The main backwards compatibility concern right now is that, with the current `AnnDataManager` setup with a batch key, the model will receive a tensor of zeroes for `batch_index` even if `batch_key` is `None`. 

There's not a straightforward way to distinguish this case from the case where a `batch_key` is registered but all observations have the same batch assignment. However, we'd like to treat the latter differently from the former because of backwards compatibility. 

This is because a `batch_index` of zeros will still be passed into the encoder (if `encode_covariates` is `True`) and the decoder, and will affect model training. So anyone repeating model training from a previous version will see different results.

There are two potential ways to address this:
1. Modify `CategoricalObsField` so that it is evaluated to `None` in `AnnDataLoader` (similar to `CategoricalJointObsField`) if `registry_key` is `None`. More complicated but the better long-term solution. I need to look into this one more.
2. Just don't register a new `CategoricalObsField` if `batch_key` is `None` in `CondSCVI.setup_anndata`. Easiest solution but you have to do a couple of checks in `CondSCVI` and `VAEC`, potentially buggy.

martinkim0:
Need to test that saving + loading is backwards compatibible

martinkim0:
@canergen Here's an initial exploration with using the batch key. TLDR Seems fine but can lead to overintegration when training for too long.
[condscvi_batch.pdf](https://github.com/scverse/scvi-tools/files/14763827/condscvi_batch.pdf)

martinkim0:
Losses are very similar across the runs:
Screenshot 2024-03-26 at 12 35 32

canergen:
We have quite some different datasets to see better integration. You need to run one UMAP per coarse celltype above. I don't think those are necessary but I requested only support to get the key into DestVI :).

PR #2632, opened 2024-03-26
Update _trainingplans.py

canergen:
Disable adversarial training when using a single batch.


martinkim0 (commented):
LGTM! A test + release note would be nice

PR #2636, opened 2024-03-29
docs: update installation instructions

martinkim0:


docs/installation.md
```diff
@@ -1,110 +1,109 @@
...
-## Prerequisites
-
-scvi-tools can be installed via conda or pip. If you don't know which to choose, we recommend conda.
+## Quick install
 
-### conda
+scvi-tools can be installed via `conda` or `pip`. We recommend installing into a virtual environment
+to avoid conflicts with other packages.
 
-1. Install conda. We typically use the [mambaforge] distribution. Use python>=3.9, and consider using mamba instead of conda. Mamba is a drop-in replacement for conda that is significantly more efficient.
```
LevyNat, line 9: can we keep this?

martinkim0: I updated it to this:

> For `conda`, we recommend using the Miniforge distribution, which is generally faster than the
official distribution and comes with conda-forge as the default channel (where scvi-tools is
hosted).

martinkim0: The repo recommends using miniforge over mambaforge now

Issue #2429, opened 2024-01-23
`SCANVI.loss` fails when `n_samples>1`


martinkim0:
Reproducible script:
```
vae = SCANVAE(n_input=100, n_labels=1)
x = torch.randint(0, 100, (10, 100), dtype=torch.float32)
batch = torch.zeros(10, dtype=torch.long)
labels = torch.zeros(10, dtype=torch.long)
tensors = {
    REGISTRY_KEYS.X_KEY: x,
    REGISTRY_KEYS.BATCH_KEY: batch,
    REGISTRY_KEYS.LABELS_KEY: labels,
}

_ = vae.forward(
    tensors,
    inference_kwargs={"n_samples": 2}
)
```

martinkim0:
sanity check with the [Atlas-level integration of lung data](https://docs.scvi-tools.org/en/stable/tutorials/notebooks/scrna/harmonization.html) tutorial: [harmonization.pdf](https://github.com/scverse/scvi-tools/files/14829605/harmonization.pdf)

ori-kron-wis:
see https://github.com/scverse/scvi-tools/pull/2902 with the fix

PR #2644, opened 2024-04-01
chore(module): remove unused code paths in scanvae loss

### LINKED ISSUE #2134: Clean up loss computation in SCANVI
There are currently conditional cases that are never reached based on the arguments that are passed into `scvi.module.SCANVAE.loss`.

martinkim0:
BREAKING CHANGE: removes unused code paths in the loss method of SCANVAE due to the default being `feed_labels=False`. internal functions that call this method either have their own default set to `feed_labels=False` or have `feed_labels=True` but don't pass it into the loss method (e.g. compute_elbo). 

also adds typing within the modified file.

closes #2134


scvi/module/_scanvae.py
```diff
@@ -323,32 +320,6 @@ def loss(
         else:
             kl_divergence_l = 0.0
 
-        if is_labelled:
```
martinkim0, line 326: `feed_labels` is always `False` => `y` is always None` => `is_labelled` is always `False`

scvi/module/_scanvae.py
```diff
@@ -275,28 +279,21 @@ def classification_loss(self, labelled_dataset):
...
+        y = None
 
-        if feed_labels:
-            y = tensors[REGISTRY_KEYS.LABELS_KEY]
-        else:
-            y = None
-        is_labelled = False if y is None else True
-
-        # Enumerate choices of label
         ys, z1s = broadcast_labels(y, z1, n_broadcast=self.n_labels)
```
martinkim0, line 297: could probably clean up this function call since `y` is always `None` - it would help with debugging and addressing #2429

martinkim0:
per our discussion in slack @canergen

martinkim0:
sanity check with the [Atlas-level integration of lung data](https://docs.scvi-tools.org/en/stable/tutorials/notebooks/scrna/harmonization.html) tutorial: [harmonization.pdf](https://github.com/scverse/scvi-tools/files/14829605/harmonization.pdf)

canergen:
LGTM! Would an old model still load? I guess so as all of these feed_labels are created dynamically. Otherwise, we need to update the scvi-hub scANVI models.

canergen:
Not sure it's breaking in this case as it was never active.

martinkim0:
Yeah I think it should be backwards compatible - we have tests for old scanvi loads.

PR #2689, opened 2024-04-04
fix(external): ensure solo soft predict returns probs

### LINKED ISSUE #2688: Document `SOLO.predict()` return values better.
The [documentation](https://docs.scvi-tools.org/en/stable/api/reference/scvi.external.SOLO.html#scvi.external.SOLO.predict
) reads:

> soft ([bool](https://docs.python.org/3/library/functions.html#bool) (default: True)) – Return probabilities instead of class label

However, the return values are not probabilities -- they range from -30 to 30 in my case. How should I interpret these? 

image

martinkim0:
closes #2688

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,13 @@ is available in the [commit logs](https://github.com/scverse/scvi-tools/commits/
 -   Remove {class}`scvi.train._callbacks.MetricsCallback` and argument `additional_val_metrics` in
     {class}`scvi.train.Trainer` {pr}`2646`.
 
+### 1.1.3 (2024-MM-DD)
+
+#### Fixed
+
+-   Fix {meth}`scvi.external.SOLO.predict` to correctly return probabiities instead of logits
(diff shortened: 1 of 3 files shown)


canergen:
It needs to be labeled as breaking change. It might be some folks use softmax after SOLO (I do). I guess it's better to also print a warning about the change for scVI-tools 1.3.

martinkim0:
> It needs to be labeled as breaking change. It might be some folks use softmax after SOLO (I do). I guess it's better to also print a warning about the change for scVI-tools 1.3.

Right - I forgot to do that. This is slated to be released with 1.1.3, should we revert it then? It's technically a bug fix so IMO it should be fine.

canergen:
I think it's fine to release with 1.1.3 and I prefer outputting probabilities. I agree the code was wrong and not the documentation. However, we should highlight this change as it might affect analysis scripts. I'm open whether it outputs a warning or not but the release note should contain a breaking change.

martinkim0:
Makes sense! I'll add a warning and a note to the changelog.

PR #2696, opened 2024-04-08
Save per-modality feature names for MuData models

ethanweinberger:
Current MuData-based models save their `var_names` as a single list, with no information on which feature came from which modality. This is problematic for certain downstream features (e.g. reference mapping [#2393](https://github.com/scverse/scvi-tools/issues/2393)) which require knowledge of features' modalities. This PR thus modifies the base `save` function to save `var_names` as a dictionary for MuData models, where each key is a modality and each value is that modality's features. In anticipation of potential backwards compatibility issues, this PR also includes a `convert_legacy_mudata_save` method to update earlier MuData models saved under the previous scheme.

@martinkim0 @canergen Per our discussion I added convenience functions for converting legacy models saved with the old MuData var names saving scheme (as opposed to just waiting until scvi-tools 2.0 and not worrying about backwards compatibility). I made this a separate PR from the scArches MuData PR (https://github.com/scverse/scvi-tools/pull/2578) to keep each PR more focused/reduce reviewer burden.


scvi/model/base/_base_model.py
```diff
@@ -773,6 +779,62 @@ def convert_legacy_save(
...
+    @classmethod
+    def convert_legacy_mudata_save(
+        cls,
+        dir_path: str,
+        output_dir_path: str,
+        overwrite: bool = False,
+        prefix: str | None = None,
+        **save_kwargs,
+    ) -> None:
+        """Converts a legacy MuData-based model (<1.XX) to the updated save format.
```
ethanweinberger, line 791: Leaving 1.XX as a placeholder pending discussion on when this change should be incorporated

martinkim0 (commented):
Hi @ethanweinberger, thanks so much for taking a stab at this. I think the changes to mudata saves makes a lot of sense. However, I'd like to approach converting legacy saves a bit differently since we already have a legacy save convert method, so this might be confusing to users.

What I'd like to propose is the following:
- new mudata-based models save per-mod var names by default (and raise a user warning here)
- detect old model saves in `load` and implicitly convert them to the new format (and raise an implicit modification warning)
- this means that new model saves won't be compatible with previous versions (but in this case we'll recommend upgrading scvi-tools)

Let me know what you think!

martinkim0:
Just to add to my comment above, we do some validation checking that the saved var names match with the loaded mudata in `load`.

In the case where an old model save is loaded with new code, we can just check that the union of per-mod var names in the mudata is equal to the var names in the model save (since I believe the top level var names in mudata corresponds to the union of all mods).

ethanweinberger:
Hi @martinkim0 

Your suggested path forward largely makes sense to me. My one question is: what's the logic behind raising a user warning with the new default save scheme (i.e., saving per-mod var names for new mudata models)?

martinkim0:
sorry for the delay - I've merged support for this with #2769. thanks for putting in this initial pr! i'll close this now.

Issue #2257, opened 2023-09-06
MultiVI with Poisson/NB likelihood for ATAC?


vitkl:
Hi @martinkim0 and all

Similarly to https://github.com/scverse/scvi-tools/pull/2249 by @lauradmartens would be great to have a MultiVI version with Poisson or even NB likelihood for ATAC data. How hard do you @martinkim0 think it is to add such option to MultiVI?

canergen:
Hi @vitkl I'm linking @marianogabitto here as he developed multiVI and was also working on a different implementation of PoissonPeakVI. What's your opinion on this request?

lauradmartens:
Yes, I also think it should be relatively straightforward and also remember that @marianogabitto already had an implementation

vitkl:
Thanks for getting back to me about this! If you @marianogabitto have an implementation of this - would be great if you could share it or add it to scvi-tools (if makes sense for @canergen)

canergen:
Happy to review the PR.

vitkl:
I don't understand the implementations well enough to easily try implementing this - but I can do a PR if I see the example implementation from @marianogabitto

canergen:
I'm sorry for the confusion that was a comment to @marianogabitto to assign me the PR if he creates one.

Juliendly:
Hi all, I am very interested in this implementation and I'm curious to know if there is any updates on it?
Thanks!

canergen:
@lauradmartens congrats on the new manuscript: https://www.biorxiv.org/content/10.1101/2024.09.19.613754v2. Any plan to continue with the Poisson MultiVI branch.

PR #2735, opened 2024-04-18
Poissonmultivi

lauradmartens:
Hi @martinkim0,

I had some time and added the Poisson loss also to the MultiVI model (POISSONMULTIVI) to the external models.

Addresses #2257 for Poisson likelihood.


lauradmartens:
I added the option to have separate size factors for RNA and ATAC, not sure if this is the best way.

canergen:
Quick check looks good. Can you explain, why we don't inherit from MULTIVI and MULTIVAE?

canergen:
Closing here due to inactivity. We have our own proposed Poisson distribution in multiVI that will be merged in the near future.

PR #2748, opened 2024-04-23
feat(model): support custom dataloaders in vaemixin

martinkim0:
adds custom dataloader arguments to vaemixin methods. currently does not check that the dataloader outputs in the correct format such that errors will be raised at the module level, which will potentially be confusing.

also includes a refactor of some of the vaemixin methods, including more detailed and consistent docstrings.


martinkim0:
for consistency, the `data_module` argument in `train` should be renamed to `datamodule`, in line with `dataloader`.

PR #2762, opened 2024-04-30
replace statsmodels with scipy stats

martinkim0:


src/scvi/external/mrvi/_model.py
```diff
@@ -1112,7 +1112,7 @@ def differential_expression(
         """
         from functools import partial
 
-        from statsmodels.stats.multitest import multipletests
+        from scipy.stats import false_discovery_rate
```
justjhong, line 1115: should be false_discovery_control, not rate. let's make sure this one passes tests.

martinkim0:
Tests are currently failing because `pvalue` has NaNs (before `false_discovery_control` is called). This wasn't being caught before for some reason (does statsmodels not check for NaNs?)

justjhong:
@martinkim0 good catch, looks like statsmodels would just output nans. This happens when the input to chi2.cdf is 0 (which usually shouldn't happen -  happens here since its barely training I assume), but we can still correctly handle this case by making the pval 1 for these examples.

justjhong (approved):

martinkim0:
Awesome, thanks for pushing the changes

PR #2769, opened 2024-05-01
feat(model): save mudata var names per modality

martinkim0:
previously, mudata-based models (totalvi and tangram) saved variable names in a flat array with values concatenated across all modalities - this is what happens when MuData.var_names is called directly. this is an issue for performing scarches on such a model as we cannot pad with zeros without being able to map individual variable names back to their respective modalities.

this commit sets the default to saving per-modality var names as a dictionary, with keys corresponding to modality keys and values to variable names. additionally, it adds a flag to the save method in case a user wants to use the legacy saving format.

old model saves will still be loaded successfully as the code will detect if the legacy save was used and validate var_names accordingly.

BREAKING CHANGE: changes the default saving scheme for mudata-based models. see above for details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,11 @@ to [Semantic Versioning]. Full commit history is available in the
     accessibility data {pr}`2739`.
 - The `data_module` argument in {meth}`scvi.model.base.UnsupervisedTrainingMixin.train` has been
     renamed to `datamodule` for consistency {pr}`2749`.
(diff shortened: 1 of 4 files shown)


martinkim0:
edge case: if a user decides to change a modality between saving and loading a model (_including_ modalities not used directly), this will raise an error on our end. this issue is also present in the old code. 

the solution would be to only save var names for modalities registered with the model - will be tackled in a separate pr.

PR #2826, opened 2024-06-03
compute DE+ and DE- separately

LevyNat:


canergen:
Fixed the wrong behavior that we have seen in different models (high significance despite low mean LFC but high standard deviation). It still requires a release note and it would be good to check in our DE tutorial whether things are changing there. I’m not sure that the Bayes factor computation is correct @PierreBoyeau?

PierreBoyeau:
Hi, LGTM. I am unsure about the Bayes factor here, since we don't have two, but three competing models now(equally expressed, upreg, and downreg). I would be in favor of not computing BFs in this case.

canergen:
I guess it's fine to remove. I would assume we do two tests: upregulated vs rest and downregulated vs rest. We output only the winner of the test for upregulated/downregulated. I guess the corresponding Bayes factor for the returned FDR is then: `upregulated/(equally+downregulated)` or `downregulated/(equally+upregulated)`.

src/scvi/model/base/_differential.py
```diff
@@ -360,7 +365,8 @@ def m1_domain_fn(samples):
             res = dict(
                 proba_de=proba_m1,
```
canergen, line 366: Why not return probe_de?

LevyNat: updated

LevyNat:
@canergen @PierreBoyeau can we merge to main? 

I put the new proba_de computation in mode='change_threeway' and kept 'vanilla' as default. Just need to think about how we explain the fix in the documentation

canergen:
Put in other fixes, that we discussed when we met. Especially the threshold was computed wrong. Subset_idx is something that I need for resolVI and it needs to be in this function. The argument is also different now with two and three.

ori-kron-wis:
we need to rerun all tutorials that use DE once merged.

ori-kron-wis:
this branch was already merged in this PR: https://github.com/scverse/scvi-tools/pull/3144

PR #2842, opened 2024-06-17
feat: support numpy 2.0

### LINKED ISSUE #2390: Add support for NumPy 2
https://pythonspeed.com/articles/numpy-2/

martinkim0:
closes #2390


martinkim0:
this is currently blocked by the following upstream dependencies not supporting numpy 2.0:
- mudata
- loompy
- pymde

Zethson:
The last release of pymde was in 2022 and the last commit a year ago. I am not sure how sustainable it is to keep depending on it...

canergen:
Agreed seems like lowly maintained. I am then also against keeping it optional and we should just highlight it in some tutorial. With RAPIDS UMAP the need is also much smaller now. Loompy still doesn't support >2.0 based on a quick check though. So let's wait until they have figured things out and drop pyMDE then.

canergen:
@ori-kron-wis Can you switch loom to anndata.io (https://anndata.readthedocs.io/en/latest/generated/anndata.io.read_loom.html)? I assume we then support numpy>2.0. We should for now make pyMDE optional and set an optional requirement to numpy<2.0 for this dependency. We should drop it for the 1.3 release.

ori-kron-wis:
@canergen 
the issue is with numpy>2.0 + loompy.connect, which is barely maintained.

At anndata at one hand does not have a numpy upper limit constraint, and at the other has its own read_loom function that uses the same loompy.connect under the hood. Tried to use that and got the same error as above, @ilan-gold can confirm he sees that error in his tests.

Therefore, unless loompy will solve it (see https://github.com/linnarsson-lab/loompy/issues/186), I dont think theres much we can do but to issue warnings for the next release + to have it optional in tests, like census. See my commits. 

As I mentioned,  cellxgene census package, which now uses tiledbsoma is also pinned for numpy<2.0

GimVI which is the only model uses loom should state it needs numpy<2.0 in its tutorial.

pymde, which is also barely maintained, is running fine on Python 3.12. no reason to remove it but to alert for the python version (the error is about an ongoing issue with numpy >=2)

We also need to decide whether we want this whole change into v1.3, hopefully, numpy 2 issues, which are globally, will be solved.

canergen:
I assume loompy is a tutorial dependency? Let's get rid of loompy. I set up dvc yesterday with gdrive backend. We could switch all our tutorials to use our own files and store them as h5ad preprocessed. It will make us less dependent on some internet resources.
We don't need pyMDE anymore (I would argue rapids UMAP is good) and it's not well maintained. It's also quite inconsistent as it's utils code.
For cellXgene-census, we should raise an issue there.

ori-kron-wis:
Pymde was removed from tutorials as well (see https://github.com/scverse/scvi-tutorials/pull/378)

PR #2867, opened 2024-06-28
fix(module): address shape mismatch for scanvae loss n_samples

### LINKED ISSUE #2429: `SCANVI.loss` fails when `n_samples>1`

martinkim0:
closes #2429

### DIFF
--- a/src/scvi/module/_scanvae.py
+++ b/src/scvi/module/_scanvae.py
@@ -287,14 +287,13 @@ class SCANVAE(VAE):
         classification_ratio: float | None = None,
     ):
         """Compute the loss."""
-        px = generative_ouputs["px"]
-        qz1 = inference_outputs["qz"]
-        z1 = inference_outputs["z"]
-        x = tensors[REGISTRY_KEYS.X_KEY]
-        batch_index = tensors[REGISTRY_KEYS.BATCH_KEY]
-        y = None
+        px: Distribution = generative_ouputs["px"]
+        qz1: torch.Tensor = inference_outputs["qz"]
+        z1: torch.Tensor = inference_outputs["z"]
+        x: torch.Tensor = tensors[REGISTRY_KEYS.X_KEY]
+        batch_index: torch.Tensor = tensors[REGISTRY_KEYS.BATCH_KEY]
 
-        ys, z1s = broadcast_labels(y, z1, n_broadcast=self.n_labels)
+        ys, z1s = broadcast_labels(z1, n_broadcast=self.n_labels)
         qz2, z2 = self.encoder_z2_z1(z1s, ys)
         pz1_m, pz1_v = self.decoder_z1_z2(z2, ys)
         reconst_loss = -px.log_prob(x).sum(-1)
--- a/src/scvi/module/_scanvae.py
+++ b/src/scvi/module/_scanvae.py
@@ -343,6 +343,10 @@ class SCANVAE(VAE):
         else:
             reconst_loss += loss_z1_weight + loss_z1_unweight
             kl_divergence = kl_divergence_z2
+            kl_divergence += kl(
+                 Categorical(probs=torch.ones([1], device=z1.device)),
+                 Categorical(probs=torch.ones([1], device=z1.device)),
+             )
 
         kl_divergence += kl_divergence_l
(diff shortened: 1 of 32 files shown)


src/scvi/module/_utils.py
```diff
@@ -1,45 +1,31 @@
...
+        # return size (batch_size, n_broadcast)
+        return torch.nn.functional.one_hot(labels.squeeze(-1), n_broadcast)
+
+    if o.ndim == 2:
+        batch_size = o.size(0)
+        ys = torch.cat([batch(batch_size, i) for i in range(n_broadcast)])
+        new_o = o.repeat(n_broadcast, 1)
+    elif o.ndim == 3:
+        n_samples = o.size(0)
+        batch_size = o.size(1)
```
canergen, line 23: o.size(-2) should work for both.

src/scvi/module/_scanvae.py
```diff
@@ -324,15 +323,26 @@ def loss(
...
-
-        kl_divergence = (kl_divergence_z2.view(self.n_labels, -1).t() * probs).sum(dim=1)
-        kl_divergence += kl(
-            Categorical(probs=probs),
-            Categorical(probs=self.y_prior.repeat(probs.size(0), 1)),
-        )
+        if self.n_labels > 1:
+            reconst_loss += loss_z1_weight + (
+                (loss_z1_unweight).view(self.n_labels, -1).t() * probs
+            ).sum(dim=1)
```
canergen, line 329: .sum(dim=-1)

src/scvi/module/_scanvae.py
```diff
@@ -324,15 +323,26 @@ def loss(
...
+            kl_divergence = (kl_divergence_z2.view(self.n_labels, -1).t() * probs).sum(dim=1)
```
canergen, line 330: this line requires a more complex transpose (preferred like torch.transpose(values, -1, -2) ) or requires using torch.einsum.

src/scvi/module/_scanvae.py
canergen: Remove this check with self.n_labels>1 again.

src/scvi/module/_utils.py
```diff
@@ -1,45 +1,31 @@
 import torch
 
 
-def iterate(obj, func):
```
canergen, line 4: Have you removed this function? Is it not in use at other locations?

src/scvi/module/_utils.py
ori-kron-wis: not so sure what you meant. o.size(-2) gave me 128 which is like o.size(1) but

canergen: You call here batch_size = o.size(1) and for ndim=2 batch_size = o.size(0). It's nicer to call o.size(-2) and do it outside of the if statement.

src/scvi/module/_scanvae.py
ori-kron-wis: This comes from the previous solution, which was fine, for the case of n_labels>1 and n_samples==1

src/scvi/module/_scanvae.py
ori-kron-wis: This comes from the previous solution, which was fine, for the case of n_labels>1 and n_samples==1

src/scvi/module/_utils.py
ori-kron-wis: This is from Martin's solution.

src/scvi/module/_utils.py
ori-kron-wis: got it.

PR #2868, opened 2024-06-28
style: add initial tensor shape typing to vae

### LINKED ISSUE #2541: Consider adding tensor/array shape annotations with jaxtyping
See #2496

martinkim0:
closes #2541.

draft implementation of adding jaxtyping tensor shape annotation to vae


martinkim0:
Pros:
- Same syntax across JAX, PyTorch, and NumPy (not sure about Xarray or Sparse yet)
- Track shape annotations for tensors, which can make it easier to catch bugs earlier
Cons:
- Can be tedious to add typing to every function - probably need to decide when it's recommended and when it's optional
- Can sometimes lead to more verbose code
Other considerations:
- Need to use the same axis names across all the models for consistency, so need a good system to sync these and manage axis name conflicts

PR #2902, opened 2024-07-16
feat(dataloaders): add argument for custom data split indexes

ori-kron-wis:
a branch to handle several data splitting issues

### DIFF
--- a/src/scvi/dataloaders/_data_splitting.py
+++ b/src/scvi/dataloaders/_data_splitting.py
@@ -22,7 +22,8 @@ from scvi.utils._docstrings import devices_dsp
 
 
 def validate_data_split(
-    n_samples: int, train_size: float, validation_size: Optional[float] = None
+    n_samples: int, train_size: float, validation_size: Optional[float] = None,
+    use_external_indexing: Optional[bool] = False, external_indexing: Optional[list] = None
 ):
     """Check data splitting parameters and return n_train and n_val.
 
@@ -34,27 +35,70 @@ def validate_data_split(
         Size of train set. Need to be: 0 < train_size <= 1.
     validation_size
         Size of validation set. Need to be 0 <= validation_size < 1
+    use_external_indexing
+        Wheter to use external supproted indexing. This bypass any other flag or input parameter that was before
+    external_indexing
+        A list of np.arrays that is always in the order of [[train_idx],[valid_idx],[test_idx]].
     """
-    if train_size > 1.0 or train_size <= 0.0:
-        raise ValueError("Invalid train_size. Must be: 0 < train_size <= 1")
+    if use_external_indexing:
+        if external_indexing is None:
+            raise ValueError("External indexing list is missing")
 
-    n_train = ceil(train_size * n_samples)
+        if type(external_indexing) is not list:
+            raise ValueError("External indexing is not of list type")
 
-    if validation_size is None:
-        n_val = n_samples - n_train
-    elif validation_size >= 1.0 or validation_size < 0.0:
-        raise ValueError("Invalid validation_size. Must be 0 <= validation_size < 1")
-    elif (train_size + validation_size) > 1:
-        raise ValueError("train_size + validation_size must be between 0 and 1")
+        #validate the structure of it
+        if type(external_indexing[0]) is not np.ndarray:
+            raise ValueError("The first element of the list of lists is not a np.array")
+
+        if external_indexing[1] is None:
+            #we can assume the 2nd element is not exists or empty list (i.e, only train index are supported)
+            external_indexing[1] = np.array([]) #empty np.array
+        if external_indexing[2] is None:
+            #we can assume the 3rd element is not exists or empty list (i.e, only train and valid index are supported)
+            external_indexing[2] = np.array([]) #empty np.array
+        if type(external_indexing[1]) is not np.ndarray:
+            raise ValueError("The second element of the list of lists is not a np.array")
+        if type(external_indexing[2]) is not np.ndarray:
+            raise ValueError("The third element of the list of lists is not a np.array")
+
+        #check for total number of indexes (overlapping or missing)
+        if (len(external_indexing[0])+len(external_indexing[1])+len(external_indexing[2]))>n_samples:
+            raise ValueError("There are overlapping indexing please fix")
+        if (len(external_indexing[0])+len(external_indexing[1])+len(external_indexing[2]))<n_samples:
+            raise ValueError("There are missing indexing please fix or remove those lines")
+        if len(np.intersect1d(external_indexing[0], external_indexing[1])) != 0:
+            raise ValueError("There are overlapping indexing between train and valid sets")
+        if len(np.intersect1d(external_indexing[0], external_indexing[2])) != 0:
+            raise ValueError("There are overlapping indexing between train and test sets")
+        if len(np.intersect1d(external_indexing[2], external_indexing[1])) != 0:
+            raise ValueError("There are overlapping indexing between test and valid sets")
+
+        n_train = len(external_indexing[0])
+        n_val = len(external_indexing[1])
+
     else:
-        n_val = floor(n_samples * validation_size)
+        # just like it used to be w/o external indexing
+        if train_size > 1.0 or train_size <= 0.0:
+            raise ValueError("Invalid train_size. Must be: 0 < train_size <= 1")
 
-    if n_train == 0:
-        raise ValueError(
-            f"With n_samples={n_samples}, train_size={train_size} and "
-            f"validation_size={validation_size}, the resulting train set will be empty. Adjust "
-            "any of the aforementioned parameters."
-        )
+        n_train = ceil(train_size * n_samples)
+
+        if validation_size is None:
+            n_val = n_samples - n_train
+        elif validation_size >= 1.0 or validation_size < 0.0:
+            raise ValueError("Invalid validation_size. Must be 0 <= validation_size < 1")
+        elif (train_size + validation_size) > 1:
+            raise ValueError("train_size + validation_size must be between 0 and 1")
+        else:
+            n_val = floor(n_samples * validation_size)
+
+        if n_train == 0:
+            raise ValueError(
+                f"With n_samples={n_samples}, train_size={train_size} and "
+                f"validation_size={validation_size}, the resulting train set will be empty. Adjust "
+                "any of the aforementioned parameters."
+            )
 
     return n_train, n_val
 
@@ -83,6 +127,11 @@ class DataSplitter(pl.LightningDataModule):
     pin_memory
         Whether to copy tensors into device-pinned memory before returning them. Passed
         into :class:`~scvi.data.AnnDataLoader`.
+    use_external_indexing
+        Wheter to use external supproted indexing. This bypass any other flag or input parameter that was before
+    external_indexing
+        A list of np.arrays that is always in the order of [[train_idx],[valid_idx],[test_idx]].
+        User is responsible to insert the correct indices, but there is overlapping/missing indeces validation checks
     **kwargs
         Keyword args for data loader. If adata has labeled data, data loader
         class is :class:`~scvi.dataloaders.SemiSupervisedDataLoader`,
@@ -108,6 +157,8 @@ class DataSplitter(pl.LightningDataModule):
         shuffle_set_split: bool = True,
         load_sparse_tensor: bool = False,
         pin_memory: bool = False,
+        use_external_indexing: bool = False,
+        external_indexing: list = None,
         **kwargs,
     ):
         super().__init__()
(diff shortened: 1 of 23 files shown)


src/scvi/dataloaders/_data_splitting.py
```diff
@@ -22,7 +22,11 @@
 
 
 def validate_data_split(
-    n_samples: int, train_size: float, validation_size: Optional[float] = None
+    n_samples: int,
+    train_size: float,
+    validation_size: Optional[float] = None,
+    use_external_indexing: Optional[bool] = False,
+    external_indexing: Optional[list] = None,
```
martinkim0, line 29: I like the use of `external_indexing` to contain all data split indices instead of having separate arguments for training and validation. To make it more concise, we could remove `use_external_indexing` and assume that `external_indexing` will be used when it is not `None`

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -34,27 +38,77 @@ def validate_data_split(
         Size of train set. Need to be: 0 < train_size <= 1.
     validation_size
         Size of validation set. Need to be 0 <= validation_size < 1
+    use_external_indexing
+        Whether to use external supported indexing. This bypass any other flag or input parameter
+        that was before
+    external_indexing
+        A list of np.arrays that is always in the order of [[train_idx],[valid_idx],[test_idx]].
```
martinkim0, line 45:
```suggestion
        A list of data split indexes in the order of training, validation, and test sets.
```
nit: Since our docs will render type annotations directly, we can remove type references in argument docstrings for less redundancy.

martinkim0: We should also mention that validation and test are not required

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -34,27 +38,77 @@ def validate_data_split(
...
     """
-    if train_size > 1.0 or train_size <= 0.0:
-        raise ValueError("Invalid train_size. Must be: 0 < train_size <= 1")
+    if use_external_indexing:
+        if external_indexing is None:
+            raise ValueError("External indexing list is missing")
+
+        if type(external_indexing) is not list:
```
martinkim0, line 51: nit: For our code style it's preferable to use `not isinstance(external_indexing, list))`

martinkim0: Applicable to other lines in this function as well.

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -34,27 +38,77 @@ def validate_data_split(
...
+            raise ValueError("External indexing is not of list type")
+
+        # validate the structure of it
+        if type(external_indexing[0]) is not np.ndarray:
+            raise ValueError("The first element of the list of lists is not a np.array")
+
+        if external_indexing[1] is None:
```
martinkim0, line 58: This will fail if `len(external_indexing) < 2)`, same for the `external_indexing[2] is None` check below.

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -34,27 +38,77 @@ def validate_data_split(
...
+    if use_external_indexing:
+        if external_indexing is None:
+            raise ValueError("External indexing list is missing")
+
+        if type(external_indexing) is not list:
+            raise ValueError("External indexing is not of list type")
+
+        # validate the structure of it
+        if type(external_indexing[0]) is not np.ndarray:
+            raise ValueError("The first element of the list of lists is not a np.array")
```
martinkim0, line 56: For the sake of not repeating these checks explicitly for each element, you can do something like:
```
if not all(isinstance(elem, np.ndarray) for elem in external_indexing)):
   # raise error here
```
No need to specify which element is not the correct type - I think the users can figure that out.

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -34,27 +38,77 @@ def validate_data_split(
...
+
+        # check for total number of indexes (overlapping or missing)
+        if (
+            len(external_indexing[0]) + len(external_indexing[1]) + len(external_indexing[2])
+        ) > n_samples:
+            raise ValueError("There are overlapping indexing please fix")
+        if (
+            len(external_indexing[0]) + len(external_indexing[1]) + len(external_indexing[2])
+        ) < n_samples:
+            raise ValueError("There are missing indexing please fix or remove those lines")
```
martinkim0, line 79: I think this scenario OK to have - some users may wish to leave out some observations. We can still raise a warning here though.

src/scvi/dataloaders/_data_splitting.py
martinkim0: We can also make the typing more explicit here with `Optional[list[npt.ArrayLike]]`

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -118,24 +181,38 @@ def __init__(
...
+        self.use_external_indexing = use_external_indexing
+        self.external_indexing = external_indexing
 
         self.n_train, self.n_val = validate_data_split(
-            self.adata_manager.adata.n_obs, self.train_size, self.validation_size
+            self.adata_manager.adata.n_obs,
+            self.train_size,
+            self.validation_size,
+            self.use_external_indexing,
+            self.external_indexing,
```
martinkim0, line 192: Since `validate_data_split` is now trying to handle two sets of arguments (one for `train_size` and the other for `external_indexing`), maybe we can refactor this out into two functions instead. What do you think?

tests/dataloaders/test_datasplitter.py
```diff
@@ -40,6 +40,38 @@ def test_datasplitter_shuffle():
     )
 
 
+def test_datasplitter_external():
```
martinkim0, line 43: Ideally we should also be testing for error cases (overlapping indexes, missing input arguments, etc). You can do this with `with pytest.raises`.

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -34,27 +38,77 @@ def validate_data_split(
...
+        if type(external_indexing[1]) is not np.ndarray:
+            raise ValueError("The second element of the list of lists is not a np.array")
+        if type(external_indexing[2]) is not np.ndarray:
+            raise ValueError("The third element of the list of lists is not a np.array")
+
+        # check for total number of indexes (overlapping or missing)
+        if (
+            len(external_indexing[0]) + len(external_indexing[1]) + len(external_indexing[2])
+        ) > n_samples:
+            raise ValueError("There are overlapping indexing please fix")
```
martinkim0, line 75: An edge case for this check is if any of the indexes contains duplicate elements, e.g., the train split having `[1, 2, 3, 1]`.

martinkim0: One way to address this would be to convert each element in the input list to a numpy set. This will also make the `np.intersect1d` checks below faster.

martinkim0 (dismissed):
Great start! One major comment regarding refactoring the logic here into two separate functions, and some minor comments about code style, error checks, and testing.

Could you also add a release note under `CHANGELOG.md` in the same format as the other entries? This can go under the `1.2.0` release.

Also, if you have time, feel free to use the newer Python 3.10+ type annotations in `src/scvi/dataloaders/_data_splitting.py` by adding a `from __future__ import annotations` at the top of the file.

martinkim0:
I'm also going to go ahead and change the title of the PR to match our [commit convention](https://docs.scvi-tools.org/en/latest/developer/maintenance.html#conventional-commits).

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: Make sense

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: gotcha. done.

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: OK I see.

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: Sure, I actually added here handling for the case where the 2nd or 3rd arrays are even not given to us, and imputed that . But made it now more compact.

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: it is now different

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: alright

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: I would keep only the intersection part and add np.unique(s)

I didn't think about duplicates.
The problem is not only here, the user can insert duplicate indexes per set (not overlapping) and its train/test/valid will hold duplicate rows. I don't think it makes sense. but maybe he will want to up/down sample for some reason.

I added to raise a warning for duplicates per set

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: Can you elaborate? Wouldn't this be much more challenging to implement and manage?

ori-kron-wis: change to sets in the beginning of validation and use that further on code.
divide only the validation part to 2 cases
We want to implement it in the semi-supervised class as well (divide to labeld/unlabled)

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -52,13 +52,74 @@ def validate_data_split(
...
+    if len(external_indexing_unique[2]) < len(external_indexing[2]):
+        raise Warning("There are duplicate indexing in test set")
+
+    # check for total number of indexes (overlapping or missing)
+    if (
+        len(external_indexing_unique[0])
+        + len(external_indexing_unique[1])
+        + len(external_indexing_unique[2])
+    ) < n_samples:
+        raise Warning("There are missing indexing please fix or remove those lines")
```
canergen, line 108: indexing = indices

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -83,6 +144,9 @@ class DataSplitter(pl.LightningDataModule):
     pin_memory
         Whether to copy tensors into device-pinned memory before returning them. Passed
         into :class:`~scvi.data.AnnDataLoader`.
+    external_indexing
+        A list of data split indexes in the order of training, validation, and test sets.
+        Validation and test set and not required and can be left empty.
```
canergen, line 149: and not = are not.
Left empty = setting it to None or not using it?

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -260,53 +345,78 @@ def __init__(
...
-            n_labeled_train, n_labeled_val = validate_data_split(
-                n_labeled_idx, self.train_size, self.validation_size
-            )
-
-            labeled_permutation = self._labeled_indices
-            if self.shuffle_set_split:
-                rs = np.random.RandomState(seed=settings.seed)
-                labeled_permutation = rs.choice(
-                    self._labeled_indices, len(self._labeled_indices), replace=False
+            # Need to separate tp the external and non-external cases of the labeled indices
```
canergen, line 356: what's tp?

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -260,53 +345,78 @@ def __init__(
...
-                    len(self._unlabeled_indices),
-                    replace=False,
+            # Need to separate tp the external and non-external cases of the unlabeled indices
+            if self.external_indexing is not None:
+                # we need to intersect the external indexing given with the labeled indices
+                unlabeled_idx_train, unlabeled_idx_val, unlabeled_idx_test = (
+                    np.intersect1d(self.external_indexing[n], self._unlabeled_indices)
+                    for n in range(3)
+                )
+                n_labeled_train, n_labeled_val = validate_data_split_with_external_indexing(
```
canergen, line 397: Should be n_unlabeled_train?

src/scvi/external/gimvi/_model.py
```diff
@@ -202,6 +203,9 @@ def train(
         plan_kwargs
             Keyword args for model-specific Pytorch Lightning task. Keyword arguments passed
             to `train()` will overwrite values present in `plan_kwargs`, when appropriate.
+        external_indexing
+            A list of data split indexes in the order of training, validation, and test sets.
+            Validation and test set and not required and can be left empty.
```
canergen, line 208: and =are
indexes = indices

tests/dataloaders/test_datasplitter.py
```diff
@@ -40,6 +40,79 @@ def test_datasplitter_shuffle():
...
+    adata = scvi.data.synthetic_iid()
```
canergen, line 44: We should test here also all modules with specific exceptions. I.e. pyro, contrastive, scanvi, scvi. Those tests should go in the respective module tests. The error with n_unlabelled above should throw an error in the scANVI test but not in the tests here.

canergen (commented):
Please add the requested tests and fix the typos in the documentation and explain how to set indices to empty in the docstring. Afterwards fine and can be merged.

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: done

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: Both will work actually.

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: type - need to be to

src/scvi/external/gimvi/_model.py
ori-kron-wis: fixed

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: yes missed that

tests/dataloaders/test_datasplitter.py
ori-kron-wis: Yes you are correct.
I will add those and it will be a good option to see how to run a different module.

canergen (approved):

canergen:
Hi @martinkim0 and @ori-kron-wis does it have to be merged again(?) after adding on-merge: backport to 1.2.X?

PR #2910, opened 2024-07-25
test(model) fixes errors in testing closeness on GPU

canergen:
Fixes tests.


canergen:
@martinkim0 does this need a label of merge_on_backport ?

PR #2913, opened 2024-07-26
fix(module) TotalVI fix activation background

canergen:
In the totalVI decoder, we were stacking two exponential activation functions. This was producing numerical instabilities. We replace this with a softplus activation function to increase numerical instability. Code is backwards compatible. Default has changed.


canergen:
Please see https://github.com/scverse/scvi-tools/pull/2913 if you still have this issue in our main branch, please create a new issue and share more information about your data (n_protein, n_cells, n_genes)

PR #2936, opened 2024-08-12
feat(train) Add support for categorial covariates in scArches

ori-kron-wis:
closes https://github.com/scverse/scvi-tools/issues/2583


canergen:
Can you please run https://docs.scvi-tools.org/en/1.0.2/tutorials/notebooks/scarches_scvi_tools.html with adding categorical covariate key in addition to the batch key and by replacing batch_key with categorical covariate key and verify that the results are the same?

PR #2938, opened 2024-08-18
Error in scanvi_ref multiclass accuracy calculation

ori-kron-wis:
added -1 to the squeeze in order to deal with the case of dim==1 in a multiclass batch where there is 1 sample for the usage of scvi-> scanvi transfer learning

close https://github.com/scverse/scvi-tools/issues/2934

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,8 @@ to [Semantic Versioning]. Full commit history is available in the
 
 #### Fixed
 
+- Fix multiclass accuracy bug which happend during SemiSupervisedDataSplitter usage of scvi->
+    scanvi transfer learning. there was a batch with 1 sample per one of the classes {pr}`2938`.
 - Disable adversarial classifier if training with a single batch.
     Previously this raised a None error {pr}`2914`.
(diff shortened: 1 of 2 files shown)


canergen:
As far as I understand it. It's not really for the case dim==1 but for the case .shape =(1,1). It doesn't require a comment and needs a CHANGELOG entry.

CHANGELOG.md
canergen, line 1: Please refer to the class for the DataSplitter. And the message should be: 'Fix logging of accuracy for cases with 1 sample per class in scANVI'. I assume the transfer actually doesn't matter.

ori-kron-wis: changed

CHANGELOG.md
```diff
@@ -97,6 +97,8 @@ to [Semantic Versioning]. Full commit history is available in the
 
 #### Fixed
 
+- Fix logging of accuracy for cases with 1 sample per class in scANVI {pr}`2938`.
+      the DataSplitter {pr}`2938`.
```
canergen, line 101:
```suggestion
```

PR #2939, opened 2024-08-18
In 1.1.5 autotune change in RunConfig parameter from local_dir to storage_path

ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/2908

This patch fix is intended to be released as 1.1.6 to meet users request

### DIFF
--- a/docs/release_notes/index.md
+++ b/docs/release_notes/index.md
@@ -39,9 +39,9 @@ is available in the [commit logs](https://github.com/scverse/scvi-tools/commits/
 
 #### Fixed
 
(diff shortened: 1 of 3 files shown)


LevyNat (approved):

docs/release_notes/index.md
canergen, line 1: Remote the not part of 1.2.x please. CHANGELOG is a chronological document and shouldn't refer forward.

canergen: Please add a reference to the issue so it becomes clear why we did it.

PR #2946, opened 2024-08-25
test for pytorch lightning v2.4

ori-kron-wis:


ori-kron-wis:
I also made a DE tutorial diff between this branch build and the latest we had in our release (scvi v1.1.0)
It can be shown here:  https://github.com/scverse/scvi-tutorials/pull/309

PR #2948, opened 2024-08-26
added the future annotations import for many models functions

ori-kron-wis:


src/scvi/external/stereoscope/_model.py
canergen, line 1: Does it work when setting it up with None?

canergen, line 1: Does it work when setting it up with None?

src/scvi/model/_condscvi.py
canergen, line 1: Same here. I think it will fail with None

src/scvi/model/_jaxscvi.py
canergen, line 1: DataLoader for JAX will fail?

src/scvi/model/_amortizedlda.py
```diff
@@ -59,8 +60,8 @@ def __init__(
         adata: AnnData,
         n_topics: int = 20,
         n_hidden: int = 128,
-        cell_topic_prior: Optional[Union[float, Sequence[float]]] = None,
-        topic_feature_prior: Optional[Union[float, Sequence[float]]] = None,
+        cell_topic_prior: float | Sequence[float] = None,
```
canergen, line 63: Has to be float | Sequence[float] | **None**

src/scvi/model/_amortizedlda.py
```diff
@@ -59,8 +60,8 @@ def __init__(
...
+        topic_feature_prior: float | Sequence[float] = None,
```
canergen, line 64: Same.

PR #2951, opened 2024-08-28
fix(model) check duplicated gene in celltype marker table

ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/2837

added assertion error in cellAssign for checking duplicates in celltype markers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ to [Semantic Versioning]. Full commit history is available in the
 
 #### Added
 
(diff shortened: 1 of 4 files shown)


canergen (approved):
We only need the first check (ll 84-86). The second one is handled by AnnData.

canergen:
I think index.has_duplicates is a slightly better check for pandas to exclude duplicated indices. Can you add a test and CHANGELOG?

canergen (approved):

PR #2959, opened 2024-09-04
docs: fix progress bar interactive style in RTD

ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/2958

### DIFF
--- a/src/scvi/data/_download.py
+++ b/src/scvi/data/_download.py
@@ -1,57 +1,19 @@
 import logging
 import os
-import urllib
 from typing import Optional
 
-import numpy as np
-
-from scvi.utils import track
+import scanpy as sc
 
 logger = logging.getLogger(__name__)
 
 
 def _download(url: Optional[str], save_path: str, filename: str):
     """Writes data from url to file."""
-    if os.path.exists(os.path.join(save_path, filename)):
-        logger.info(f"File {os.path.join(save_path, filename)} already downloaded")
+    path_filename = os.path.join(save_path, filename)
+    if os.path.exists(path_filename):
+        logger.info(f"File {path_filename} already downloaded")
         return
     elif url is None:
-        logger.info(f"No backup URL provided for missing file {os.path.join(save_path, filename)}")
+        logger.info(f"No backup URL provided for missing file {path_filename}")
         return
-    req = urllib.request.Request(url, headers={"User-Agent": "Magic Browser"})
-    try:
-        r = urllib.request.urlopen(req)
-        if r.getheader("Content-Length") is None:
-            raise FileNotFoundError(
-                f"Found file with no content at {url}. "
-                "This is possibly a directory rather than a file path."
-            )
-    except urllib.error.HTTPError as exc:
-        if exc.code == "404":
-            raise FileNotFoundError(f"Could not find file at {url}") from exc
-        raise exc
-    logger.info(f"Downloading file at {os.path.join(save_path, filename)}")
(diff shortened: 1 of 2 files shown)


src/scvi/data/_download.py
```diff
@@ -1,57 +1,19 @@
 import logging
 import os
-import urllib
 from typing import Optional
 
-import numpy as np
-
-from scvi.utils import track
+import scanpy as sc
```
canergen, line 5: We have no requirement on scanpy in scvi-tools and would like to keep it (the one that have are optional imports). You are using a private (_XXX) function from scanpy, see the Zulip discussion from last week why this is a bad idea.

src/scvi/utils/_track.py
canergen, line 1: Could we deprecate track all together and only use the lightning track?

src/scvi/data/_download.py
ori-kron-wis: Yes of course. This will be removed once we figure out what works and what doesn't

src/scvi/data/_download.py
```diff
@@ -1,56 +1,91 @@
...
 
-from scvi.utils import track
 
-logger = logging.getLogger(__name__)
+def _download_scanpy_style(url: str, path: Path):
+    try:
+        import ipywidgets  # noqa: F401
+        from tqdm.auto import tqdm
+    except ImportError:
+        from tqdm import tqdm
```
canergen, line 13: What’s this test about? To support old versions can we just bump the requirement?

src/scvi/data/_download.py
```diff
@@ -1,56 +1,91 @@
...
+                while block:
+                    f.write(block)
+                    blocknum += 1
+                    t.update(len(block))
+                    block = resp.read(blocksize)
+
+    except (KeyboardInterrupt, Exception):
+        # Make sure file doesn’t exist half-downloaded
+        if path.is_file():
+            path.unlink()
```
canergen, line 59: Does this remove the file?

src/scvi/utils/_track.py
```diff
@@ -4,7 +4,7 @@
 
 from rich.console import Console
 from rich.progress import track as track_base
-from tqdm import tqdm as tqdm_base
+from tqdm.auto import tqdm as tqdm_base
```
canergen, line 7: This fixes the progress bar for training?

canergen (commented):
LGTM. See minor comments.

canergen:
Is the download copied from scanpy? Should we raise with them that having a developer API for this would be preferred?

src/scvi/utils/_track.py
ori-kron-wis: yep

src/scvi/data/_download.py
canergen: We should remove it as in the other function below we just import from auto. Can you check that the requirement is correct and we always install a tqdm version that has auto?

src/scvi/data/_download.py
```diff
@@ -1,56 +1,91 @@
...
 
-import numpy as np
+logger = logging.getLogger(__name__)
 
-from scvi.utils import track
 
-logger = logging.getLogger(__name__)
+def _download_scanpy_style(url: str, path: Path):
+    try:
+        import ipywidgets  # noqa: F401
```
canergen, line 10: ipywidget is not used. Copy the import to line 4.

PR #2969, opened 2024-09-15
Update pyproject.toml

ori-kron-wis:


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/2967

PR #2982, opened 2024-09-19
test: Fix hugging face create repo test which was failing

ori-kron-wis:
Added a check on whether repo exists before creating it, which failed …previous test

### DIFF
--- a/tests/hub/test_hub_model.py
+++ b/tests/hub/test_hub_model.py
@@ -224,13 +224,15 @@ def test_hub_model_large_training_adata(request, save_path):
 def test_hub_model_create_repo_hf(save_path: str):
     from huggingface_hub import delete_repo, repo_exists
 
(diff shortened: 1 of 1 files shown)


tests/hub/test_hub_model.py
```diff
@@ -222,14 +222,15 @@ def test_hub_model_large_training_adata(request, save_path):
...
-
-    hub_model = prep_scvi_hub_model(save_path)
-    hub_model.push_to_huggingface_hub(
-        "scvi-tools/test-scvi-create",
-        os.environ["HF_API_TOKEN"],
-        repo_create=True,
-    )
+    from huggingface_hub import delete_repo, repo_exists
+
+    if not repo_exists("scvi-tools/test-scvi-create"):
```
canergen, line 227: We should do it repo_exists delete_repo and then execute the real test.

PR #2991, opened 2024-09-23
release: updated v1.2.0 and RTD for py312

ori-kron-wis:


src/scvi/module/_totalvae.py
```diff
@@ -744,10 +744,10 @@ def on_load(self, model: BaseModelClass):
         manager = model.get_anndata_manager(model.adata)
         source_version = manager.registry[_constants._SCVI_VERSION_KEY]
         version_split = source_version.split(".")
-        if int(version_split[0]) >= 1 and int(version_split[1]) >= 2:
+        if int(version_split[0]) >= 1 and int(version_split[1]) >= 3:
```
canergen, line 747: This is wrong. Pre 1.2 we want this activation function. I assume the assert statement was wrong.

ori-kron-wis: this is a patch. please suggest a fix.

canergen (approved):

PR #3007, opened 2024-10-02
Change outlier detection for MRVI to ball admissibility calculation

justjhong:
The outlier detection was mistakenly set to the "ap" (aggregated posterior) thresholding setting when it should have been the "ball" thresholding calculation. 

The only changes in this pr are in the `get_outlier_cell_sample_pairs`, everything else is auto-linted.

### DIFF
--- a/src/scvi/external/mrvi/_model.py
+++ b/src/scvi/external/mrvi/_model.py
@@ -765,7 +765,10 @@ class MRVI(JaxTrainingMixin, BaseModelClass):
         indices: npt.ArrayLike | None = None,
         batch_size: int = 256,
     ) -> Distribution:
-        """Compute the aggregated posterior over the ``u`` latent representations.
+        """Computes the aggregated posterior over the ``u`` latent representations.
+
+        For the specified samples, it computes the aggregated posterior over the ``u`` latent
+        representations. Returns a NumPyro MixtureSameFamily distribution.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


src/scvi/external/mrvi/_model.py
```diff
@@ -995,21 +996,32 @@ def get_outlier_cell_sample_pairs(
...
+                np.expand_dims(adata_s.obsm["U"], ap.mixture_dim)
+            ).sum(axis=1)
+            log_probs_s = rowwise_max_excluding_diagonal(in_max_comp_log_probs)
+
             log_probs_ = []
+            n_splits = adata.n_obs // batch_size
             for u_rep in np.array_split(adata.obsm["U"], n_splits):
-                log_probs_.append(jax.device_get(ap.log_prob(u_rep).sum(-1, keepdims=True)))
+                log_probs_.append(
+                    jax.device_get(
```
PierreBoyeau, line 1008: Could you add a comment to describe the shape of the tensor of probs (inside the `device_get`)? It would make it easier to double check what we are summing/taking the max of.

src/scvi/external/mrvi/_model.py
```diff
@@ -995,21 +996,32 @@ def get_outlier_cell_sample_pairs(
             adata_s = adata[sample_idxs]
 
             ap = self.get_aggregated_posterior(adata=adata, indices=sample_idxs)
```
PierreBoyeau, line 998: Could you quickly update the docstring for this method?
Also, could you remind me what is the purpose of `threshs` again?

PierreBoyeau (approved):
LGTM, the two comments I have is to 1. maybe add an inline command to keep track of the object shape 2. update the docstring!

src/scvi/external/mrvi/_model.py
justjhong: its a threshold that determines if a cell-sample pair is admissible or not

canergen:
Can you also backport to 1.2.x otherwise the fox only becomes active in 1.3.x

justjhong:
@meeseeksdev backport to 1.2.x

PR #3015, opened 2024-10-15
feat: implement Decipher model in external

justjhong:
CC @ANazaret

Implements Decipher model (https://github.com/azizilab/decipher, https://www.biorxiv.org/content/10.1101/2023.11.11.566719v1) into external/

For now, it only includes base implementation without many of the downstream workflows from the original implementation.
Includes minor non-breaking changes to the `LowLevelPyroTrainingPlan`.

Test: was able to approximately reproduce figures from the tutorial (https://github.com/azizilab/decipher/blob/main/examples/1-tutorial.ipynb), some of the v plots for several random seeds below:

Original implementation:
decipher_orig_0
decipher_orig_1030

New implementation:
decipher_tutorial_scvi_0
decipher_tutorial_scvi_1
decipher_tutorial_scvi_2
decipher_tutorial_scvi_3


justjhong:
Not sure why the cuda tests are failing on import of the Decipher class. Any ideas? @canergen @ori-kron-wis

ori-kron-wis:
@justjhong  I updated the main branch with the updated cuda test file we need now to run on WIS servers and merged this PR with it. Now the tests work. 
However, test_decipher is failing. Not sure why, I can only say it is passing in a non-GPU environment, and failing in both single and multi GPU environments. Can it be related to the fact that it is Pyro based?

justjhong:
@ori-kron-wis thanks for your help confirming its a real bug. I will debug this on my local CUDA machine and update the PR accordingly

ori-kron-wis:
you are right, it is still not solved. I will have to take another look

justjhong:
@ori-kron-wis Haha these flaky things are never satisfying but thank you for finding the fix! I will add decipher to the docs first before asking you for the final review.

Aside: do you think there should be a simple PR checklist for the standard things for typical contributions? I keep forgetting things (e.g., CHANGELOG, function level documentation, add classes to docs if necessary, add backport label if needed before merge)

justjhong:
@ori-kron-wis once checks pass, this PR should be ready for review and merge 👍

justjhong:
@ori-kron-wis cuda tests failing after merging with main. do you mind taking a look before the pr gets out of date again?

justjhong:
Thanks @ori-kron-wis ! if everything looks can you approve the PR? @canergen

justjhong:
@ori-kron-wis @canergen bump on this PR

.github/workflows/test_linux_cuda.yml
```diff
@@ -48,6 +48,8 @@ jobs:
       #  run: echo "BRANCH_NAME=$(echo $GITHUB_REF | awk -F'/' '{print $3}')" >> $GITHUB_ENV
 
       - uses: actions/checkout@v4
+        with:
```
canergen, line 51: What's this?

src/scvi/external/decipher/_components.py
```diff
@@ -0,0 +1,102 @@
...
+        self.context_dim = context_dim
+        self.hidden_dims = hidden_dims
+        self.output_dims = output_dims
+        self.deep_context_injection = deep_context_injection
+        self.n_output_layers = len(self.output_dims)
+        self.output_total_dim = sum(self.output_dims)
+
+        # The multiple outputs are computed as a single output layer, and then split
+        indices = np.concatenate(([0], np.cumsum(self.output_dims)))
+        self.output_slices = [slice(s, e) for s, e in zip(indices[:-1], indices[1:], strict=False)]
```
canergen, line 53: Can you explain?

src/scvi/external/decipher/_components.py
```diff
@@ -0,0 +1,102 @@
...
+
+            layers.append(
+                torch.nn.Linear(hidden_dims[-1] + deep_context_dim, self.output_total_dim)
+            )
+        else:
+            layers.append(torch.nn.Linear(input_dim + context_dim, self.output_total_dim))
+
+        self.layers = torch.nn.ModuleList(layers)
+
+        self.f = activation
```
canergen, line 76: Please call it self.activation_fn

src/scvi/external/decipher/_components.py
```diff
@@ -0,0 +1,102 @@
...
+                h = torch.cat([context, h], dim=-1)
+            h = layer(h)
+            if i < len(self.layers) - 1:
+                h = self.batch_norms[i](h)
+                h = self.f(h)
+
+        if self.n_output_layers == 1:
+            return h
+        else:
+            h = h.reshape(list(x.size()[:-1]) + [self.output_total_dim])
```
canergen, line 96: Comment this line.

src/scvi/external/decipher/_model.py
```diff
@@ -0,0 +1,153 @@
...
+class Decipher(PyroSviTrainMixin, BaseModelClass):
+    """Decipher model for single-cell data analysis :cite:p:`Nazaret23`.
+
+    Parameters
+    ----------
+    adata
+        AnnData object that has been registered via
+        :meth:`~scvi.model.Decipher.setup_anndata`.
+    dim_v
+        Dimension of the interpretable latent space v.
```
canergen, line 31: better word than interpretable?

src/scvi/external/decipher/_model.py
```diff
@@ -0,0 +1,153 @@
...
+        training_plan: PyroTrainingPlan | None = None,
+        datasplitter_kwargs: dict | None = None,
+        plan_kwargs: dict | None = None,
+        **trainer_kwargs,
+    ):
+        if "early_stopping_monitor" not in trainer_kwargs:
+            trainer_kwargs["early_stopping_monitor"] = "nll_validation"
+        datasplitter_kwargs = datasplitter_kwargs or {}
+        if "drop_last" not in datasplitter_kwargs:
+            datasplitter_kwargs["drop_last"] = True
```
canergen, line 101: Do you feel strong about this default?

canergen: This is not a remark to fix but just curiosity.

src/scvi/external/decipher/_components.py
```diff
@@ -0,0 +1,102 @@
...
+                )
+                batch_norms.append(nn.BatchNorm1d(hidden_dims[i]))
+
+            layers.append(
+                torch.nn.Linear(hidden_dims[-1] + deep_context_dim, self.output_total_dim)
+            )
+        else:
+            layers.append(torch.nn.Linear(input_dim + context_dim, self.output_total_dim))
+
+        self.layers = torch.nn.ModuleList(layers)
```
canergen, line 74: Is there any advantage here from ModuleList over just the list?

src/scvi/external/decipher/_module.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+
+        self.theta = None
+
+        self._epsilon = 1e-5
+
+        self.n_obs = None  # Populated by PyroTrainingPlan
+
+        # Hack: to allow auto_move_data to infer device.
+        self._dummy_param = nn.Parameter(torch.empty(0))
```
canergen, line 74: Can you expand on this?

canergen: It's likely because you don't have a single parameter?

src/scvi/external/decipher/_module.py
```diff
@@ -0,0 +1,179 @@
...
+        x = tensor_dict[REGISTRY_KEYS.X_KEY]
+        return (x,), {}
+
+    @auto_move_data
+    def model(self, x: torch.Tensor):
+        pyro.module("decipher", self)
+
+        self.theta = pyro.param(
+            "theta",
+            x.new_ones(self.dim_genes),
```
canergen, line 91: You could set your dummy to be this one.

src/scvi/external/decipher/_module.py
```diff
@@ -0,0 +1,179 @@
...
+            z_loc, z_scale = self.encoder_x_to_z(x)
+            z_scale = softplus(z_scale)
+            posterior_z = dist.Normal(z_loc, z_scale).to_event(1)
+            z = pyro.sample("z", posterior_z)
+
+            zx = torch.cat([z, x], dim=-1)
+            v_loc, v_scale = self.encoder_zx_to_v(zx)
+            v_scale = softplus(v_scale)
+            with poutine.scale(scale=self.beta):
+                posterior_v = dist.Normal(v_loc, v_scale).to_event(1)
```
canergen, line 138: No warm up?

src/scvi/external/decipher/_module.py
```diff
@@ -0,0 +1,179 @@
...
+            Number of passes through the dataloader (default is 5).
+
+        Returns
+        -------
+        float
+            The average estimated predictive log-likelihood across multiple runs.
+        """
+        log_weights = []
+        old_beta = self.beta
+        self.beta = 1.0
```
canergen, line 166: why not make beta an optional argument to model and guide and overwrite it then. This is cleaner.

canergen: Why do you need a try block here?

src/scvi/external/decipher/_trainingplan.py
```diff
@@ -0,0 +1,143 @@
...
+        # See configure_optimizers for what this does
+        self._dummy_param = torch.nn.Parameter(torch.Tensor([0.0]))
+
+    def on_validation_model_train(self):
+        """Prepare the model for validation by switching to train mode."""
+        super().on_validation_model_train()
+        if self.current_epoch > 0:
+            # freeze the batch norm layers after the first epoch
+            # 1) the batch norm layers helps with the initialization
+            # 2) but then, they seem to imply a strong normal prior on the latent space
```
canergen, line 65: LayerNorm also has this problem? Any literature that does something similar?

src/scvi/external/decipher/_trainingplan.py
```diff
@@ -0,0 +1,143 @@
...
+        PyTorch Lightning wants to take steps on an optimizer
+        returned by this function in order to increment the global
+        step count. See PyTorch Lighinting optimizer manual loop.
+
+        Here we provide a shim optimizer that we can take steps on
+        at minimal computational cost in order to keep Lightning happy :).
+        """
+        return torch.optim.Adam([self._dummy_param])
+
+    def optimizer_step(self, *args, **kwargs):
```
canergen, line 139: Are these missing from LowLevelPyroTrainingPlan? Should we add them there?

src/scvi/external/decipher/_module.py
```diff
@@ -0,0 +1,179 @@
...
+        self.layers_z_to_x = layers_z_to_x
+        self.beta = beta
+
+        self.decoder_v_to_z = ConditionalDenseNN(dim_v, layers_v_to_z, [dim_z] * 2)
+        self.decoder_z_to_x = ConditionalDenseNN(dim_z, layers_z_to_x, [dim_genes])
+        self.encoder_x_to_z = ConditionalDenseNN(dim_genes, [128], [dim_z] * 2)
+        self.encoder_zx_to_v = ConditionalDenseNN(
+            dim_genes + dim_z,
+            [128],
+            [dim_v, dim_v],
```
canergen, line 64: 2 * [dim_v]

canergen: Is dim_v fixed or do you use other values than 2?

.github/workflows/test_linux_cuda.yml
justjhong: Not sure, I believe @ori-kron-wis made this change while debugging the tests. Could you explain Ori?

src/scvi/external/decipher/_components.py
justjhong: this produces slices along the output, which will just be a concatenated version of all outputs needed, so that the output can be separated by the provided input_dims. It's a bit tricky, but I think the comment above serves the purpose

src/scvi/external/decipher/_model.py
justjhong: Unfortunately not, since it captures the same info as z, just meant to be more directly interpretable

src/scvi/external/decipher/_model.py
justjhong: I believe there were bugs with batch norm when thiswasn't there and the final batch was of size 1

src/scvi/external/decipher/_components.py
justjhong: Module list ensures the element parameters are properly registered. If they are just in a list and the modules arent direct parameters of the nn.Module class I think it doesnt get picked up by the optimizer

src/scvi/external/decipher/_module.py
justjhong: I think it's a relatively clean way to make automove work for pyro modules. The pyro params don't get stored in the same way as normal torch parameters, which is why auto_move_breaks if there are no actual parameters on the class. We do the same thing for the other pyro modules in the repo (e.g., https://github.com/scverse/scvi-tools/blob/main/src/scvi/module/_amortizedlda.py#L95)

src/scvi/external/decipher/_module.py
justjhong: pyro.param doesn't actually create a torch parameter logged in module.parameters()

src/scvi/external/decipher/_module.py
justjhong: great idea, implemented this

src/scvi/external/decipher/_trainingplan.py
justjhong: not actually sure, this was copied over from original impl, will ask

src/scvi/external/decipher/_trainingplan.py
justjhong: these are only in the PyroTrainingPlan, not the LowLevelPyroTrainingPlan. Whether we want to push it back into the low-level class is up to you

src/scvi/external/decipher/_module.py
justjhong: It can be changed, but in practice it is always set to 2 for the sake of visualization. I think we can leave it as a hyperparameter here since dim_v > 2 is potentially useful.

canergen: Please change to [dim_v] * 2 for consistency.

.github/workflows/test_linux_cuda.yml
ori-kron-wis: yeah I wanted to see if the cuda test error we had was somehow related to the commit history depth, so I tried it in the branch. But eventually it was about cleaning local caches + updating packages on the linux machine.
we can revert this

src/scvi/external/decipher/_module.py
justjhong: this was addressed in the latest commit

canergen (approved):
Please provide more information on what context means and how to use it in the user guide.

PR #3030, opened 2024-10-24
Update _base_components.py

canergen:
Add small offset to scale in totalVI decoder


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3028

PR #3032, opened 2024-10-27
feat: continuous, embedded covariates re-injected during training

ori-kron-wis:
Added other covariates types (continuous, embedded) to be able to reinjected, and not just to the input layer


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3021

src/scvi/nn/_base_components.py
```diff
@@ -15,7 +15,13 @@ def _identity(x):
 
 
 class FCLayers(nn.Module):
-    """A helper class to build fully-connected layers for a neural network.
+    """FCLayers class of scvi-tools adapted to also inject continous covariates.
+
+    The only adaptation is addition of `n_cont` parameter in init and `cont` in forward,
```
Hrovatin, line 20: Please remove this docstring part  - I only added it as info for implementing the change.

src/scvi/nn/_base_components.py
```diff
@@ -135,13 +145,18 @@ def _hook_fn_zero_out(grad):
                     b = layer.bias.register_hook(_hook_fn_zero_out)
                     self.hooks.append(b)
 
-    def forward(self, x: torch.Tensor, *cat_list: int):
+    def forward(
+        self, x: torch.Tensor, cat_list: list | None = None, cont: torch.Tensor | None = None
```
Hrovatin, line 149: I changed the forward parametrization from `(x, *cat_list)` to `(x,cat_list,cont)` - so this will break with other `Modules` using `FCLayers` by passing `cat_list` in expanded `*` format. So either the parametrization here needs to be fixed or the use of `FCLayers` elsewhere.

Hrovatin (changes requested):

src/scvi/nn/_base_components.py
Hrovatin: See my comment in https://github.com/scverse/scvi-tools/issues/3021

ori-kron-wis: Thanks. I thought I would be able to finish this in one shot for all models, but its a bit more complex than I thought. So, I'll be focusing on scvi and scanvi for now only.

ori-kron-wis:
This kind of change requires touching the module for each model. I don't think there is anywhere around it. Its risky and this is why I implemented it only for scvi &  scanvi at the moment (maybe others will work also but I didn't test thoroughly). the rest of the changed modules are mainly things in function headers and placeholders to able to cope with the major changes in base_components (i.e cont_cov is usually None and cant be injected to their hidden layers). 

To summarise we have 2 parameters to think of:
- deeply_inject (bool): covariates (continuous or categorial whether being encoded or not) are always injected to the first layer. If this parameter is True they will also be injected to the hidden layer of the encoder/decoder.
- encode_covariates (bool): this is only relevant for the categorial covariates. If this parameter is false, they will be injected as one-hot, if true, they will be encoded as an embedded matrix (this is dependent on other meta-parameters such as the size of embedded vector and the batch_representation which should be "embedded" as well if we want it, otherwise it will remain one-hot)

@Hrovatin please try to use this branch

src/scvi/module/_vae.py
```diff
@@ -372,17 +382,17 @@ def _regular_inference(
         if self.batch_representation == "embedding" and self.encode_covariates:
             batch_rep = self.compute_embedding(REGISTRY_KEYS.BATCH_KEY, batch_index)
             encoder_input = torch.cat([encoder_input, batch_rep], dim=-1)
-            qz, z = self.z_encoder(encoder_input, *categorical_input)
+            qz, z = self.z_encoder(encoder_input, cont_covs, *categorical_input)
```
Hrovatin, line 385: I think here batch_rep should be also concat to cont_covs to be treated as continuous covariate no?

src/scvi/module/_vae.py
```diff
@@ -360,7 +370,7 @@ def _regular_inference(
         if self.log_variational:
             x_ = torch.log1p(x_)
 
-        if cont_covs is not None and self.encode_covariates:
+        if cont_covs is not None:
             encoder_input = torch.cat((x_, cont_covs), dim=-1)
```
Hrovatin, line 374: Why do you always concatenate continous cov to expression instead of handling them separately as a covariate (sam as for one-hot)?

src/scvi/nn/_base_components.py
```diff
@@ -45,13 +47,26 @@ class FCLayers(nn.Module):
         Whether to inject covariates in each layer, or just the first (default).
     activation_fn
         Which activation function to use
+    encode_covariates
```
Hrovatin, line 50: This is not used anywhere?

Hrovatin: `encode_covariates` and `batch_representation` are used in Module (e.g. vae) for handling of covariates not here

src/scvi/nn/_base_components.py
```diff
@@ -61,26 +76,30 @@ def __init__(
...
     ):
         super().__init__()
         self.inject_covariates = inject_covariates
+        self.batch_representation = batch_representation
+        self.encode_covariates = encode_covariates
         layers_dim = [n_in] + (n_layers - 1) * [n_hidden] + [n_out]
 
-        if n_cat_list is not None:
-            # n_cat = 1 will be ignored
+        if n_cat_list is not None and self.batch_representation == "one-hot":
```
Hrovatin, line 88: I do not think you should check `batch_representation` here - if `n_cat_list` is not empty the covariates need to be accounted for here irrespective of what `batch_representation`  is - as the `batch_representation` is used in Module to determine if batch will be added to the cat list or not. So the input cat cov list will already contain batch if it is not embedded ,besides other cat covariates

src/scvi/nn/_base_components.py
```diff
@@ -45,13 +47,26 @@ class FCLayers(nn.Module):
         Whether to inject covariates in each layer, or just the first (default).
     activation_fn
         Which activation function to use
+    encode_covariates
+        If ``True``, covariates are concatenated to gene expression prior to passing through
+        the encoder(s). Else, only gene expression is used.
+    batch_representation
```
Hrovatin, line 53: This should be also removed and only kept in Module

src/scvi/nn/_base_components.py
```diff
@@ -173,14 +194,18 @@ def forward(self, x: torch.Tensor, *cat_list: int):
...
-                                    o.unsqueeze(0).expand((x.size(0), o.size(0), o.size(1)))
-                                    for o in one_hot_cat_list
-                                ]
-                            else:
-                                one_hot_cat_list_layer = one_hot_cat_list
-                            x = torch.cat((x, *one_hot_cat_list_layer), dim=-1)
+                            if i > 0 and self.inject_covariates and cont_covs is not None:
+                                # Need to inject the continous covariates to hidden layers
+                                x = torch.cat((x, cont_covs), dim=-1)
+                            if self.batch_representation == "one-hot":
```
Hrovatin, line 200: this should be done irregardless of what is batch representation for all covariates that were passed as categorical

Hrovatin (changes requested):

Hrovatin:
@ori-kron-wis I think there are things that need to be fixed as per my review comments. To summarise:
I think the behaviour should be as follows:
- Module passes to layers expression, continuous covariates (can be None), and categorical covariates that need to be one-hot encoded (can be empty list)
- Module is the one that decides if batch will be embedded and concatenated to continuous covariates or added to the categorical list
- Layers should not check how the batch is encoded as their init params already tell the number of continous/categorical covariates as determined by the Module.
- The layers then either inject covariates (continuous and one-hote encoded categorical) into all layers or only add to the first one. - So module should not be concatenating continuous covariates to expression.

src/scvi/external/methylvi/_module.py
```diff
@@ -163,7 +163,7 @@ def inference(self, mc, cov, batch_index, cat_covs=None, n_samples=1):
         else:
             categorical_input = ()
 
-        qz, z = self.z_encoder(methylation_input, batch_index, *categorical_input)
+        qz, z = self.z_encoder(methylation_input, None, batch_index, *categorical_input)
```
canergen, line 166: Can you add the None as an argument in line 148. It makes it cleaner where it comes from.

canergen: This is throughout this PR.

src/scvi/module/_multivae.py
```diff
@@ -51,9 +53,9 @@ def __init__(
         )
         self.output = torch.nn.Sequential(torch.nn.Linear(n_hidden, 1), torch.nn.LeakyReLU())
 
-    def forward(self, x: torch.Tensor, *cat_list: int):
+    def forward(self, x: torch.Tensor, cont_covs: torch.Tensor | None = None, *cat_list: int):
```
canergen, line 56: Here it's correct.

src/scvi/module/_multivae.py
```diff
@@ -581,7 +604,7 @@ def inference(
         mask_acc = x_chr.sum(dim=1) > 0
         mask_pro = y.sum(dim=1) > 0
 
-        if cont_covs is not None and self.encode_covariates:
+        if cont_covs is not None:
```
canergen, line 607: We shouldn't remove the self.encode covariates check.

src/scvi/module/_multivae.py
```diff
@@ -597,21 +620,21 @@ def inference(
 
         # Z Encoders
         qzm_acc, qzv_acc, z_acc = self.z_encoder_accessibility(
-            encoder_input_accessibility, batch_index, *categorical_input
+            encoder_input_accessibility, None, batch_index, *categorical_input
```
canergen, line 623: why is this hard-coded?

src/scvi/module/_peakvae.py
```diff
@@ -185,6 +189,7 @@ def __init__(
             n_output=self.n_latent,
             n_hidden=self.n_hidden,
             n_cat_list=encoder_cat_list,
+            n_cont=n_continuous_cov,
```
canergen, line 192: only if encode covariates

src/scvi/module/_scanvae.py
```diff
@@ -298,18 +312,31 @@ def loss(
         z1: torch.Tensor = inference_outputs["z"]
         x: torch.Tensor = tensors[REGISTRY_KEYS.X_KEY]
         batch_index: torch.Tensor = tensors[REGISTRY_KEYS.BATCH_KEY]
+        if "extra_continuous_covs" in tensors.keys():
+            cont_covs: torch.Tensor = tensors[REGISTRY_KEYS.CONT_COVS_KEY]
+            cont_covs = broadcast_labels(cont_covs, n_broadcast=self.n_labels)[1]
+        else:
+            cont_covs = None
```
canergen, line 319: please explain.

src/scvi/module/_scanvae.py
```diff
@@ -298,18 +312,31 @@ def loss(
...
 
         ys, z1s = broadcast_labels(z1, n_broadcast=self.n_labels)
-        qz2, z2 = self.encoder_z2_z1(z1s, ys)
-        pz1_m, pz1_v = self.decoder_z1_z2(z2, ys)
+        qz2, z2 = self.encoder_z2_z1(
+            torch.cat((z1s, cont_covs), dim=-1) if cont_covs is not None else z1s, cont_covs, ys
```
canergen, line 323: no this is independent of cont_covs

src/scvi/module/_scanvae.py
```diff
@@ -298,18 +312,31 @@ def loss(
...
+        )
+        pz1_m, pz1_v = self.decoder_z1_z2(
+            torch.cat((z2, cont_covs), dim=-1) if cont_covs is not None else z2, cont_covs, ys
```
canergen, line 326: this to independent of cont_covs

src/scvi/module/_vae.py
canergen: I would prefer to keep all seperated. So you could do deep injection of the batch embedding or covariate.

Hrovatin: That also sounds good as loong as all can be injected

PR #3036, opened 2024-10-31
fix(train): in case of last batch <=2, move to validation if possible

ori-kron-wis:
In case that train_size is None and the size of the last batch during training is <=2 , we adaptively move those samples from training to validation if possible. If train_size is set by user we do no fix this error and let the user change its train_size, selected indices or use drop last batch option.
close https://github.com/scverse/scvi-tools/issues/3035

### DIFF
--- a/src/scvi/dataloaders/_data_splitting.py
+++ b/src/scvi/dataloaders/_data_splitting.py
@@ -166,9 +166,7 @@ def validate_data_split_with_external_indexing(
 
     if batch_size is not None:
         num_of_cells = n_train % batch_size
-        if (num_of_cells < 3 and num_of_cells > 0) and not (
-            num_of_cells == 1 and drop_last is True
-        ):
+        if (num_of_cells < 3 and num_of_cells > 0) and drop_last is False:
             warnings.warn(
                 f"Last batch will have a small size of {num_of_cells} "
                 f"samples. Consider changing settings.batch_size or batch_size in model.train "
--- a/src/scvi/dataloaders/_data_splitting.py
+++ b/src/scvi/dataloaders/_data_splitting.py
@@ -69,15 +69,12 @@ def validate_data_split(
 
     if batch_size is not None:
         num_of_cells = n_train % batch_size
-        if (num_of_cells < 3 and num_of_cells > 0) and not (
-            num_of_cells == 1 and drop_last is True
-        ):
+        if (num_of_cells < 3 and num_of_cells > 0) and drop_last is False:
             if not train_size_is_none:
                 warnings.warn(
                     f"Last batch will have a small size of {num_of_cells} "
                     f"samples. Consider changing settings.batch_size or batch_size in model.train "
-                    f"from currently {batch_size} to avoid errors during model training, "
-                    f"or use drop_last parameter if there is 1 cell left",
+                    f"from currently {batch_size} to avoid errors during model training.",
                     UserWarning,
                     stacklevel=settings.warnings_stacklevel,
                 )
@@ -86,7 +83,8 @@ def validate_data_split(
                 if n_val > 0:
                     n_val += num_of_cells
                     warnings.warn(
-                        f"{num_of_cells} cells moved from training set to validation set",
+                        f"{num_of_cells} cells moved from training set to validation set."
+                        f" if you want to avoid it please use train_size parameter during train.",
                         UserWarning,
                         stacklevel=settings.warnings_stacklevel,
                     )
@@ -175,8 +173,7 @@ def validate_data_split_with_external_indexing(
                 f"Last batch will have a small size of {num_of_cells} "
                 f"samples. Consider changing settings.batch_size or batch_size in model.train "
                 f"from currently {settings.batch_size} to avoid errors during model training "
-                f"or change the given external indices accordingly or use drop_last parameter if "
-                f"there is 1 cell left",
+                f"or change the given external indices accordingly.",
                 UserWarning,
                 stacklevel=settings.warnings_stacklevel,
             )
--- a/src/scvi/dataloaders/_data_splitting.py
+++ b/src/scvi/dataloaders/_data_splitting.py
@@ -163,8 +163,8 @@ def validate_data_split_with_external_indexing(
         if n_train % batch_size < 3 and n_train % batch_size > 0:
             warnings.warn(
                 f"Last batch will have a small size of {n_train % batch_size} "
-                f"samples. Consider changing settings.batch_size or batch_size in model.train from "
-                f"currently {settings.batch_size} to avoid errors during model training "
+                f"samples. Consider changing settings.batch_size or batch_size in model.train "
+                f"from currently {settings.batch_size} to avoid errors during model training "
                 f"or change the given external indices accordingly",
                 UserWarning,
                 stacklevel=settings.warnings_stacklevel,
--- a/src/scvi/dataloaders/_data_splitting.py
+++ b/src/scvi/dataloaders/_data_splitting.py
@@ -243,7 +243,7 @@ class DataSplitter(pl.LightningDataModule):
             self.n_train, self.n_val = validate_data_split_with_external_indexing(
                 self.adata_manager.adata.n_obs,
                 self.external_indexing,
-                self.data_loader_kwargs["batch_size"],
+                self.data_loader_kwargs.pop("batch_size", settings.batch_size),
                 self.drop_last,
             )
         else:
@@ -251,7 +251,7 @@ class DataSplitter(pl.LightningDataModule):
                 self.adata_manager.adata.n_obs,
                 self.train_size,
                 self.validation_size,
-                self.data_loader_kwargs["batch_size"],
+                self.data_loader_kwargs.pop("batch_size", settings.batch_size),
                 self.drop_last,
             )
 
@@ -423,7 +423,7 @@ class SemiSupervisedDataSplitter(pl.LightningDataModule):
                 n_labeled_train, n_labeled_val = validate_data_split_with_external_indexing(
                     n_labeled_idx,
                     [labeled_idx_train, labeled_idx_val, labeled_idx_test],
-                    self.data_loader_kwargs["batch_size"],
+                    self.data_loader_kwargs.pop("batch_size", settings.batch_size),
                     self.drop_last,
                 )
             else:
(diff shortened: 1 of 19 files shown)


src/scvi/dataloaders/_data_splitting.py
```diff
@@ -132,6 +159,17 @@ def validate_data_split_with_external_indexing(
     n_train = len(external_indexing[0])
     n_val = len(external_indexing[1])
 
+    if batch_size is not None and not drop_last:
```
canergen, line 162: No need to. If the user provides indices it should just break.

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -205,10 +243,16 @@ def __init__(
             self.n_train, self.n_val = validate_data_split_with_external_indexing(
                 self.adata_manager.adata.n_obs,
                 self.external_indexing,
+                self.data_loader_kwargs.pop("batch_size", settings.batch_size),
```
canergen, line 246: .get is preferable over pop as it keeps the dictionary unchanged.

tests/model/test_scvi.py
```diff
@@ -475,12 +475,20 @@ def test_scvi_n_obs_error(n_latent: int = 5):
     adata = adata[0:129].copy()
     SCVI.setup_anndata(adata)
     model = SCVI(adata, n_latent=n_latent)
-    with pytest.raises(ValueError):
+    with pytest.warns(UserWarning):
         model.train(1, train_size=1.0)
```
canergen, line 479: train_size=1.0 should lead to exactly this train size. There needs to be an option to fix train_size. None could lead to train_size 0.9 if possible and otherwise adjust it.

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: yeah, for some reason get didnt work here

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -183,7 +164,7 @@ class DataSplitter(pl.LightningDataModule):
     adata_manager
         :class:`~scvi.data.AnnDataManager` object that has been created via ``setup_anndata``.
     train_size
-        float, or None (default is 0.9)
+        float, or None (default is None, which is practicaly 0.9 + adaptive last batch)
```
canergen, line 169: I.e. 0.9 and potentially adding small last batch to validation cells

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -230,7 +211,8 @@ def __init__(
     ):
         super().__init__()
         self.adata_manager = adata_manager
-        self.train_size = float(train_size)
+        self.train_size_was_none = not bool(train_size)
```
canergen, line 216: explicit (train_size is None)

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -64,14 +61,15 @@ def validate_data_split(
             "any of the aforementioned parameters."
         )
 
-    if batch_size is not None and not drop_last:
-        if n_train % batch_size < 3 and n_train % batch_size > 0:
-            num_of_cells = n_train % batch_size
+    if batch_size_for_adaptive_last_batch is not None:
```
canergen, line 64: Do we need this if drop_last is true?

canergen: I'm not sure how drop_last is handled now?

src/scvi/dataloaders/_data_splitting.py
canergen: typo and I think the comment is not very clear.

src/scvi/dataloaders/_data_splitting.py
canergen: Warning should still be printed, if train size is not None.

canergen: Best case you change the prodecure here. Keep batch_size and add another argument, adapt_train_size to highlight this use case.

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -564,7 +541,7 @@ class DeviceBackedDataSplitter(DataSplitter):
     adata_manager
         :class:`~scvi.data.AnnDataManager` object that has been created via ``setup_anndata``.
     train_size
-        float, or None (default is 0.9)
+        float, or None (default is None, which is practicaly 0.9 + adaptive last batch)
```
canergen, line 549: Change documentation as above.

src/scvi/external/contrastivevi/_contrastive_data_splitting.py
```diff
@@ -81,15 +81,15 @@ def __init__(
...
+                self.data_loader_kwargs.pop("batch_size", settings.batch_size) if
+                self.train_size_was_none and not self.drop_last else None,
             )
             self.n_target_train, self.n_target_val = validate_data_split(
                 self.n_target,
                 self.train_size,
                 self.validation_size,
-                self.data_loader_kwargs.pop("batch_size", settings.batch_size),
-                self.drop_last,
+                self.data_loader_kwargs.pop("batch_size", settings.batch_size) if
```
canergen: Are we sure that the pop is ok and it still uses the correct batch size afterward?

ori-kron-wis: pop is what we have now in the main branch. tests are ok.

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -59,16 +63,41 @@ def validate_data_split(n_samples: int, train_size: float, validation_size: floa
...
             f"With n_samples={n_samples}, train_size={train_size} and "
-            f"validation_size={validation_size}, the resulting train set will be empty. Adjust"
+            f"validation_size={validation_size}, the resulting train set will be empty. Adjust "
             "any of the aforementioned parameters."
         )
 
+    if batch_size is not None:
+        num_of_cells = n_train % batch_size
+        if (num_of_cells < 3 and num_of_cells > 0) and not (
+            num_of_cells == 1 and drop_last is True
```
canergen, line 73: This one if confusing if drop_last it will drop the last batch no matter how many cells it contains.

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -59,16 +63,41 @@ def validate_data_split(n_samples: int, train_size: float, validation_size: floa
...
+                f"or use drop_last parameter if there is 1 cell left",
+                UserWarning,
+                stacklevel=settings.warnings_stacklevel,
+            )
+            if train_size_is_none:
+                n_train -= num_of_cells
+                if n_val > 0:
+                    n_val += num_of_cells
+                    warnings.warn(
+                        f"{num_of_cells} cells moved from training set to validation set",
```
canergen, line 88: to avoid error during training. Set train_size to a fixed size to avoid this.

canergen (approved):

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -59,16 +63,41 @@ def validate_data_split(n_samples: int, train_size: float, validation_size: floa
...
+    if batch_size is not None:
+        num_of_cells = n_train % batch_size
+        if (num_of_cells < 3 and num_of_cells > 0) and not (
+            num_of_cells == 1 and drop_last is True
+        ):
+            warnings.warn(
+                f"Last batch will have a small size of {num_of_cells} "
+                f"samples. Consider changing settings.batch_size or batch_size in model.train "
+                f"from currently {batch_size} to avoid errors during model training, "
+                f"or use drop_last parameter if there is 1 cell left",
```
canergen, line 79: this one is wrong

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -132,6 +165,21 @@ def validate_data_split_with_external_indexing(
...
+        num_of_cells = n_train % batch_size
+        if (num_of_cells < 3 and num_of_cells > 0) and not (
+            num_of_cells == 1 and drop_last is True
+        ):
+            warnings.warn(
+                f"Last batch will have a small size of {num_of_cells} "
+                f"samples. Consider changing settings.batch_size or batch_size in model.train "
+                f"from currently {settings.batch_size} to avoid errors during model training "
+                f"or change the given external indices accordingly or use drop_last parameter if "
+                f"there is 1 cell left",
```
canergen, line 178: this one to

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: True. this part is not about the drop_last logic.

this part of code is about when to show the warning. and a warning will be shown if we do have 1-2 cells in last batch but also when user didnt select drop_last with cell==1. whoever did it, doesnt need to see this warning, because it will not fail for him

only if also train_size_is_none there will be adaptive cell transferring to the validation set, if exists

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: whats wrong?

src/scvi/dataloaders/_data_splitting.py
canergen: Again the check doesn't need to be num_of_cells == 1 and drop_last is True. It should be: (num_of_cells < 3 and num_of_cells > 0) and drop_last is False

src/scvi/dataloaders/_data_splitting.py
canergen: I think we want a message how to avound moving cells from training to validation. The user avoids it when stating train_size=0.9

src/scvi/dataloaders/_data_splitting.py
canergen: This is confusing. See comment above - line 179 should be gone.

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: ok

src/scvi/dataloaders/_data_splitting.py
ori-kron-wis: yes

src/scvi/dataloaders/_data_splitting.py
```diff
@@ -132,6 +164,20 @@ def validate_data_split_with_external_indexing(
     n_train = len(external_indexing[0])
     n_val = len(external_indexing[1])
 
+    if batch_size is not None:
+        num_of_cells = n_train % batch_size
+        if (num_of_cells < 3 and num_of_cells > 0) and not (
```
canergen, line 169: Update drop_last here.

PR #3045, opened 2024-11-18
feat(external): adding batch_key and labels_key

mys721tx:
Adds batch_key and labels_key per documentation. Also corrected the reference to scvi.external.

### DIFF
--- a/tests/external/scar/test_scar.py
+++ b/tests/external/scar/test_scar.py
@@ -7,8 +7,11 @@ from scvi.external import SCAR
 def test_scar():
     n_latent = 5
     adata = synthetic_iid()
+
+    adata.obs["batch"] = adata.obs["batch"].cat.rename_categories(["batch_2", "batch_3"])
+
     adata.X = scipy.sparse.csr_matrix(adata.X)
-    SCAR.setup_anndata(adata)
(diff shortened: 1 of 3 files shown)


canergen (changes requested):
Thanks for the contribution. LGTM. Please add a release note and a test for both fields.

tests/external/scar/test_scar.py
```diff
@@ -7,8 +7,11 @@
 def test_scar():
     n_latent = 5
     adata = synthetic_iid()
+
+    adata.obs["batch"] = adata.obs["batch"].cat.rename_categories(["batch_2", "batch_3"])
+
     adata.X = scipy.sparse.csr_matrix(adata.X)
-    SCAR.setup_anndata(adata)
+    SCAR.setup_anndata(adata, batch_key="batch")
```
canergen, line 14: Please add label_key="batch". It's good to verify that handling label_key is not broken in the future.

mys721tx: Looks like some other test cases the renaming is not strictly necessary. Should we keep that?

mys721tx (commented):
Both `batch_key` and `labels_key` are added to the test case.

PR #3092, opened 2024-12-13
Add Decipher Trajectory methods

justjhong:
Implements the methods that require a pre-computed trajectory onto the Decipher model class.

cc @ANazaret


src/scvi/external/decipher/_model.py
```diff
@@ -1,11 +1,11 @@
-import logging
 from collections.abc import Sequence
 
 import numpy as np
 import pyro
 import torch
 import torch.nn.functional as F
 from anndata import AnnData
+from sklearn.neighbors import KNeighborsRegressor
```
canergen, line 8: We don't have a dependency of sklearn? Please use lazy import.

src/scvi/external/decipher/utils/__init__.py
```diff
@@ -1,3 +1,4 @@
-from ._utils import rotate_decipher_components
+from ._rotate import rotate_decipher_components
```
canergen, line 1: Im confused about the rotate file? It doesn't exist right?

canergen:
LGTM. See two point above, fix pre-commit missing empty line and add CHANGELOG.

src/scvi/external/decipher/utils/__init__.py
justjhong: it exists within the utils folder

PR #3095, opened 2024-12-16
fix: Explicity use weights_only=False in torch.load functions

ori-kron-wis:
Added weights_only=False for torch load functions for future proof of error message from torch


canergen (approved):
LGTM

PR #3100, opened 2024-12-17
feat: Support MPS during training and inference

ori-kron-wis:
add support for Mac GPU (m1,m2,m3) for scvi and revert it to be default option when running from Mac (if available)

Testing are done manually on my MAC (has m3) + verification that still works on CPU/CUDA here

References:
https://github.com/pytorch/pytorch/issues/132605
https://github.com/pytorch/pytorch/issues/77764
https://discourse.scverse.org/t/macbook-m1-m2-mps-acceleration-with-scvi/2075/7


canergen:
Please enable the Mac Runner for changes about MPS.

ori-kron-wis:
A comparison for training CPU, MPS M3 AND NVidia RTX 6000 Ada 48GB GDDR6
image

canergen:
The difference was much bigger in my hands when increasing batch size (larger batch size creates larger matrix multiplications where MPS is more efficient than CPU) and compilation. It was actually faster than an A100 on Google Colab then. We should optimize this a bit.

src/scvi/data/_preprocessing.py
```diff
@@ -123,12 +123,18 @@ def poisson_gene_selection(
...
         # Calculate empirical statistics.
         sum_0 = np.asarray(data.sum(0)).ravel()
-        scaled_means = torch.from_numpy(sum_0 / sum_0.sum()).to(device)
         total_counts = torch.from_numpy(np.asarray(data.sum(1)).ravel()).to(device)
-
-        observed_fraction_zeros = torch.from_numpy(
-            np.asarray(1.0 - (data > 0).sum(0) / data.shape[0]).ravel()
-        ).to(device)
+        # in MPS we need to first change to float 32, as the MPS framework doesn't support float64.
+        if device.type == "mps":
```
canergen, line 128: We could also do it for other devices. Float32 is sufficient for all computations.

tests/model/test_destvi.py
```diff
@@ -31,7 +31,8 @@ def test_destvi():
 
     # step 3 learn destVI with multiple amortization scheme
 
-    for amor_scheme in ["both", "none", "proportion", "latent"]:
+    for amor_scheme in ["both", "proportion", "latent", "none"]:
+        print("IIIII", amor_scheme)
```
canergen, line 35: Remove this.

canergen (approved):
See two minor comments. One tensor should be float32 independent of MPS second printing during test. Add warning that MPS can be enabled now. Afterwards merge.

PR #3110, opened 2024-12-23
bug: CUDA fix in decipher

ori-kron-wis:
@justjhong  I already fixed one of the tests (but you should review it)

close https://github.com/scverse/scvi-tools/issues/3109

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
         types: [yaml]
 
   - repo: https://github.com/executablebooks/mdformat
-    rev: 0.7.19
+    rev: 0.7.21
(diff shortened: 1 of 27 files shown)


justjhong (approved):
Sorry about this and thanks for squashing the bug! Is the other test fixed as well or should I take a stab at it?

ori-kron-wis:
@justjhong no the other test is failing, I wasnt sure what will be the way to go, please see if you can fix it.
yes it is basically on device handling (cpu and gpu calculation at the same time, but in the trajectory part, which is all cpu)

justjhong:
@ori-kron-wis @canergen remaining test should be fixed now

PR #3121, opened 2025-01-05
feat: add get_normalized_expression to more models

ori-kron-wis:
add this down stream analysis function to several models which don't have it yet


src/scvi/external/cellassign/_model.py
```diff
@@ -33,7 +33,7 @@
 B = 10
 
 
-class CellAssign(UnsupervisedTrainingMixin, BaseModelClass):
+class CellAssign(UnsupervisedTrainingMixin, BaseModelClass, RNASeqMixin, VAEMixin):
```
canergen, line 36: Is CellAssign intended for these functions? I would remove it here.

src/scvi/module/_mrdeconv.py
```diff
@@ -179,24 +187,47 @@ def __init__(
...
+        """High level inference method.
+
+        Runs the inference (encoder) model.
+        """
+        encoder_input = [x]
+
+        z = self.V_encoder(*encoder_input)
+        # z = self.gamma_encoder(*encoder_input)
+
+        return {
```
canergen, line 225: This doesn't really make sense. Please remove VAEmixin from DestVI

src/scvi/external/gimvi/_module.py
```diff
@@ -418,11 +435,15 @@ def generative(
         )
         px_rate = px_scale * torch.exp(library)
 
+        if transform_batch is not None:
+            batch_index = torch.ones_like(batch_index) * transform_batch
+
         return {
             "px_scale": px_scale,
-            "px_r": px_r,
+            "px": px_r,
```
canergen, line 443: this is wrong. px_r need to be the distribution like Poisson or NB.

src/scvi/model/_destvi.py
```diff
@@ -390,6 +397,7 @@ def setup_anndata(
         cls,
         adata: AnnData,
         layer: str | None = None,
+        batch_key: str | None = None,
```
canergen, line 400: this is not supported. only supported in destvi-v2.

src/scvi/module/_mrdeconv.py
```diff
@@ -206,6 +214,9 @@ def generative(self, x, ind_x):
         x_ = torch.log(1 + x)
         # subsample parameters
 
+        if transform_batch is not None:
```
canergen, line 217: it is actually fully ignored.

src/scvi/module/_peakvae.py
```diff
@@ -322,14 +326,14 @@ def generative(
...
-        return {"p": p}
+        return {"px": p}
 
     def loss(self, tensors, inference_outputs, generative_outputs, kl_weight: float = 1.0):
         """Compute the loss."""
         x = tensors[REGISTRY_KEYS.X_KEY]
         qz = inference_outputs["qz"]
         d = inference_outputs["d"]
-        p = generative_outputs["p"]
+        p = generative_outputs["px"]
```
canergen, line 336: I prefer it if it is the same. so change to px here.

src/scvi/module/_vaec.py
```diff
@@ -156,7 +156,7 @@ def inference(
         x_ = x
         library = x.sum(1).unsqueeze(1)
         if self.log_variational:
-            x_ = torch.log1p(x_)
+            x = torch.log1p(x_)
```
canergen, line 159: keep it to x_ to be consistent with other models.

canergen (approved):
see additional requests.

src/scvi/external/gimvi/_module.py
ori-kron-wis: the idea (with every model change in this branch) is to make each model already compatible to the get_norm function we have in rnaseqmixin, which sometimes means changing the generative function return keys (in this example from 'px_r' to 'px'). Otherwise, we will have to create a unique get_norm function for each model separately. 
There is no change in what is the 'px_r' of gimvi.

src/scvi/module/_mrdeconv.py
ori-kron-wis: yes I disable the use of it but Its still there for the same reason of supporting functions signatures.

src/scvi/model/_destvi.py
ori-kron-wis: its here only for it to work with rest of changes (esp mrdeconv), the get_norm is deprecated.
there is no intention to start the logic of batch_key here.

src/scvi/module/_peakvae.py
ori-kron-wis: it was originally 'p', changed now to 'px' - but this goes under changing the peakvi model signature

src/scvi/module/_vaec.py
ori-kron-wis: ok. this was actually a bug in the log transform

src/scvi/model/_destvi.py
canergen: You should not expose it as a function argument. It should be handled in mrdeconv. Same for other models. The user expects an effect if it is exposed.

src/scvi/external/gimvi/_module.py
canergen: I get it, but px is the generative distribution. You should define it here (negative binomial of scale and px_r. You should not just store something under this value.

src/scvi/model/_destvi.py
ori-kron-wis: ok, yes make sense

PR #3125, opened 2025-01-07
feat: Adding multiGPU option to SCANVI

ori-kron-wis:


ori-kron-wis:
Running multigpu scanvi on an interactive session is not simple, because we usually need to train scvi first and then init the scanvi from it.
Better to do it all in one script and not from jupyter, but if we have to do in jupyter or other interactive shells (which make sense in order to see figures), we can do it in 2 different sessions (while saving and loading the scvi model using cpu) or 1 session but then to train the scvi with cpu. but we cant have scvi and scanvi running on the same interactive session with multiGPU (this is a lightning issue)

ori-kron-wis:
another caveat is that it cant run with early stopping callbacks (and some models, like totalvi, run it by default), but it works for any model practically (anndata or mudata)

ori-kron-wis:
In order to run a multi-GPU training in an interactive session, like JupyetNB, the strategy should be:"ddp_notebook_find_unused_parameters_true", in such case, however, we cant see the model.history

canergen (approved):

canergen:
We should mention it in the tutorial where things are changed now that this uses multi-GPU. We should create a page in our docs discussing multi-GPU.

canergen (approved):

ori-kron-wis:
> Does scANVI now work with partially labeled data? Does the subsample callback work? I don't see any code changes for those so I assume yes?

Yes. as discussed, we are having trouble seeing multigpu ddp running on pytest on remote servers. It works by just running Python scripts end to end (I might return to dig into this in the future). So right now, all is manually checked, and yes, it can work with partially labeled data or not.

The resampling works. the change was not in its callback itself or its sub-functions but in the fact of how dataloaders were concatenated when using ddp, this is seen in the changes done for files semi and concat dataloaders in the case of DDP use.

PR #3146, opened 2025-01-14
vmap False by default

### LINKED ISSUE #3145: Adjust MrVI running hyperparameters for 1000s of samples
The default execution of MrVI core functions relies on vmap, which fastens execution and increases memory usage.
This memory cost is not sustainable in scenarios with 1000s of samples.
A first step in this direction would be to disable vmap by default. @justjhong @canergen what do you think?

PierreBoyeau:
Fixes #3145

### DIFF
--- a/src/scvi/external/mrvi/_model.py
+++ b/src/scvi/external/mrvi/_model.py
@@ -309,7 +309,7 @@ class MRVI(JaxTrainingMixin, BaseModelClass):
         adata: AnnData | None = None,
         indices: npt.ArrayLike | None = None,
         batch_size: int | None = None,
-        use_vmap: bool = True,
+        use_vmap: bool = False,
         norm: str = "l2",
         mc_samples: int = 10,
     ) -> xr.Dataset:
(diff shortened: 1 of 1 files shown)


canergen:
@PierreBoyeau please provide an update about the use case with 1k samples?

.gitmodules
```diff
@@ -1,4 +0,0 @@
-[submodule "docs/tutorials/notebooks"]
```
justjhong, line 1: why is this deleted?

src/scvi/external/mrvi/_model.py
```diff
@@ -1557,3 +1583,9 @@ def _construct_design_matrix(
         covariates_require_lfc = jnp.array(covariates_require_lfc)
 
         return Xmat, Xmat_names, covariates_require_lfc, offset_indices
+
+    def update_sample_info(self, adata):
```
justjhong, line 1587: great idea. can you expand on the docstring with the Adata parameter? since it relies on _scvi_sample, it requires it have setup anndata called on it. Maybe, we can just call validate_anndata in the first line of this function.

justjhong (commented):
Looks great, thanks for addressing the numpy issue as well. One important comment on why the notebook submodule is deleted, and one minor comment. let me know when i can look again for approval

PierreBoyeau:
> Looks great, thanks for addressing the numpy issue as well. One important comment on why the notebook submodule is deleted, and one minor comment. let me know when i can look again for approval

Hi Justin, I have no idea how I managed to delete `.gitmodules`, I reverted the change. I also improved the docstring, including adding a comment, to `update_sample_info`, thanks!

justjhong (approved):
LGTM

PR #3164, opened 2025-01-27
feat: SemiSupervised Training Mixin class

ori-kron-wis:
use what @ethanweinberger  already made as a basis to semisupervisedtrainingmixin
We would like to add more functionality to it (e.g: interoperability)


src/scvi/model/base/_training_mixin.py
```diff
@@ -143,3 +158,316 @@ def train(
...
+            Minibatch size for data loading into model. Defaults to `scvi.settings.batch_size`.
+        use_posterior_mean
+            If ``True``, uses the mean of the posterior distribution to predict celltype
+            labels. Otherwise, uses a sample from the posterior distribution - this
+            means that the predictions will be stochastic.
+        ig_interpretability
+            If True, run the integrated circuits interpretability per sample and returns a score
+            matrix, in which for each sample we score each gene for its contribution to the
+            sample prediction
+        shap_values
```
canergen, line 215: have you tried it? does it take forever?

canergen (approved):

src/scvi/model/base/_training_mixin.py
ori-kron-wis: I guess I need to also check for large dataset.
IG works fast , shap is dependent on the number of classes

canergen: Yes shap is known to be slow. I'm not sure we need it. IG is fine. Is this function argument exposed?

canergen: Can we link to https://www.nature.com/articles/s41592-024-02511-3 for why this is useful and validation of this approach?

ori-kron-wis: yes

PR #3177, opened 2025-02-10
first possibility of updated tutorials index page

lordy5:


ori-kron-wis:
very nice @lordy5 !
Im attaching an image here so we can have an idea how it turns to look like (you can do the same if there will be more PR's)
image

PR #3178, opened 2025-02-10
second possibility of updated tutorials index page

lordy5:


lordy5:
Screenshot 2025-02-10 at 9 39 53 AM

canergen:
Best one of the three style-wise. I don't think it will get much nicer. It will be essential to add describing text to the tutorials to make navigation easier. Maybe we disable the whole list on the left and also make it expandible subheaders.

lordy5:
I fixed the sidebar so it is hierarchical again, although the dropdowns look slightly different because of some changes I had to make. I will probably include the remaining changes on a new branch since things will be significantly different. Do we want to use this version while I'm working on the other features?

canergen:
Do you need to have the headings twice? Fine with me, just trying to understand whether it's some effort to have it flat within each section.

lordy5:
> Do you need to have the headings twice? Fine with me, just trying to understand whether it's some effort to have it flat within each section.

I was trying to fix this but couldn't find a simple way without manually linking each tutorial instead of using the table of contents directive. I could do this but not sure if it's worth it since things will be changed a lot soon.

PR #3187, opened 2025-02-12
add additional features to tutorial index page

lordy5:


lordy5:
@ori-kron-wis @canergen For the group and tutorial descriptions, some of them will likely have to be changed but I wanted to get something down for now. Let me know if there are any obvious changes I should make. I'm also still working on assigning all the tags and adding more tags.

PR #3193, opened 2025-02-18
Support of dask arrays in AnnTorchDataset.

LArnoldt:
Dask Array Support has been added to AnnData in the recent years (see e.g. https://anndata.readthedocs.io/en/latest/tutorials/notebooks/anndata_dask_array.html). The commit adds dask array support for the AnnTorchDataset.

### DIFF
--- a/src/scvi/data/_anntorchdataset.py
+++ b/src/scvi/data/_anntorchdataset.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import logging
 from typing import TYPE_CHECKING
 
-import dask.array as da
 import h5py
 import numpy as np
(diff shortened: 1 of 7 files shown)


src/scvi/data/_anntorchdataset.py
```diff
@@ -153,6 +154,8 @@ def __getitem__(
                 # used to record the type data minification
                 # TODO: Adata manager should have a list of which fields it will load
                 continue
+            elif issparse(data) or isinstance(data, da.Array):
```
canergen, line 157: Why do we compute for sparse data? Rest LGTM.

PR #3204, opened 2025-02-24
docs: SysVI docs

ori-kron-wis:
@Hrovatin feel free to add a model description to /Users/orikr/PycharmProjects/scvi-tools/docs/user_guide/models/sysvi.md, in this PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ to [Semantic Versioning]. Full commit history is available in the
     AutoZI, CellAssign and GimVI. {pr}`3121`.
 - Add {class}`scvi.external.RESOLVI` for bias correction in single-cell resolved spatial
     transcriptomics {pr}`3144`.
-- Add {class}`scvi.external.SysVI` SysVI for cycle consistency loss and VampPrior {pr}`3195`.
+- Add {class}`scvi.external.SysVI` for cycle consistency loss and VampPrior {pr}`3195`.
 
 #### Fixed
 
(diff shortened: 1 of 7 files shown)


Hrovatin:
@ori-kron-wis thank you very much.
I see many of the other models have small tutorials in the md file and a lot of additional theory behind it. What would be your general guide on how much info we should add here, especially since we have a detailed tutorial.

ori-kron-wis:
@Hrovatin , well, it's up to you. 
While the tutorial aim is to show an end-to-end typical use case of your model, the model description aim is to have a more theoretical background on it (see some other example, usually we expect to see advantages & limitations of model , the generative and inference processes scheme and tasks it can perform). 
I can see your tutorial has some theoretical parts, it's fine, they can have some overlap between them.

As I said, it's up to you, we do have other external models with no theoretical background in RTD, which is a bit pitty, as it reduces the chances of users to really adopting the model.

Hrovatin:
@ori-kron-wis I added the model description here https://github.com/Hrovatin/scvi-tools/blob/Ori-sysvi-docs-ready/docs/user_guide/models/sysvi.md

Should I open a new PR for it?

The description I decide on is more general (without strict mathematical formulation) to appeal to users without CS background.

Also, I was not able to build the docs due to error with:

```
Extension error (sphinx.ext.autosummary):
Handler <function process_generate_options at 0x105df6d40> for event 'builder-inited' threw an exception (exception: no module named scvi.autotune)
```
So please have a look that the two added images display properly.

ori-kron-wis:
@Hrovatin thanks! looks great. I will take care on putting this in our RTD.

PR #3225, opened 2025-03-02
docs: update sysvi docs and images

ori-kron-wis:
@Hrovatin


Hrovatin:
@ori-kron-wis thank you for taking care of it.

PR #3243, opened 2025-03-10
fix: SHAP expose params to user

ori-kron-wis:
updates in shap


ori-kron-wis:
putting this on hold for now.

PR #3246, opened 2025-03-11
feat: TQDM to transform_batch for loop & Update to WIS AWS S3

ori-kron-wis:


src/scvi/external/contrastivevi/_model.py
```diff
@@ -419,7 +420,7 @@ def get_normalized_expression(
             batch_index = tensors[REGISTRY_KEYS.BATCH_KEY]
             background_per_batch_exprs = []
             salient_per_batch_exprs = []
-            for batch in transform_batch:
+            for batch in tqdm(transform_batch):
```
canergen, line 423: @ori-kron-wis: Michal was complaining about this, as it prints quite a lot. Revert?

ori-kron-wis: It was here at first place cause she didn't know how long it will take, however since it prints a lot, I added a follow up PR that has a silence parameter which is True by default.

ori-kron-wis: https://github.com/scverse/scvi-tools/pull/3280

PR #3247, opened 2025-03-12
feat: added python 3.13 support

ori-kron-wis:


ori-kron-wis:
waiting for ray autotune to support py3.13 (this is also why RTD is failing)

Zethson:
https://github.com/ray-project/ray/blob/master/pyproject.toml#L2

are they not?

ori-kron-wis:
> https://github.com/ray-project/ray/blob/master/pyproject.toml#L2
> 
> are they not?

Yep. Thanks for noticing!

PR #3255, opened 2025-03-17
fix: sysVI get normalzied expression fix

ori-kron-wis:


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3251

canergen (approved):

PR #3257, opened 2025-03-18
docs: general fixes to docs + expose SHAP params

ori-kron-wis:


ori-kron-wis:
added here also the exposing of SHAP params and putting tune tests under label

PR #3259, opened 2025-03-18
feat: TotalANVI model updated

ori-kron-wis:


ori-kron-wis:
Also added the scib-metrics optimization support for this model in this branch

PR #3324, opened 2025-05-14
fix: mrvi - change self.adata to just adata inside the function compute local statistics

ori-kron-wis:


justjhong (approved):
Looks good thanks ori! Thanks for catching the other instances of the bug as well

PR #3326, opened 2025-05-14
Remove collection.load() from MappedCollectionDataModule

Koncopd:
`collection.load()` loads all `AnnData` objects into memory, this is very problematic for large datasets.


ori-kron-wis:
Thanks @Koncopd for noticing it!

PR #3328, opened 2025-05-14
Ensure that compatible bionty version is installed

Zethson:
Might not make a difference in most or even all cases but this is how you're supposed to install lamindb and the correct bionty version.


ori-kron-wis:
@Zethson Thanks. Added

PR #3362, opened 2025-05-29
feat: adding annCollection dataloader

ori-kron-wis:
based on https://github.com/scverse/scvi-tools/issues/3018


docs/user_guide/use_case/custom_dataloaders.md
```diff
@@ -130,7 +132,14 @@ model.train(
 )
 ...
 ```
-Key Differences between them in terms of Custom Dataloaders:
+
+3. In addition to those we also added a {class}`scvi.dataloaders.CollectionAdapter` dataloader which is a scvi-tools wrapper for Anndata's AnnCollection.
```
canergen, line 136: In addition to those we also added a {class}`scvi.dataloaders.CollectionAdapter` dataloader which is an scvi-tools wrapper for AnnData's AnnCollection.
This API is intended for training models on a set of AnnData files. AnnCollection creates a layy concatenation of disk-backed files. The dataloader jointly subsets these datasets along the observations axis.
This wrapper mimics the standard API, soo in practice, users wrap their collection objects then proceed with the scvi-tools workflow as normal and not using the custom dataloader workflow.
Note that in order to use it count data should be in sparse form (sparse.csr_matrix).

src/scvi/dataloaders/_anncollection.py
```diff
@@ -0,0 +1,174 @@
...
+            batch = self.collection[start:end].layers[self.key]
+            nnz += batch.getnnz()
+            start = end
+            end = min(end + self.batch_size, self.collection.shape[0])
+
+        return nnz
+
+    def _mul_vector(self, other):
+        """Included to try and trick issparse"""
+        raise NotImplementedError("This should basically never be called.")
```
canergen, line 46: Do we need those functions. In case, it might get called we should improve the error message but likely we can just remove it.

src/scvi/dataloaders/_anncollection.py
```diff
@@ -0,0 +1,174 @@
...
+        self,
+    ):
+        """Convert to a dense array.
+
+        Notes
+        -----
+        Ths would require full dense representation in memory. It shouldn't be
+        used.
+        """
+        raise NotImplementedError("This should basically never be called.")
```
canergen, line 64: Improve error message. Would this fail? I could see calling it on a subset of the data and it shouldn't fail then?

src/scvi/dataloaders/_anncollection.py
ori-kron-wis: yes, just a legacy from the implementation that we took it from.

canergen:
There seems to be more legacy. Can your remove all and tag me? I can have another look then.

PR #3369, opened 2025-06-04
feat: added posterior predictive samples batch projection

ori-kron-wis:
fix in LDA perplexity calc also


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3352

PR #3378, opened 2025-06-08
Ori destvi v2

ori-kron-wis:


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/1826

PR #3430, opened 2025-07-22
fix: add SupervisedModuleClass to the classifier 

ori-kron-wis:
also add test for save and load SOLO model, and updated the docs


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3429

PR #3443, opened 2025-07-27
feat: Downstream Analysis functions multi GPU support

ori-kron-wis:
Added general data loader kwargs to down stream analysis functions to be able to run a distributed sampler with them. Supported in scvi main models currently.


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3442

ori-kron-wis:
Should we also add to other models than the defaults?

PR #3452, opened 2025-08-03
fix: library_size in totalvi, summary_stats duplicate fields and checkpointing with autotune

ori-kron-wis:
fix in library size in totalvi, added a check for existing fields in summary_stats and added checkpointing for autotune
close https://github.com/scverse/scvi-tools/issues/1906
close https://github.com/scverse/scvi-tools/issues/1744
close https://github.com/scverse/scvi-tools/issues/2580


ori-kron-wis:
also updated the installation page and the cuda workstations yaml files (to support jax changes)

PR #3466, opened 2025-08-15
Fix scarches surgery in SysVI

Hrovatin:
Changes to scarches implementation from @ori-kron-wis in https://github.com/scverse/scvi-tools/pull/3448
Not yet tested on real data.

### DIFF
--- a/src/scvi/external/sysvi/_model.py
+++ b/src/scvi/external/sysvi/_model.py
@@ -168,9 +168,9 @@ class SysVI(UnsupervisedTrainingMixin, RNASeqMixin, VAEMixin, ArchesMixin, BaseM
         if "transfer_batch" in kwargs:
             _ = kwargs.pop("transfer_batch")
             warnings.warn(
-                "The setting of transfer_batch is disabled in SysVI "+
-                "and is automatically set to False.",
-                stacklevel=settings.warnings_stacklevel
+                "The setting of transfer_batch is disabled in SysVI "
+                + "and is automatically set to False.",
+                stacklevel=settings.warnings_stacklevel,
             )
         return super().load_query_data(*args, transfer_batch=False, **kwargs)
 


ori-kron-wis:
@Hrovatin made small fixes (pre-commit issue) + I had the transfrom_batch as a parameter in archesmixin, as we generally don't use globals in a model (except for classes names) and more models might benefit from this perhaps.
But perhaps youll want to emphasis this in the tutorial now.

beside that looks good. also tutorial. is there anything else missing? or we good to merge?

Hrovatin:
@ori-kron-wis  The purpose of the class attribute for preventing batch surgery in SysVI is that we want to disable surgery on batch whenever SysVI is used, as discussed [here](https://github.com/scverse/scvi-tools/pull/3448#discussion_r2264607523) with @canergen. 
Changing it to a method parameter defeats its purpose. I would strongly push for having it as a class param as this is the only way I could find to message the user that model can not be used with batch surgery (raise error automatically). But if you have another suggestion on how to achieve this feel free to adjust. 

EDIT: I guess another option would be to add a validator hook to that scarches function that I can override in SysVI if that fits better

ori-kron-wis:
> @ori-kron-wis The purpose of the class attribute for preventing batch surgery in SysVI is that we want to disable surgery on batch whenever SysVI is used, as discussed [here](https://github.com/scverse/scvi-tools/pull/3448#discussion_r2264607523) with @canergen. Changing it to a method parameter defeats its purpose. I would strongly push for having it as a class param as this is the only way I could find to message the user that model can not be used with batch surgery (raise error automatically). But if you have another suggestion on how to achieve this feel free to adjust.
> 
> EDIT: I guess another option would be to add a validator hook to that scarches function that I can override in SysVI if that fits better

The reason is clear to me. Lets think of a solution that involves the change in the sysvi model class and not in the scarches mixin class.

Hrovatin:
@ori-kron-wis, do you have a suggestion on how to achieve it? All my ideas (above) require modification of scarches mixin

ori-kron-wis:
yes I just added load_query_data wrapper to sysvi with forcing transfer_batch=False. thats it.

Hrovatin:
@ori-kron-wis I fixed it a bit to ensure it is used properly and extended a test to test for it.

ori-kron-wis:
please fix the pre-commit errors so we can merge this

Hrovatin:
@ori-kron-wis done

PR #3482, opened 2025-08-22
chore(deps): add constraint to jax and jaxlib

TTTPOB:
jax has announced deprecation of `from jaxlib.xla_extention import Device`. I've tested it still works in 0.6.0 but not in 0.6.2. `conda-forge` did not have a 0.6.1 package so I didn't test that.

Related url: https://github.com/jax-ml/jax/releases/tag/jax-v0.6.0

It said removal will happen in 0.7.0, but in my test, that was not the case.
image

It the future, maybe we can consider move to public stable api.


ori-kron-wis:
Closing as we put JAX optional dep now

PR #3501, opened 2025-09-07
feat: add a checkpoint to trainer fit

ori-kron-wis:


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3500

alehuerta1 (approved):

ori-kron-wis:
@alehuerta1 , can I ask you to kindly add a unit test to support your case, as currently, I only added the general model save/load checkpoint using this addition. I understood you would like to dig in into the optimizer state.

marianogabitto:
Ori,
   I believe  there is no additional unit test for it, to be sincere, we just passed a saved checkpoint. We are good as it is. What were you thinking on us adding?

Best,
M

ori-kron-wis:
We already have a SaveCheckpoint class in scvi-tools for those cases (you can see it [here](https://github.com/scverse/scvi-tools/pull/3501/files#diff-36666f7fb1d6479c683ac9a7d5025a42b6cb821f7959c5c3727baefcf66b78ea)), which inherits ModelCheckpoint and adds more capabilites.
What is the benefit for this addition we are doing here? as currently it is almost the same like what we already have.
Show in a unit test that its adds the capabilities you need, and more than that, adding the unit test will make sure your new case won't break with future developments.
Otherwise, if theres no difference, why not using SaveCheckpoint we have now?

alehuerta1:
ok I will add a unit test to check for optimizer and scheduler states

PR #3506, opened 2025-09-09
tests: change the tests to be on py3.13 by default

ori-kron-wis:
to see all new installations are working


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3453

PR #3516, opened 2025-09-10
Fix: Added missing model history in case of MultiGPU and able to log on step

ori-kron-wis:
Adding model history for global rank 0 only, in case of DDP, by using a temporary history file on master disk.
close https://github.com/scverse/scvi-tools/issues/3483


ori-kron-wis:
Also added the fix to be able to log within step.
close https://github.com/scverse/scvi-tools/issues/3020

PR #3538, opened 2025-09-21
fix: UnboundLocalError for sliced_data in AnnTorchDataset.__getitem__

ori-kron-wis:


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3535

PR #3561, opened 2025-10-19
Fix: Update totalvi legacy load function with new state_dict parameters

ori-kron-wis:
In order to deal with log_per_batch_efficiency


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3557

PR #3570, opened 2025-10-21
Model: DiagVI

ori-kron-wis:


ori-kron-wis:
closing here, a duplicate of https://github.com/scverse/scvi-tools/pull/3575

PR #3572, opened 2025-10-22
Fix installation command quotes in installation.md

alexandre-jf-fernandes:
Updated installation commands to use single quotes for package names.


ori-kron-wis:
Hey @alexandre-jf-fernandes ,
While this is true for mac OS, in Linux (the most popular platform to run scvi-tools), you can run those commands without the quotations. In what OS did you try to run?

In any case, I agree we can add a comment for other platforms (but we don't want to give the whole pip install constraints - its not related to scvi-tools)

ori-kron-wis:
will be part of https://github.com/scverse/scvi-tools/pull/3579

PR #3579, opened 2025-10-29
Feat: Adding get_normalized_expression to mrvi in torch

ori-kron-wis:


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3546

ori-kron-wis:
The only changes in get_normalized_expression for mrvi in torch are:
- library_size is set to "latent" by default, return_mean is set to False. 
- inference_kwargs gets "use_mean" and "mc_samples" as keys (instead of just "n_samples")
- in case of n_samples == 1 (mc_samples in practice), need to deal with extra dim in mrvi's inference_outputs & generative_outputs

PR #3603, opened 2025-11-18
Update contributing guide link in README.md

agdenadel:


ori-kron-wis:
@agdenadel Thank you for this!

PR #3607, opened 2025-11-18
refactor: use the scverse cookiecutter template

### LINKED ISSUE #3600: Use scverse template
Hey,

from what I can see, you're currently not (fully) using the scverse template. Would it be possible to do so, please? I'd be happy of we could converge on a more consistent developer experience across all of our repositories.

Then we can also enable the sync.

ori-kron-wis:
Aligns the repository with the [scverse cookiecutter template](https://github.com/scverse/cookiecutter-scverse) v0.8.0 and tracks it in `.cruft.json`, so future template changes arrive as automated pull requests.

Closes #3600.

## Packaging

- `dev`, `test` and `doc` dependencies move from `[project.optional-dependencies]` to PEP 735 `[dependency-groups]`. The `tests`, `test`, `editing`, `dev`, `docs` and `docsbuild` extras are gone; the feature extras (`autotune`, `cuda`, `hub`, `rapids`, …) stay.
- Every test suite is now a hatch environment in `pyproject.toml`: the `hatch-test` matrix for the default suite, plus `test-cuda`, `test-metal`, `test-dataloaders` and `test-resolution` for the opt-in ones. `pyproject.toml` is the single point of truth — CI runs exactly what `hatch test` runs locally.
- `nvidia-nccl` moved from ad-hoc CI steps into the `cuda`/`cuda13` extras, and `session-info2` was added as a dependency because the new issue form asks reporters for its output.

## CI

- `build.yml` → `build.yaml`, `test_linux.yml` → `test.yaml`, `release.yml` → `release.yaml`, with the template's contents: hatch-driven, `permissions:` blocks, actions pinned by digest, and an `alls-green` check so branch protection needs a single required job.
- The ten specialized workflows were renamed off the `test_linux_*` prefix and rewritten to run through the hatch environments above. Their triggers and label gating are unchanged.
- The release workflow creates the tag with `gh release create` instead of a checkout that persists push credentials, and the Docker matrix uses extras that still exist (`""`, `optional`, `tutorials`).
- Added `.github/dependabot.yml`.

## Tooling

- `.pre-commit-config.yaml` matches the template: biome, `pyproject-fmt` and zizmor replace prettier, mdformat, markdownlint and blacken-docs. `.markdownlint.yaml` is removed; the local "don't commit .rej files" hook stays.
- Added `biome.jsonc` and `.vscode/{settings,launch,extensions}.json`; aligned `.editorconfig`, `.gitignore`, `.readthedocs.yaml` and the issue templates (now GitHub forms).

## Documentation

The docs are now built with `-W`, as the template does. Getting there meant fixing roughly 400 pre-existing warnings rather than muting them:

- Corrected cross-references that pointed at private module paths, wrong modules (`resolvi.train.PyroTrainingPlan`, `pytorch_lightning.trainer.Trainer`) or the wrong role, and fixed the pyro/torch targets to the names those projects actually publish.
- Fixed malformed numpydoc parameter blocks and type names (`np.array` used as a type, `torch.tensor` in annotations, `list[np.array, np.array, np.array]`).
- Documented public classes that were referenced but missing from the API reference, and gave undocumented overrides real docstrings instead of inheriting lightning's.
- Removed dead cross-references from historical `CHANGELOG.md` entries.
- Optional dependencies are now mocked based on whether they actually import, rather than on `READTHEDOCS`, so local, CI and Read the Docs builds behave the same.
- `nitpick_ignore` shrank from 15 entries to 8, all of them third-party or CPython inventory mismatches we cannot address here.

`.readthedocs.yaml` now builds through `uvx hatch run docs:build`. It stays on Python 3.12 rather than the template's 3.14 because the docs pull in the full optional dependency set.

## Notes for reviewers

- Workflow renames change the check names, so branch protection needs updating. `Tests pass in all hatch environments` is the new single required check for the default suite.
- Coverage thresholds and the per-suite flags in `.codecov.yaml` are unchanged; the template's looser defaults were not adopted.
- The tutorials submodule now points at `scverse/scvi-tutorials` rather than the old `YosefLab` URL.


Zethson (changes requested):
I think that's a start but there's more. I still think that the better way would be to recreate the project from scratch but if this is easier for you, we should ensure that we do it right.

1. What about the issue templates?
2. What about the hatch test set up?
3. What about the whole CI set up? Yours differs greatly from the template and I'm a bit surprised at the number of workflows that you have.
4. The pre-commit setup needs tons of updates, please.
5. The whole configuration of the linters as well -> please also update the pyproject.toml
6. The RTD setup needs updates
7. There's probably more that I missed...

Zethson:
@ori-kron-wis where are we with this PR? Should we have an agent finish it? I'd love to see a full version merged.

ori-kron-wis:
> @ori-kron-wis, where are we with this PR? Should we have an agent finish it? I'd love to see a full version merged.

Do we have an agent for this specific task? If so, sure, let's run it.

Zethson:
Not yet. I can take care of this as it pains me that this isn't done yet. I'll ping you when I'm done.

Zethson:
@ori-kron-wis there might still be minor docs issues but I'll fix these. Generally, you can already have a look now, please. I know it's a lot of changes but I'm confident that this will help contributions and generally the software quality of scvi-tools (not saying it bad was but the new automations are nice)

ori-kron-wis:
@Zethson, note I updated the tutorials sub-module on main (after recent mergers on scvi-tutorials). Please update it here (currently in conflict and we need to see the updated docs renders)

Zethson:
@ori-kron-wis updated the submodule. Please have a look. Thank you!

ori-kron-wis:
Because of so many changes, well have to test everything

docs/conf.py
```diff
@@ -1,15 +1,18 @@
...
 import subprocess
 import sys
-import types
+import warnings
 from pathlib import Path
 from importlib.metadata import metadata
 from datetime import datetime
 from typing import TYPE_CHECKING
 
+from sphinx.deprecation import RemovedInSphinx90Warning
```
ori-kron-wis: 1. (High — latent build-breaker) docs/conf.py imports sphinx.deprecation.RemovedInSphinx90Warning unguarded to suppress a hoverxref warning. That symbol only exists pre-Sphinx-9; once the       
  sphinx>=8.1,<9 pin is ever bumped (which this same PR's cruft-tracking sets up to happen automatically), conf.py fails to import and the whole doc build hard-crashes. No try/except, unlike the  
  new _importable() helper a few lines below for other optional deps.

ori-kron-wis:
close https://github.com/scverse/scvi-tools/pull/3937

Zethson:
Looks like everything is green 🚀

PR #3616, opened 2025-11-24
feat: differential abundance for scvi-tools models

lordy5:
still in progress

I still need to update comments, function headers, figure out return type for DA (dataframe didn't work when using sparse matrix for log probabilities), add testing

### DIFF
--- a/src/scvi/model/base/_da_pytorch.py
+++ b/src/scvi/model/base/_da_pytorch.py
@@ -1,14 +1,12 @@
 from collections.abc import Sequence
 
 import numpy as np
-
 import pandas as pd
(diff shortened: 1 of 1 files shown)


ori-kron-wis:
@lordy5 Please add only relevant files to this PR (right now all other code base files are deleted, I dont think you meant that) + tests + changlog + precommit fixes

lordy5:
> @lordy5 Please add only relevant files to this PR (right now all other code base files are deleted, I dont think you meant that) + tests + changlog + precommit fixes

Yes I didn't mean to do that I will fix it.

PR #3620, opened 2025-11-25
feat: Adding support for Annbatch

ori-kron-wis:


src/scvi/model/base/_base_model.py
```diff
@@ -1063,6 +1063,200 @@ def setup_anndata(
...
+                shuffle=shuffle,
+                dataset_size=dataset_size,
+                var_subset=var_subset,
+                load_adata=_load_adata_from_path,
+            )
+            return coll
+
+        def _open_collection(store_path: str) -> DatasetCollection:
+            store = zarr.open(store_path, mode="r")
+            return DatasetCollection(store)
```
ilan-gold: You can also do further validation on whether or not the collection contains data: https://annbatch.readthedocs.io/en/stable/generated/annbatch.DatasetCollection.html#annbatch.DatasetCollection.is_empty because people could provide random zarr paths here that are not actually collections

src/scvi/model/base/_base_model.py
```diff
@@ -1063,6 +1063,200 @@ def setup_anndata(
         on a model-specific instance of :class:`~scvi.data.AnnDataManager`.
         """
 
+    @classmethod
+    @dependencies("annbatch", "zarr")
+    def setup_annbatch(
+        cls,
+        paths: list[str],
```
ilan-gold: Make this optional? This way people could provide just a `zarr_path` (which maybe should be named `dataset_collection_path` or something) without having to provide input h5ads. The `obs` keys could then work on the existing store.

src/scvi/model/base/_base_model.py
```diff
@@ -1063,6 +1063,200 @@ def setup_anndata(
...
+            batch_size=batch_size,
+            chunk_size=chunk_size,
+            preload_nchunks=preload_nchunks,
+            preload_to_gpu=preload_to_gpu,
+            to_torch=True,
+        )
+        ds.use_collection(collection, load_adata=_load_adata_from_zarr)
+
+        ds_val = None
+        if paths_val is not None:
```
ilan-gold: Not sure here, maybe Felix has an opinion.  Your categories for hold-out validation could be spread out across files to begin with - T cells across a bunch of h5ads or something.  In this sense, it might make sense to expose `_load_adata_from_path` with an extra `train` / `val` arg of some sort to the user so they can choose which datasets/categories get loaded.

src/scvi/model/base/_base_model.py
```diff
@@ -1063,6 +1063,200 @@ def setup_anndata(
...
+            ds_val = Loader(
+                batch_size=batch_size,
+                chunk_size=chunk_size,
+                preload_nchunks=preload_nchunks,
+                preload_to_gpu=preload_to_gpu,
+                to_torch=True,
+            )
+            ds_val.use_collection(collection_val, load_adata=_load_adata_from_zarr)
+
+        return ZarrSparseDataModule(
```
ilan-gold: New name?

src/scvi/model/base/_base_model.py
```diff
@@ -1063,6 +1063,200 @@ def setup_anndata(
...
+        def _get_var_names(path: str) -> list[str]:
+            adata_ = ad.experimental.read_lazy(path)
+            if layer is None and getattr(adata_, "raw", None) is not None:
+                var = adata_.raw.var
+            else:
+                var = adata_.var
+            var_df = var.to_memory() if hasattr(var, "to_memory") else var
+            return list(var_df.index)
+
+        def _load_adata_from_path(path: str) -> ad.AnnData:
```
ilan-gold: Is there any case where people would want more than one `X` to have for future use - PCA isn't a big deal for scvi, I guess, but maybe a different data representation in `layers`? xxxVI is more opinionated in this respect, so just thinking out load.

src/scvi/dataloaders/_custom_dataloaders.py
```diff
@@ -975,3 +980,426 @@ def __iter__(self):
...
+        :class:`~scvi.dataloaders.MappedCollectionDataModule`.
+        """
+        X_np = batch["X"]
+        obs = batch.get("obs")
+
+        # Convert sparse batches to dense, then to float32 torch tensor.
+        if isinstance(X_np, torch.Tensor) and X_np.is_sparse_csr:
+            x = X_np.to_dense().float()
+        elif issparse(X_np):
+            x = torch.from_numpy(X_np.toarray()).float()
```
ilan-gold: Why not 
1. to torch csr data structure
2. to gpu
3. to dense
4. then to float
?

ilan-gold (commented):
High level stuff but overall very good and cool! Happy to help with docs as well!

ilan-gold:
@felix0097 please review!!

src/scvi/model/base/_base_model.py
```diff
@@ -1063,6 +1063,200 @@ def setup_anndata(
         on a model-specific instance of :class:`~scvi.data.AnnDataManager`.
         """
 
+    @classmethod
+    @dependencies("annbatch", "zarr")
+    def setup_annbatch(
```
ilan-gold, line 1180: Similarly, would it make sense to allow users to present their own `Loader` objects? I think things like `continuous_covariate_keys` and `categorical_covariate_keys` could be inferred directly from what is present in `obs`.  You could enforce that the `batch` key has a certain name (like `batch`) or that `batch_key` is an argument that can be applied.

src/scvi/model/base/_base_model.py
ori-kron-wis: no more than one X. we do have a different layers option. added a test for this.

src/scvi/model/base/_base_model.py
ori-kron-wis: We need to leave those open for users; covariates can be anything, really, as well as batch_key. we cant enforce their data

src/scvi/model/base/_base_model.py
ilan-gold: Ah, so this I would say is hardcoded in `annbatch`.  We only look in `X` (and this will remain the default although we'll probably allow new arguments when https://anndata.readthedocs.io/en/latest/accessors.html is released). So what you have now makes sense - just put in `X`.

src/scvi/model/base/_base_model.py
ilan-gold: My thinking goes that if a user presents a `Loader` with preloaded `obs`:

1. User specificies `batch_key` and `labels_key`, `unlabeled_category` to `scvi` (optionally)
2. Everything that is categorical (checkable by `dtype`) in `obs` outside of the above special keys is a  `categorical_covariate_keys`
3. The numeric keys are `continuous_covariate_keys`

ori-kron-wis: there can be more metadata which is categorical or cont but users will not want to set it up in their data object.

src/scvi/model/base/_base_model.py
ori-kron-wis: that ok - we follow anndata principles

src/scvi/model/base/_base_model.py
ilan-gold: Totally agreed, but I think then they should simply not put that metadata in the `Loader` when they load their anndata objects/`DatasetCollection` into the `Loader`. But if you want to make this explicit, also good.

ori-kron-wis:
Resolvi annabtch implementation still loads data into memory and can't be based on annbatch, although what was committed here. There is no real gain from it right now; need to think about graph dataloader for it as @canergen suggested.
Can it then be optimized with the annbatch backend? This is what we would expect to be in scviva-tools as a spatial dataloader, @ilan-gold (until we think about a spatialdata loader support)

I will remove those commits from here and put them on a side branch for later use: Ori-annbatch-resolvi

src/scvi/dataloaders/_custom_dataloaders.py
```diff
@@ -975,3 +980,443 @@ def __iter__(self):
...
+        if self._validset is not None:
+            self.n_val = self._validset.n_obs
+        else:
+            self.n_val = 0
+        self.n_train = dataset.n_obs
+
+    # The annbatch Loader is already an iterable that yields batches.
+    # We wrap it to hide __len__ (which returns n_obs, not n_batches)
+    # because Lightning uses len(dataloader) to compute val_check_batch;
+    # a misleading length prevents validation from ever triggering.
```
ilan-gold: I believe this [has been resolved](https://github.com/scverse/annbatch/pull/144): https://github.com/scverse/annbatch/blob/main/src/annbatch/loader.py#L222-L223

PR #3628, opened 2025-11-30
Fix: loading a model with adata=None and adata not exists

ori-kron-wis:
Have support to still load model with adata=None where adata not exists (like with adata False) and later use it in a new model


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3627

ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3623

PR #3629, opened 2025-11-30
Fix: batch size setting problem in ContrastiveVI

### LINKED ISSUE #3626: Batch_size setting problem
[Batch_size]

When calling the ContrastiveDataLoader to split the data, the batch_size is passed via self.data_loader_kwargs. However, when the parent class ConcatDataLoader(DataLoader) is invoked, batch_size is defined again with a fixed value. This causes a problem: the batch size configured for the model cannot be changed by the external parameter.

ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3626


ori-kron-wis:
Also added  `or settings.batch_size` in case batch_size is not set inside data loaders, for all models.

Consider to set the same batch_size for all models by default (right now, it varies between models, from 64 to 512)

PR #3638, opened 2025-12-08
refactor: pandas3 support

ori-kron-wis:
need to wait for anndata, mudata, scanpy updates


ori-kron-wis:
also mlflow is not yet supported with pandas3

ori-kron-wis:
We will have to remove all `pip install git+https://github.com/scverse/anndata appears` in the yaml of test once anndata is released as well as the changes in pyproject (will not need to pin anything)

PR #3679, opened 2026-01-19
fix: Adding a flag of whether to run Importance Sampling in RESOLVI Differential Experssion

ori-kron-wis:


canergen:
There is already a flag: weights. It's just the value is ignored. Otherwise, thanks for proposing the fix :).

ori-kron-wis:
> There is already a flag: weights. It's just that the value is ignored. Otherwise, thanks for proposing the fix :).

No No. weights, is still used and is doing a weighted choice of indexes at the very end.

It says it's supposed to not run IS if selected as "uniform", but IS will be run regardless. This is what you meant.

Up to you if to approve and merge this PR, or suggest a better fix, as I think the original intention was different from what is implemented in get_normalized_expression_importance right now. 

In any case the flag of IS + num of samples / Monte Carlo steps should be open for user, as it can really take long time for them.

PR #3699, opened 2026-02-04
Add DRVI

moinfar:
Dear Maintainers,

Here, I have prepared [DRVI](https://github.com/theislab/drvi) for contribution as an external model, matching the latest version of scvi-tools.
In short, it contains:

- DRVI models with all the required PyTorch modules I wrote. Specifically, 
  - FCLayers are modified to decode multiple parallel splits of the latent space at once. For this, we have a factory that generates hidden layers based on user setup.
  - The scArches Q2R mapping is implemented differently but shares some functionality.
  - The gene likelihood somehow registers parameters it needs to the model. So the wrappers are implemented separately.
- There are some additional functions to iterate over data and provide model outputs for interpretability, reconstruction, etc. Some of them seem redundant, but they have some details currently not available in RNASeqMixins.
- Interpretability functions to interpret additive decoders and plot latent space details and relevant genes.
- All the functions have tests to my knowledge.

I removed some extra stuff. The model may still contain much extra functionality. We can prune some after a discussion.
Please let me know if you need any further explanation of the code's functionality.

Thanks for your time.

### DIFF
--- a/docs/user_guide/use_case/llm_assisted_analysis.md
+++ b/docs/user_guide/use_case/llm_assisted_analysis.md
@@ -6,7 +6,7 @@ Large language models (LLMs) can significantly lower the barrier to using scvi-t
 
 ## scvi-tools MCP Server
 
-The **scvi-tools MCP server** ([`scvi-tools-mcp`](https://github.com/ori-kron-wis/scvi-tools-mcp)) is a [Model Context Protocol](https://modelcontextprotocol.io/) server that gives any MCP-compatible LLM structured access to scvi-tools knowledge: model documentation, tutorials, API reference, workflow templates, and a curated FAQ—with no runtime model execution.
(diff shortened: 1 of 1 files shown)


src/scvi/data/fields/_arraylike_field.py
```diff
@@ -418,7 +418,7 @@ def _make_array_categorical(
         df = getattr(adata, self.attr_name)[self.attr_key]
         for key in self.attr_keys:
             categorical_dtype = (
-                CategoricalDtype(categories=category_dict[key])
+                CategoricalDtype(categories=category_dict[key], ordered=True)
```
moinfar: This is required so that we can have query to reference mapping beyond just batch_key (also for cols in cat_covs_key)

ori-kron-wis:
Hey @moinfar , can you update on the status of this?

moinfar:
Hi @ori-kron-wis 

I don't know why this was automatically closed after I hard reset my main branch in the forked repo.
I made a new pull request #3866 with much more reuse of scvi code and removed as much extra functionality as possible (2k compared to 8k lines here). I added more info there.

PR #3702, opened 2026-02-08
Feat: Asymmetric n_layers for  encoder/decoder

### LINKED ISSUE #3698: [Bug/Feature] Asymmetric `n_layers` for encoder and decoder.
I am attempted to experiment with a weakened decoder as a means of reducing latent variable collapse in my scVI model. The idea being to have the **encoder use 4 hidden layers** and the **decoder use 1 hidden layers**. I have been combing the documentation for a means of doing this and intuitive it appeared to be possible via: [VAE.extra_encoder_kwargs and VAE. extra_decoder_kwargs](https://docs.scvi-tools.org/en/stable/api/reference/scvi.module.VAE.html#scvi.module.VAE). However upon implementing this I was met with the error shown below. I went digging into the source code and found that the `n_layers` argument is explicitly passed from scVI to the VAE and again explicitly passed to the Encoder/SCVIDecoder even when not specified globally.

Whilst asymmetric VAE is not by any means a default users should start with, the flexibility to make one should arguably be possible. The framework to allow it exists and is specified in the API as a potential possibility as such a refactor for the implicit communication of `n_layers` would be great. Alternatively `kwargs` could be given priority when passed to encoder/decoder.

If you have any suggestions for any simple workarounds that I could otherwise try in the short term, I would love to hear them.

```python
...

model = scvi.model.SCVI(
      adata, 
      ...
      extra_encoder_kwargs={'n_layers': 4},
      extra_decoder_kwargs={'n_layers': 1}
)

...
```

```pytb
TypeError: scvi.nn._base_components.Encoder() got multiple values for keyword argument 'n_layers'
```

#### Versions:

> VERSION: 1.4.1

ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3698


ori-kron-wis:
Only VAE has this change for now

PR #3749, opened 2026-03-16
fix: PyTorch MRVI to match JAX implementation and work on GPU

justjhong:
Fixes the PyTorch MRVI implementation to be numerically equivalent to the JAX version and work correctly on GPU. With identical weights, all forward pass outputs and gradients now match between backends to within float32 precision (~1e-6).

Architecture fixes
- AttentionBlock: Rewrote to match Flax MultiHeadDotProductAttention — old version had extra projection layers and wrong output dimension (outerprod_dim * n_channels * n_heads vs correct
   outerprod_dim * n_channels)
- GELU: Changed to approximate='tanh' to match JAX's nn.gelu(approximate=True)
- LayerNorm epsilon: Set to 1e-6 to match Flax default (PyTorch defaults to 1e-5)
- EncoderUZ isomorphic case: Skip linear projection when n_latent == n_latent_u, matching JAX

DE bug fix
- Changed mapped_inference_fn to extract "eps" instead of "z" — using "z" caused z_base to be double-counted in the LFC computation path via compute_h_from_x_eps

GPU fixes
- pz_scale and n_obs_per_sample registered as buffers so they move to GPU with the model
- Tensor creation in compute_local_statistics and differential_expression now uses correct device
- Removed unnecessary @torch.inference_mode() from _get_inference_input/_get_generative_input
- Changed z_u_prior_scale default from mutable torch.zeros(1) to float

Minor fixes
- MLP default activation: nn.ReLU (class) → nn.functional.relu (function)
- label_index: added .to(torch.int64) cast, fixed shape from 1D to 2D
- inference(): replaced stringly-typed type(x).__name__ check with isinstance
- Used inference_outputs["eps"] directly in loss instead of recomputing z - z_base

Parameter/loss equivalence tests Pytorch vs JAX:
image

DE Equivalence tests Pytorch vs JAX:
image


justjhong:
@ori-kron-wis having trouble with codecov if you have suggestions or if it looks like somethings broken with the coverage computation

ori-kron-wis:
> @ori-kron-wis having trouble with codecov if you have suggestions or if it looks like somethings broken with the coverage computation

Absolutely, something I had to fix sometime ago. It's about adding carry‑forward flags for codecov, because we have more labels now. Fix is already merged to main

ori-kron-wis (commented):
I validated the mrvi_tutorial we have and it's much closer to the original jax version. Looks great!

ori-kron-wis (approved):

ori-kron-wis:
@justjhong can  we merge?

PR #3756, opened 2026-03-21
fix: use persistent=False for n_obs_per_sample buffer in TorchMRVAE to avoid state_dict conflict on load of custom dataloader module

ori-kron-wis:


justjhong (approved):
Thanks!

PR #3763, opened 2026-03-25
fix(train): add flax to Jax import guard in _trainingplans

### LINKED ISSUE #3762: import scvi fails when jax+optax installed without flax (guard mismatch in _trainingplans.py)
### Bug

`import scvi` fails with `ImportError: cannot import name 'JaxBaseModuleClass'` when `jax` and `optax` are installed but `flax` is not.

### Cause

Guard mismatch between `_trainingplans.py` and `_base_module.py`:

- [`_trainingplans.py`](https://github.com/scverse/scvi-tools/blob/main/src/scvi/train/_trainingplans.py#L1617-L1622) guards the `JaxTrainingPlan` block with:
  ```python
  if is_package_installed("jax") and is_package_installed("optax"):
      from scvi.module.base._base_module import JaxBaseModuleClass, TrainStateWithState
  ```
- [`_base_module.py`](https://github.com/scverse/scvi-tools/blob/main/src/scvi/module/base/_base_module.py#L448) only defines `JaxBaseModuleClass` when:
  ```python
  if is_package_installed("jax") and is_package_installed("flax"):
  ```

When `jax + optax` are present (e.g., via `ott-jax`) but `flax` is not installed, `_trainingplans.py` tries to import `JaxBaseModuleClass` from `_base_module.py`, where it was never defined.

Note: `module/base/__init__.py` already handles this correctly via lazy `__getattr__` that checks for `flax`, but `_trainingplans.py` bypasses it by importing directly from `_base_module`.

### Suggested fix

Add `flax` to the guard in `_trainingplans.py`:

```python
if is_package_installed("jax") and is_package_installed("optax") and is_package_installed("flax"):
```

### Reproduction

```bash
pip install scvi-tools jax ott-jax  # installs jax + optax, but NOT flax
python -c "import scvi"
```

### Traceback

```
ImportError                               Traceback (most recent call last)
...
File .../scvi/train/_trainingplans.py:1622
   1619 import jax.numpy as jnp
   1620 import optax
-> 1622 from scvi.module.base._base_module import JaxBaseModuleClass, TrainStateWithState

ImportError: cannot import name 'JaxBaseModuleClass' from 'scvi.module.base._base_module'
```

### Versions

- scvi-tools: latest (pip)
- jax: <0.9.0
- optax: installed via ott-jax
- flax: NOT installed
- Python: 3.12

LiudengZhang:
# References and relevant issues

Closes #3762

## Description

When `jax` and `optax` are installed without `flax` (e.g., via `ott-jax`), `import scvi` fails because `_trainingplans.py` guards its Jax imports with only `jax + optax`, but the symbols it imports (`JaxBaseModuleClass`, `TrainStateWithState`) are defined in `_base_module.py` under a `jax + flax` guard.

This adds `flax` to the import guard in `_trainingplans.py` to match the upstream dependency, plus a regression test.

### DIFF
--- a/src/scvi/train/_trainingplans.py
+++ b/src/scvi/train/_trainingplans.py
@@ -1614,7 +1614,7 @@ class ClassifierTrainingPlan(pl.LightningModule):
         return optimizer
 
 
-if is_package_installed("jax") and is_package_installed("optax"):
+if is_package_installed("jax") and is_package_installed("optax") and is_package_installed("flax"):
     import jax
     import jax.numpy as jnp
     import optax
--- a/tests/train/test_trainingplans.py
+++ b/tests/train/test_trainingplans.py
@@ -3,7 +3,7 @@ import pytest
 import scvi
 from scvi.data import synthetic_iid
 from scvi.model import SCVI
-from scvi.train import TrainingPlan
+from scvi.train import TrainingPlan, _trainingplans
 from scvi.train._constants import METRIC_KEYS
 from scvi.train._trainingplans import _compute_kl_weight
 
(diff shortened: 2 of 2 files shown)


ori-kron-wis (approved):

ori-kron-wis:
Added it without the additional test

PR #3766, opened 2026-04-03
docs: updated for mlx

ori-kron-wis:


ori-kron-wis:
Added general updates to RTD following its update

PR #3776, opened 2026-04-12
fix: AttributeError after rich upgrade and MultiVI errors after Mudata upgrade

ori-kron-wis:


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3775
close https://github.com/scverse/scvi-tools/issues/3774
close https://github.com/scverse/scvi-tools/issues/3774

PR #3786, opened 2026-04-19
refactor: Dropping Jax from scvi-tools

ori-kron-wis:


ori-kron-wis:
We will need to transform Tangram to Torch before we remove Jax. Afterwards it will also move to scviva-tools @canergen

PR #3789, opened 2026-04-22
fix: differential abundance and aggregated posterior computations

lordy5:
Fixes:
- Use the original anndata for aggregated posterior computation, not the subset of the anndata. All cells in a sample should be considered when computing the aggregated posterior (up to num_cells_posterior)
- Update the tests accordingly
- Correctly pass scales into torch distributions, not variances. get_latent_representation returns variances, not scales, while the aggregated posterior code previously expected scales, and so accidentally passed variances into torch distributions (Normal and Student's T expect scales instead)


ori-kron-wis:
@lordy5, please remove the redundant .vscode or .claude

ori-kron-wis (approved):

PR #3799, opened 2026-05-06
🐛 fix: set LD_LIBRARY_PATH for Kubernetes GPU compatibility

FloWuenne:
On Kubernetes (GKE, EKS, AKS), the NVIDIA device plugin allocates GPUs via CDI and does not set NVIDIA_VISIBLE_DEVICES. Without it, the nvidia-container-runtime skips libcuda.so injection, causing torch.cuda.is_available() to return False even when a GPU is allocated.

GKE always mounts driver libraries at /usr/local/nvidia/lib64 regardless. Setting LD_LIBRARY_PATH to include that path lets PyTorch find libcuda.so without relying on runtime injection.

The existing value is preserved so HPC module systems (which manage their own LD_LIBRARY_PATH via `module load cuda`) are not affected.

An alternative fix is to rebase the image on nvidia/cuda:12.x-runtime-*
instead of python:3.12. NVIDIA base images set LD_LIBRARY_PATH and
NVIDIA_DRIVER_CAPABILITIES correctly out of the box, making the container
work on Kubernetes without any workaround.

To reproduce: 
Run scvi-tools on GKE with nvidia.com/gpu: 1 — model.to_device(0)
raises RuntimeError: Found no NVIDIA driver on your system.
Works on local Docker because --gpus all bypasses the CDI mechanism.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 FROM nvidia/cuda:12.4.0-runtime-ubuntu22.04
 FROM python:3.12 AS base
 
+ENV LD_LIBRARY_PATH=/usr/local/nvidia/lib64:/usr/local/nvidia/lib${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}
+
 RUN pip install --no-cache-dir uv
 
 RUN uv pip install --system --no-cache torch torchvision torchaudio


ori-kron-wis:
Thanks @FloWuenne, that makes a lot of sense to add.
Will it work in other K8S systems? (EKS, AKS) Not just GKE?

> An alternative fix is to rebase the image on nvidia/cuda:12.x-runtime-* instead of python:3.12. NVIDIA base images set LD_LIBRARY_PATH and NVIDIA_DRIVER_CAPABILITIES correctly out of the box, making the container work on Kubernetes without any workaround.

Can you suggest the fix in this case?

FloWuenne:
The fix I suggested should work on all k8s systems @ori-kron-wis . Appending the ENV variable should also make sure it doesn't break systems where the LD_LIBRARY_PATH is already set by the cuda libraries. 

The alternative as mentioned would be to build off base cuda to make sure this is all included by default but I can understand if you want the GPU build to be as close as possible to the cpu build in terms of setup.

ori-kron-wis:
Reverted as it made problems in our GitHub actions self hosted CUDA stations; Well reconsider again in the future

PR #3835, opened 2026-06-08
ci: bump codecov-action to v6 and fail CI on upload error

timtreis:
Two changes to every codecov upload step in this repo:

1. **codecov-action -> `@v6`.** Versions below 6.0.2 verify the codecov CLI binary against Codecov's old Keybase key, which they bricked after migrating accounts, so uploads now fail with `Could not verify signature`. `@v6` (= 6.0.2) ships the updated key.
2. **`fail_ci_if_error: true`.** Matches the `cookiecutter-scverse` template default so a broken upload fails CI loudly instead of passing silently.

Ref: https://github.com/codecov/codecov-action/issues/1956

### DIFF
--- a/.github/workflows/test_linux.yml
+++ b/.github/workflows/test_linux.yml
@@ -56,7 +56,8 @@ jobs:
           coverage run -m pytest -v --color=yes
           coverage report
 
-      - uses: codecov/codecov-action@v4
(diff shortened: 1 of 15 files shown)


ori-kron-wis:
thanks @timtreis !

PR #3866, opened 2026-06-21
feat: Add DRVI

moinfar:
## Add DRVI

This PR adds **DRVI** as `scvi.external.DRVI`.

This is a rewrite of #3699, reusing scvi-tools functionality and removing extras as much as possible.

### Design:

`DRVI` now subclasses `scvi.model.SCVI` directly and only changes what is actually different:

- Decoder is replaced by a splitting + additive decoder + pooling (`DecoderDRVI` extends `DecoderSCVI`, `SplitFCLayers` extends `FCLayers`). Disentanglement is induced in the decoder — the latent is reshaped into `n_split_latent` independent groups, each decoded separately, then aggregated.
- Adds log-space likelihood for negative binomial.
- Adds latent stats and interpretability mixins to interpret latent dimensions (plotting functions are removed).
- Reuses everything else from SCVI unchanged: encoder, prior, covariate/batch handling (including `batch_representation="embedding"`), RNASeqMixin, size factors, minified mode, scArches transfer, training via a datamodule, and `setup_anndata`.

To enable the decoder swap without forking `FCLayers`, `scvi.nn.FCLayers` gained a few small overridable seams (`_build_layer`, `_apply_layer`, `_apply_batch_norm`). These are behavior-preserving for existing models (SCVI, ...) —`SplitFCLayers` overrides them to use per-split weights.

### Likelihoods

In addition to scvi's `nb`, `zinb` and `poisson`, DRVI adds log-space nb:

- `pnb` (default of DRVI) — parametrized NB, same mean as `nb` but evaluated in log space via the new
  `LogNegativeBinomial` distribution (should be numerically more stable but equivalent to scvi's `NegativeBinomial`).
- `normal` / `normal_unit_var` — Gaussian with the mean modeled directly (no library/softmax), for
  continuous / log-normalized input.

### Interpretability

Because the decoder is additive over splits, each latent dimension's contribution to every gene can be measured directly. The included methods — `set_latent_dimension_stats`, `calculate_interpretability_scores`, `get_interpretability_scores` — quantify these contributions and extract top genes per dimension.

### Dependencies

Adds `stacked-linear>=0.2.0` (used for the parallel per-split linear layers). That package requires no dependencies other than PyTorch. Also, the same logic can be included as a snippet instead.


moinfar:
The code is ready for review.
Please let me know if you need any further explanation of the code's functionality.

Here are some TODOs in my list that will be issues in the next few days:
- A tutorial is prepared. A PR for that will be created.
- I am still not 100% sure on scArches compatibility. It runs, but extra tests will be added.

After the first iteration of review/changes, I will finally run the DRVI benchmark to make sure it performs the same.

src/scvi/external/drvi/_generative_mixin.py
```diff
@@ -0,0 +1,151 @@
...
+                        REGISTRY_KEYS.CAT_COVS_KEY: cat_tensor,
+                    },
+                    inference_outputs={
+                        MODULE_KEYS.Z_KEY: z_tensor,
+                        MODULE_KEYS.LIBRARY_KEY: library,
+                    },
+                )
+                yield self.module.generative(**gen_input)
+        finally:
+            self.module.inspect_mode = False
```
ori-kron-wis: The finally block resets inspect_mode but not the training/eval state. If the module was in train() mode when the generator is consumed (e.g., from a training loop or a test that doesn't call   
  model.train() beforehand), it stays in eval mode permanently. Fix: save was_training = self.module.training before eval() and restore with self.module.train(was_training) in finally.  

also self.module.fully_deterministic = False  - always, even if it was True before 

Is this behavior expected? what about saving/loading models? is this still holds?

src/scvi/nn/_base_components.py
ori-kron-wis, line 1: The changes are minimal, justified, and safe. But this should go in a separate preparatory PR (Im not sure it is mandatory for this drVI PR, no?)

src/scvi/external/drvi/_model.py
```diff
@@ -0,0 +1,107 @@
...
+        self,
+        adata: AnnData | None = None,
+        registry: dict | None = None,
+        n_latent: int = 32,
+        n_split_latent: int | None = None,
+        split_method: Literal["split_mask", "split_map"] = "split_map",
+        split_aggregation: Literal["mean", "logsumexp"] = "logsumexp",
+        gene_likelihood: Literal[
+            "nb", "pnb", "zinb", "poisson", "normal", "normal_unit_var"
+        ] = "nb",
```
ori-kron-wis: while its gene_likelihood="pnb" by default in the module, its "nb" in the model, is it ok?

src/scvi/external/drvi/_base_components.py
```diff
@@ -0,0 +1,295 @@
...
+    tensor of shape ``(n_obs, n_split, n_features)`` (one independent "split" per channel) instead
+    of the usual 2D ``(n_obs, n_features)``. Only small construction/application seams of
+    :class:`~scvi.nn.FCLayers` are overridden; ``forward`` and ``set_online_update_hooks`` (the
+    scArches hooks) are inherited unchanged.
+
+    Parameters
+    ----------
+    n_split
+        Number of parallel splits (channels) carried in the second tensor dimension.
+    reuse_weights
```
ori-kron-wis, line 30: perhaps better to validate the structure of reuse_weights and raise ValueError if fails

src/scvi/external/drvi/_base_components.py
```diff
@@ -0,0 +1,295 @@
...
+            self.activation_fn() if self.use_activation else None,
+            nn.Dropout(p=self.dropout_rate) if self.dropout_rate > 0 else None,
+        )
+
+    def _is_linear_layer(self, layer: nn.Module) -> bool:
+        return isinstance(layer, (nn.Linear, StackedLinearLayer))
+
+    def _apply_layer(self, layer, x, cov_list, layer_index):
+        if self._is_linear_layer(layer) and self.inject_into_layer(layer_index):
+            assert x.dim() in (3, 4), "SplitFCLayers works only with 3D tensors."
```
ori-kron-wis: Better to have ValueError with a message than an assert (which is mainly for tests)

src/scvi/external/drvi/_interpretability_mixin.py
```diff
@@ -0,0 +1,510 @@
...
+        log-fold-change style effects (``(n_splits, n_genes)`` or ``(2, n_splits, n_genes)`` if
+        directional). Requires ``embed.var`` to contain ``min`` and ``max``
+        (see :meth:`set_latent_dimension_stats`).
+        """
+        if self.module.n_latent != self.module.n_split_latent:
+            raise NotImplementedError(
+                "out-of-distribution interpretability requires one split per latent."
+            )
+
+        assert n_steps % 2 == 0, "n_steps must be even"
```
ori-kron-wis: Better to have ValueError with a message than an assert (which is mainly for tests)

src/scvi/external/drvi/_interpretability_mixin.py
```diff
@@ -0,0 +1,510 @@
...
+        dim_mins = np.minimum(embed.var["min"].values, 0.0)
+        dim_maxs = np.maximum(embed.var["max"].values, 0.0)
+
+        n_cat_total = [self.summary_stats.n_batch] + self._get_n_cats_per_cov()
+        all_cat_combinations = np.asarray(
+            list(itertools.product(*[range(n) for n in n_cat_total]))
+        )
+        all_cat_combinations = np.random.permutation(all_cat_combinations)[:n_samples]
```
ori-kron-wis, line 471: if we have a random here, I would also expect the option for a seed in sake of reproducibility, no?

src/scvi/external/drvi/_interpretability_mixin.py
```diff
@@ -0,0 +1,510 @@
...
+                    + torch.exp(directional_max_per_split)
+                )
+                - lse_min
+            )
+            min_possible = torch.log(
+                torch.exp(lse_max)
+                - torch.exp(max_per_split)
+                + torch.exp(directional_max_per_split)
+            ) - torch.log(torch.exp(lse_max) - torch.exp(max_per_split) + torch.exp(min_per_split))
+            combined = max_possible * min_possible
```
ori-kron-wis: is this metric and calculation explained somewhere in the docs?

ori-kron-wis (commented):
overall looks very good. several few points to think about

cvduffy:
Are there any notable differences between this implementation and the drvi-py standalone one? I trained models with drvi-py and am considering whether to re-train with this implementation once it's merged/released. Thanks!

moinfar:
Hi @cvduffy,

I'm not changing any of DRVI's logic for the scvi-tools port, which means no retraining will be needed!

I'll also do my best to add a function/snippet in drvi-py for backward compatibility, making it easy to move already-trained models over to the new scvi-tools version (but this may take a bit of time).

tests/external/drvi/test_drvi.py
```diff
@@ -0,0 +1,525 @@
...
+
+
+# def test_drvi_query_to_reference_mapping_embedding_batch():
+#     """With ``batch_representation="embedding"`` the reference stays intact but the query is NOT
+#     updated: scvi's scArches freezes the whole batch-embedding parameter (incl. the new query
+#     row). This is a core-scvi limitation (identical in :class:`~scvi.model.SCVI`), inherited by
+#     DRVI; use one-hot batches for query adaptation."""
+#     reference_change, query_change = _run_query_to_reference(batch_representation="embedding")
+#     assert reference_change < 1e-6  # reference embeddings intact
+#     assert query_change < 1e-6  # query frozen too (documented core-scvi limitation)
```
moinfar: @ori-kron-wis 
I opened a PR (#3879) to fix this in scvi. When merged, this will be uncommented

src/scvi/external/drvi/_base_components.py
moinfar: Validation added.

src/scvi/external/drvi/_interpretability_mixin.py
moinfar: Added seed param. sets seed if not none.

src/scvi/external/drvi/_generative_mixin.py
moinfar: This scenario does not happen in normal pipelines.
Fixed (revert to initial state in `finally`) to be on the safe side.

src/scvi/external/drvi/_model.py
moinfar: Thanks for noticing this. `pnb` should be the default. Fixed.

src/scvi/external/drvi/_base_components.py
moinfar: Done.

src/scvi/external/drvi/_interpretability_mixin.py
moinfar: Done.

src/scvi/external/drvi/_interpretability_mixin.py
moinfar: I've now added extensive explanations and docs for the interpretability mixin.

src/scvi/nn/_base_components.py
moinfar: Moved the changes on base components to #3880 

I merged that branch here (to keep the tests passing).
So, the changes will still be visible. When that branch is merged, it will go away from here.

moinfar (commented):
Thanks for reviewing the code.
I addressed all the comments / issues.

moinfar:
I will write here if I add or fix anything beyond the above discussions.

The last two commits have these two changes:

Commit e5f40d976194fbfc0971510e7bdc017a255aea15 fixes a bug where splits are multidimensional (minor scenario), and IND interpretability is used. It was not consistent with the directional single-dimension case.

Commit ddb3ba5211028a5d0d456aaa8566e34f2c76ba6d add mean_activation to enforce, for example, non-negativity to the latent space. I decided to also include this feature from the drvi-py since some models (e.g., the one we used for Human Cytokine Dictionary) already impose this.

ori-kron-wis:
@moinfar I merged https://github.com/scverse/scvi-tools/pull/3880 and https://github.com/scverse/scvi-tools/pull/3879 after some additions, added some considerations here for dropout rate and size_factor, and uncommented the scarches test function.
Perhaps you will want to take a look at it before we finally merge it.

Besides that, the tutorial is all that is missing now.

CHANGELOG.md
```diff
@@ -9,11 +9,14 @@ to [Semantic Versioning]. The full commit history is available in the [commit lo
 
 #### Added
 
-- Add scvi-tools MCP package that gives any MCP-compatible LLM access to scvi-tools knowledge
+- Add [scvi-tools MCP](%22https://scvi-tools-mcp.readthedocs.io/en/latest/index.html%22) package
```
moinfar: @ori-kron-wis 
Is this change intentional?

ori-kron-wis: you can keep

moinfar:
@ori-kron-wis 
Thanks for your efforts and for merging the required PRs.
I made a PR for a minimal tutorial: [https://github.com/scverse/scvi-tutorials/pull/539](https://github.com/scverse/scvi-tutorials/pull/539)
All your changes look good to me. I only left one comment.

Only two things from my side:
1. I will run a complete benchmark (should be finished by the end of the week) to make sure this port indeed produces the same output as drvi-py.
2. There is a decision that for DRVI, we anneal KL for max_epoch epochs (compared to the default 400 epochs in scvi-tools). I applied this behavior in the tutorial only (when calling model.train) to keep the changes in scvi-tools minimal. Would you suggest this approach or applying it as below?


```{py}
class DRVI(SCVI, GenerativeMixin, InterpretabilityMixin):
    ...
    _training_plan_cls = DRVITrainingPlan
    ...
    

class DRVITrainingPlan(TrainingPlan):
    """:class:`~scvi.train.TrainingPlan` that anneals the KL weight over the whole training run.

    DRVI's disentanglement objective benefits from spreading the KL warmup across the entire run,
    especially when the number of epochs is small (compare to the default of ``400``).
    With the default ``n_epochs_kl_warmup="auto"``, the warmup length is derived from the
    trainer's ``max_epochs`` at train time, so it matches the requested number of epochs.
    Passing an explicit ``n_epochs_kl_warmup`` (an ``int`` or ``None``) restores the standard
    behavior.
    """

    def __init__(self, *args, n_epochs_kl_warmup: int | str | None = "auto", **kwargs):
        self._auto_kl_warmup = n_epochs_kl_warmup == "auto"
        super().__init__(
            *args,
            n_epochs_kl_warmup=None if self._auto_kl_warmup else n_epochs_kl_warmup,
            **kwargs,
        )

    @property
    def kl_weight(self):
        """Scaling factor on KL divergence, warming up over ``max_epochs`` when ``"auto"``."""
        n_epochs_kl_warmup = self.n_epochs_kl_warmup
        if self._auto_kl_warmup:
            try:
                n_epochs_kl_warmup = self.trainer.max_epochs
            except RuntimeError:
                # Not yet attached to a trainer (``kl_weight`` is read once during __init__).
                # The exact warmup length is irrelevant here since ``current_epoch`` is 0; any
                # positive value yields ``min_kl_weight``.
                n_epochs_kl_warmup = 1
        klw = _compute_kl_weight(
            self.current_epoch,
            self.global_step,
            n_epochs_kl_warmup,
            self.n_steps_kl_warmup,
            self.max_kl_weight,
            self.min_kl_weight,
        )
        return torch.tensor(klw).to(self.device)

```

ori-kron-wis:
> @ori-kron-wis Thanks for your efforts and for merging the required PRs. I made a PR for a minimal tutorial: [scverse/scvi-tutorials#539](https://github.com/scverse/scvi-tutorials/pull/539) All your changes look good to me. I only left one comment.
> 
> Only two things from my side:
> 
> 1. I will run a complete benchmark (should be finished by the end of the week) to make sure this port indeed produces the same output as drvi-py.
> 2. There is a decision that for DRVI, we anneal KL for max_epoch epochs (compared to the default 400 epochs in scvi-tools). I applied this behavior in the tutorial only (when calling model.train) to keep the changes in scvi-tools minimal. Would you suggest this approach or applying it as below?
> 
> ```
> class DRVI(SCVI, GenerativeMixin, InterpretabilityMixin):
>     ...
>     _training_plan_cls = DRVITrainingPlan
>     ...
>     
> 
> class DRVITrainingPlan(TrainingPlan):
>     """:class:`~scvi.train.TrainingPlan` that anneals the KL weight over the whole training run.
> 
>     DRVI's disentanglement objective benefits from spreading the KL warmup across the entire run,
>     especially when the number of epochs is small (compare to the default of ``400``).
>     With the default ``n_epochs_kl_warmup="auto"``, the warmup length is derived from the
>     trainer's ``max_epochs`` at train time, so it matches the requested number of epochs.
>     Passing an explicit ``n_epochs_kl_warmup`` (an ``int`` or ``None``) restores the standard
>     behavior.
>     """
> 
>     def __init__(self, *args, n_epochs_kl_warmup: int | str | None = "auto", **kwargs):
>         self._auto_kl_warmup = n_epochs_kl_warmup == "auto"
>         super().__init__(
>             *args,
>             n_epochs_kl_warmup=None if self._auto_kl_warmup else n_epochs_kl_warmup,
>             **kwargs,
>         )
> 
>     @property
>     def kl_weight(self):
>         """Scaling factor on KL divergence, warming up over ``max_epochs`` when ``"auto"``."""
>         n_epochs_kl_warmup = self.n_epochs_kl_warmup
>         if self._auto_kl_warmup:
>             try:
>                 n_epochs_kl_warmup = self.trainer.max_epochs
>             except RuntimeError:
>                 # Not yet attached to a trainer (``kl_weight`` is read once during __init__).
>                 # The exact warmup length is irrelevant here since ``current_epoch`` is 0; any
>                 # positive value yields ``min_kl_weight``.
>                 n_epochs_kl_warmup = 1
>         klw = _compute_kl_weight(
>             self.current_epoch,
>             self.global_step,
>             n_epochs_kl_warmup,
>             self.n_steps_kl_warmup,
>             self.max_kl_weight,
>             self.min_kl_weight,
>         )
>         return torch.tensor(klw).to(self.device)
> ```

@moinfar Lets not change scvi-tools' n_epochs_kl_warmup right now to be reproducible (perhaps on a new branch) and add the additional DRVITrainingPlan class to DRVI

moinfar:
Thanks for the comments. It makes sense to keep changes minimal for now and set n_epochs_kl_warmup manually.

One update: I benchmarked this port against DRVI on drvi-py. It looked good, but the results were not identical. So, in this commit (632ee4860a144ab999e6a990d8061071234ad5d7), I fixed one difference that I missed: decoder should not have dropout. There is still one minor difference in the dispersion initialization that I prefer not to change to be more similar to scvi. But if the results were significantly different, I would change that as well and benchmark again.

I am running the benchmark again and will get back to you soon.

ori-kron-wis:
> I am running the benchmark again and will get back to you soon.

Do you think we can push it today or tomorrow? As we want to have it in the new release which is ready.

moinfar:
Yes. The branch can be merged.
Thanks for your efforts on this PR.

Details:
The benchmarking has finished. Now, there is no significant difference between `scvi.external.DRVI` and `drvi.model.DRVI`.
After this merge, I will publish a porting utility in drvi-py so users can port their already trained models to this implementation and deprecate the previous one.

ori-kron-wis:
@moinfar What about DRVITrainingPlan then? We keep it manually for now?

moinfar:
Yes. I think you also suggested keeping it manual. Right? I am totally fine with it.
Or did I misunderstand?

If you prefer the DRVITrainingPlan, I can create it in minutes (already a git patch).

ori-kron-wis:
> Yes. I think you also suggested keeping it manual. Right? I am totally fine with it. Or did I misunderstand?
> 
> If you prefer the DRVITrainingPlan, I can create it in minutes (already a git patch).

Sorry, yes, there was a misunderstanding - let's add DRVITrainingPlan as its a good practice to have it and it guarantees the correct usage as you described. 
I thought you meant an additional change in base scvi, but I guess not.

moinfar:
No worries. Thanks for the clarification.
I pushed the patch (22ee41acc3bd691dfd4d71e74002e212e820123b).

moinfar:
@cvduffy 
This will address your comment: https://github.com/theislab/DRVI/pull/130

Issue #1456, opened 2022-03-22
Within cluster DE test


zoesteier:
scVI used to have a within cluster DE test that I used often, but I no longer see this functionality in scvi-tools. Is it possible to add it back?

This would be a differential expression function where a "groupby" argument would specify a set of groups (e.g., cell clusters) and a "states" argument would specify the relevant comparison to make (e.g., case vs control). The output would include the DE test results from a comparison of cases vs controls within each specified group.

adamgayoso:
@PierreBoyeau what do you think?

PierreBoyeau:
It sounds like it would be relevant to include it back. Do you remember why we removed this feature?

PR #3871, opened 2026-06-22
docs: add within-cluster differential expression example

joeydufourd:
Addresses #1456.

This adds a documentation example showing how to perform within-cluster differential expression using the existing `idx1` and `idx2` arguments to `differential_expression()`.

The example loops over cell types, compares two conditions within each cell type, and concatenates the per-cluster differential expression results.

Checked locally with:

```bash
git diff --check
```

### DIFF
--- a/docs/user_guide/background/differential_expression.md
+++ b/docs/user_guide/background/differential_expression.md
@@ -46,9 +46,9 @@ This guide has two aims.
 First, it aims to provide insight as to how scVI-tools' differential expression module works for transcript expression (`scVI`), surface protein expression (`TOTALVI`), or chromatin accessibility (`PeakVI`).
 More precisely, we explain how it can:
 
--   approximate population-specific normalized expression levels
--   detect biologically relevant features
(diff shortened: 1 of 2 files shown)


ori-kron-wis:
@joeydufourd thank you for the addition. 

I think we mostly describe the theory of DE in the file that you changed, like everything else under docs/user_guide/background/, and do not show code parts, so I think it's less relevant to show that example there, which is also a specific use case of running DE.

We usually show code examples as part of a tutorial or as an answer to user questions here or in our discourse forum.

Another place that can show your addition is here: https://docs.scvi-tools.org/en/latest/user_guide/use_case/downstream_analysis_tasks.html (2nd bullet) - we can rewrite this part and show more code examples, so please go ahead and make the changes there.

joeydufourd:
Thanks for the guidance! I’ll move the example out of the background/theory section and adapt it into the downstream analysis tasks page instead, around the differential expression use case.

Issue #3674, opened 2026-01-15
Resolvi.load overflows memory


jairaj-mathur:
When I load a saved model via RESOLVI.load, the memory consumption to just load the model is higher than while the model was being trained. I am using a 256gb ram and 22 GB GPU. While loading, the model, the GPU only uses about 200 MB VRAM while the RAM overshoots the 256 GB limit. During training, model used about 180 GB RAM and 2-5 GB VRAM. 

Xenium dataset, 200k cells 5k genes

```python
RESOLVI.load(/path/to/model, data, accelerator='gpu')
```

```pytb
The python process exited with exit code 137 (SIGKILLED: Killed). This may have been caused by an OOM error. Check your commands memory usage.
```

#### Versions:

> 1.4.1

canergen:
I actually assumed this was fixed. Did you run with the most recent scvi-tools version?

ori-kron-wis:
@jairaj-mathur please use the branch I suggested in the other issue, see if this issue is solved as well.

LizEve:
@ori-kron-wis I'm a coworker of @jairaj-mathur. I think our issue is that Databricks upgraded to scipy==1.17.0 recently. Setting scipy==1.16.3 seems to solve my OOM issue when loading models.

ori-kron-wis:
@LizEve Thanks for letting us know.
I can confirm the memory footprint increased by 10% on my test of loading resolvi model with scipy 1.17.

canergen:
@ori-kron-wis do we need a fix for it or do we hope scipy will fix it upstream? My intuition is that the error comes from https://github.com/scverse/scvi-tools/blob/54fd6135263b02d803ff07fa653267d3635ca51a/src/scvi/model/base/_archesmixin.py#L240.

ori-kron-wis:
It's up to you, I guess. It's a minor bug, not affecting logic. 
If we can improve the memory consumption of this or any other model, of course, it's a good thing.

Qiaoli-Li-Res:
I identified two sources of excess memory during `RESOLVI.load()` and opened a PR in #3886:

1. `setup_anndata()` was re-running `_prepare_data()` on every load, causing an unnecessary `adata.copy()` + `sc.pp.neighbors()` call.
2. `on_load()` was running `train(max_steps=1)` unnecessarily — `RESOLVAE` uses a manually-written guide, so the Pyro param store can be restored directly.

[...]

PR #3886, opened 2026-06-26
fix(resolvi): reduce peak memory during RESOLVI.load()

Qiaoli-Li-Res:
## Problem

Addresses  #3674

`RESOLVI.load()` consumed significantly more RAM than training, causing OOM on large datasets (e.g., 200k cells Xenium). Two sources of unnecessary memory allocation were identified:

### Fix 1: `setup_anndata()` re-ran `_prepare_data()` on every load

During `load()`, `setup_anndata()` is called with the original saved arguments (`prepare_data=True`). This re-ran `_prepare_data()` even when `index_neighbor` and `distance_neighbor` already existed in `adata.obsm`, causing:
- A full `adata.copy()` of all cells
- `sc.pp.neighbors()` recomputation on the full dataset

**Fix:** Skip `_prepare_data()` when neighbors are already valid (keys exist and indices are within bounds of the current adata).

### Fix 2: `on_load()` ran `train(max_steps=1)` unnecessarily

The generic `PyroBaseModuleClass.on_load()` runs `model.train(max_steps=1)` to initialize Pyro param shapes before restoring the saved param store. However, `RESOLVAE` uses a manually-written guide with statically known parameter names and shapes — `pyro.get_param_store().set_state()` is self-contained and does not require a prior forward pass.

**Fix:** Add a lightweight `on_load()` to `RESOLVAE` that directly restores the Pyro param store without a training step.

## Verification

Tested with synthetic data (20k cells × 2k genes):

| Version | tracemalloc peak |
|---|---:|
| Original | 0.109 GB |
| Patched | 0.036 GB |

- tracemalloc peak allocation reduced by **66%**
- Unnecessary `train(max_steps=1)` call eliminated during load
- All existing RESOLVI tests pass (14/14)
- Latent representations are identical before and after load

> **Note:** RSS-level difference requires 100k+ cells to be measurable. Full memory validation with the original reporter's Xenium dataset (200k cells) is recommended to confirm complete OOM resolution.

### DIFF
--- a/src/scvi/external/resolvi/_model.py
+++ b/src/scvi/external/resolvi/_model.py
@@ -355,7 +355,13 @@ class RESOLVI(
         if prepare_data:
             if prepare_data_kwargs is None:
                 prepare_data_kwargs = {}
-            RESOLVI._prepare_data(adata, batch_key=batch_key, **prepare_data_kwargs)
+            neighbors_valid = (
+                "index_neighbor" in adata.obsm
+                and "distance_neighbor" in adata.obsm
+                and int(adata.obsm["index_neighbor"].max()) < adata.n_obs
+            )
+            if not neighbors_valid:
+                RESOLVI._prepare_data(adata, batch_key=batch_key, **prepare_data_kwargs)
 
         anndata_fields = [
             LayerField(REGISTRY_KEYS.X_KEY, layer, is_count_data=True),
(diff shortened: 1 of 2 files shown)


ori-kron-wis:
copy of https://github.com/scverse/scvi-tools/pull/3887

Issue #3980, opened 2026-09-01
MPS accuracy warning in parse_device_args() is not supported by measurement


BenjaminDEMAILLE:
### Summary

`parse_device_args()` warns that MPS models "might be slower and **less accurate** than
usual" ([`src/scvi/model/_utils.py:129`](https://github.com/scverse/scvi-tools/blob/main/src/scvi/model/_utils.py#L129)).
I measured this and could not reproduce an accuracy penalty: on the ops that matter for
SVI, MPS float32 is indistinguishable from CPU float32 relative to a float64 reference,
and on the ELBO gradient it was marginally *closer* to the reference.

The warning was added in #2045 (May 2023), when MPS was much less mature, and it points
at [pytorch#77764](https://github.com/pytorch/pytorch/issues/77764), which tracks
**operator coverage**, not numerics. I think the coverage half is still worth warning
about and the accuracy half is no longer supported by evidence.

### Measurements

M4 Max, macOS 26.6.2, torch 2.10.0, pyro 1.9.1. Max absolute error against a CPU
float64 reference, 4096x512 inputs:

| Op | CPU fp32 | MPS fp32 |
|---|---|---|
| `lgamma` | 3.6e-07 | 8.9e-07 |
| `digamma` | 3.9e-06 | 5.4e-06 |
| `log` | 1.8e-07 | 3.3e-07 |
| `exp` | 1.6e-05 | 1.9e-05 |
| `softmax(-1)` | 4.3e-09 | 3.6e-09 |
| `sum(-1)` | 1.8e-04 | 1.7e-04 |
| `matmul` 1024x1024 | 2.07e-06 | 2.07e-06 |

The figure that actually governs training, the gradient of a negative-binomial ELBO
term w.r.t. an AutoNormal location (262k parameters, gradient scale 125):

```
cpu32   max relative deviation from float64   3.8e-07
mps32   max relative deviation from float64   3.7e-07
```

Transcendentals are 1.2x-2.5x further from the reference on MPS, i.e. a ULP or two.
Reductions and matmul are indistinguishable. This is float32 noise on both sides.

<details>
<summary>Reproducer</summary>

```python
import torch, pyro.distributions as dist

def cmp(name, fn, shape=(4096, 512)):
    a64 = torch.rand(*shape, dtype=torch.float64) * 4 + 0.1
    a32 = a64.float()
    ref, cpu = fn(a64), fn(a32).double()
    mps = fn(a32.to("mps")).cpu().double()
[...]

PR #3981, opened 2026-09-01
Take the MPS CPU detour only when the torch build needs it

BenjaminDEMAILLE:
## What this changes

`src/scvi/distributions/_negative_binomial.py` sends Gamma and Poisson to the CPU
whenever the distribution's parameters are on `mps`. That is hardcoded, and the code
says why:

```python
Gamma(concentration=concentration.to("cpu"), rate=rate.to("cpu"))
if on_mps  # TODO: NEED TORCH MPS FIX for 'aten::_standard_gamma'
```
```python
PoissonTorch(l_train).sample().to("mps")
if self.on_mps  # TODO: NEED TORCH MPS FIX for 'aten::poisson'
```

Both TODOs are now stale. `aten::_standard_gamma` gained an MPS kernel in torch 2.12
(pytorch/pytorch#179228) and `aten::poisson` in 2.14 (pytorch/pytorch#173319). On a
current torch, every `NegativeBinomial.sample()` pays two device transfers to reach a
CPU kernel that is no longer the only option.

This makes the detour conditional on what the running build actually provides. Where the
kernel is missing, the behaviour is exactly what it is today.

## Why probe instead of comparing `torch.__version__`

`torch` is an unpinned dependency here (`pyproject.toml`), so this code runs against
whatever the user installed, including builds whose version string a comparison reads
wrongly. A source build reports `2.14.0a0+gitXXXXXXX`, which sorts *below* `2.14.0` and
would disable a kernel that is present; nightlies have the mirror problem. So
`_mps_supports(op)` calls the op once on a one-element `mps` tensor, caches the answer,
and treats `NotImplementedError` as "not available". It is only ever reached from a
distribution whose parameters are already on `mps`.

The two kernels are probed separately, because torch 2.12 and 2.13 have Gamma but not
Poisson, and that intermediate case is worth getting right rather than rounding down.

## Measurements

`NegativeBinomial(mu, theta)` with `mu`, `theta` of shape (4000, 2000), `sample()`, mean
of 5 runs on an M-series Mac:

| torch | before | after |
|---|---|---|
| 2.14.0 | 0.399 s | **0.004 s** |
| 2.13.0 | 0.399 s | 0.283 s |
| cpu reference | 0.397 s | |

Before the change, the `mps` figure equalling the `cpu` figure is the tell: all the work
was already on the CPU. 2.13 improves only partway because it has the Gamma kernel and
not the Poisson one, which is what the split probe is for.

Sample mean and standard deviation match the analytic values to the tolerance the
existing test uses.

## Also fixed

`NegativeBinomialMixture.sample()` returned a **CPU** tensor for parameters on `mps`,
unlike `NegativeBinomial.sample()` directly above it, which moved its result back. Both
now return on the device the parameters came from.

## Deliberately left alone

`torch_lgamma_mps` and the `.contiguous()` calls in `__init__`. They work around a
different problem, a broadcasting issue rather than a missing kernel, and separating
them deserves its own check rather than riding along here.

## Testing

```
python -m pytest tests/distributions/ -q          # 13 passed
ruff check src/scvi/distributions/_negative_binomial.py tests/distributions/test_negative_binomial.py
ruff format --check src/scvi/distributions/_negative_binomial.py tests/distributions/test_negative_binomial.py
```

Run against three torch builds, since the point of the change is that behaviour depends
on the build: 2.14.0 (both kernels), 2.13.0 (Gamma only), and with `_mps_supports` forced
to return `False`, which reproduces the current behaviour exactly and stands in for a
pre-2.12 build.

Two tests are added. They skip without an MPS device, so they will not run in CI as it
stands; `test_gamma_stays_on_mps_when_the_kernel_exists` asserts against the same probe
rather than against a version, so it stays correct on any torch.

---

*Written with the help of an AI assistant; the diff and the measurements above were
reviewed by hand.*

### DIFF
--- a/src/scvi/distributions/_negative_binomial.py
+++ b/src/scvi/distributions/_negative_binomial.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import warnings
+from typing import TYPE_CHECKING
 
 import torch
 import torch.nn.functional as F
@@ -16,7 +17,10 @@ from torch.distributions.utils import (
 from scvi import settings
 
 from ._constraints import optional_constraint
-from ._utils import _needs_cpu_detour
+from ._utils import _mps_supports_lgamma_on_noncontiguous, _needs_cpu_detour
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
 
 
 def torch_lgamma_mps(x: torch.Tensor) -> torch.Tensor:
@@ -32,6 +36,13 @@ def torch_lgamma_mps(x: torch.Tensor) -> torch.Tensor:
     lgamma tensor that performs on a copied version of the tensor
     """
     return torch.lgamma(x.contiguous())
+
+
+def _lgamma_fn(on_mps: bool) -> Callable[[torch.Tensor], torch.Tensor]:
+    """Pick the lgamma implementation, using the contiguous-copy detour only if still needed."""
+    if on_mps and not _mps_supports_lgamma_on_noncontiguous():
+        return torch_lgamma_mps
+    return torch.lgamma
 
 
 def log_zinb_positive(
@@ -448,7 +459,7 @@ class NegativeBinomial(Distribution):
                     stacklevel=settings.warnings_stacklevel,
                 )
 
-        lgamma_fn = torch_lgamma_mps if self.on_mps else torch.lgamma  # TODO: TORCH MPS FIX
+        lgamma_fn = _lgamma_fn(self.on_mps)
         return log_nb_positive(
             value, mu=self.mu, theta=self.theta, eps=self._eps, lgamma_fn=lgamma_fn
         )
@@ -575,7 +586,7 @@ class ZeroInflatedNegativeBinomial(NegativeBinomial):
                 UserWarning,
                 stacklevel=settings.warnings_stacklevel,
             )
-        lgamma_fn = torch_lgamma_mps if self.on_mps else torch.lgamma  # TODO: TORCH MPS FIX
+        lgamma_fn = _lgamma_fn(self.on_mps)
         return log_zinb_positive(
             value, self.mu, self.theta, self.zi_logits, eps=1e-08, lgamma_fn=lgamma_fn
         )
@@ -688,7 +699,7 @@ class NegativeBinomialMixture(Distribution):
                 UserWarning,
                 stacklevel=settings.warnings_stacklevel,
             )
-        lgamma_fn = torch_lgamma_mps if self.on_mps else torch.lgamma  # TODO: TORCH MPS FIX
+        lgamma_fn = _lgamma_fn(self.on_mps)
         return log_mixture_nb(
             value,
             self.mu1,
--- a/src/scvi/distributions/_negative_binomial.py
+++ b/src/scvi/distributions/_negative_binomial.py
@@ -1,8 +1,6 @@
 from __future__ import annotations
 
 import warnings
-from functools import cache
-from typing import TYPE_CHECKING
 
 import torch
 import torch.nn.functional as F
@@ -18,9 +16,7 @@ from torch.distributions.utils import (
 from scvi import settings
 
 from ._constraints import optional_constraint
-
-if TYPE_CHECKING:
-    from collections.abc import Callable
+from ._utils import _needs_cpu_detour
 
 
 def torch_lgamma_mps(x: torch.Tensor) -> torch.Tensor:
@@ -36,30 +32,6 @@ def torch_lgamma_mps(x: torch.Tensor) -> torch.Tensor:
     lgamma tensor that performs on a copied version of the tensor
     """
     return torch.lgamma(x.contiguous())
-
-
-@cache
-def _mps_supports(op: Callable[[torch.Tensor], torch.Tensor]) -> bool:
-    """Whether ``op`` has an MPS kernel in the torch build in use.
-
-    ``aten::_standard_gamma`` gained one in torch 2.12 and ``aten::poisson`` in
-    2.14, so the answer depends on the installed version. It is probed rather
-    than version-compared because source and nightly builds report versions
-    that a comparison reads wrongly.
-
-    Callers must check ``torch.backends.mps.is_available()`` first; this is only
-    reached from a distribution whose parameters already live on ``mps``.
-    """
-    try:
-        op(torch.ones(1, device="mps"))
-    except (NotImplementedError, RuntimeError):
-        return False
-    return True
-
-
-def _needs_cpu_detour(on_mps: bool, op: Callable[[torch.Tensor], torch.Tensor]) -> bool:
-    """Whether sampling ``op`` has to be routed through the CPU."""
-    return on_mps and not _mps_supports(op)
 
 
 def log_zinb_positive(
--- a/src/scvi/distributions/_negative_binomial.py
+++ b/src/scvi/distributions/_negative_binomial.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import warnings
+from functools import cache
+from typing import TYPE_CHECKING
 
 import torch
 import torch.nn.functional as F
@@ -17,6 +19,9 @@ from scvi import settings
 
 from ._constraints import optional_constraint
 
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
 
 def torch_lgamma_mps(x: torch.Tensor) -> torch.Tensor:
     """Used in Mac Mx devices while broadcasting a tensor
@@ -31,6 +36,30 @@ def torch_lgamma_mps(x: torch.Tensor) -> torch.Tensor:
     lgamma tensor that performs on a copied version of the tensor
     """
     return torch.lgamma(x.contiguous())
+
+
+@cache
+def _mps_supports(op: Callable[[torch.Tensor], torch.Tensor]) -> bool:
+    """Whether ``op`` has an MPS kernel in the torch build in use.
+
+    ``aten::_standard_gamma`` gained one in torch 2.12 and ``aten::poisson`` in
+    2.14, so the answer depends on the installed version. It is probed rather
+    than version-compared because source and nightly builds report versions
+    that a comparison reads wrongly.
+
+    Callers must check ``torch.backends.mps.is_available()`` first; this is only
+    reached from a distribution whose parameters already live on ``mps``.
+    """
+    try:
+        op(torch.ones(1, device="mps"))
+    except (NotImplementedError, RuntimeError):
+        return False
+    return True
+
+
+def _needs_cpu_detour(on_mps: bool, op: Callable[[torch.Tensor], torch.Tensor]) -> bool:
+    """Whether sampling ``op`` has to be routed through the CPU."""
+    return on_mps and not _mps_supports(op)
 
 
 def log_zinb_positive(
@@ -273,13 +302,10 @@ def _convert_counts_logits_to_mean_disp(
 def _gamma(theta: torch.Tensor, mu: torch.Tensor, on_mps: bool = False) -> Gamma:
     concentration = theta
     rate = theta / mu
+    if _needs_cpu_detour(on_mps, torch._standard_gamma):
+        concentration, rate = concentration.to("cpu"), rate.to("cpu")
     # Important remark: Gamma is parametrized by the rate = 1/scale!
-    gamma_d = (
-        Gamma(concentration=concentration.to("cpu"), rate=rate.to("cpu"))
-        if on_mps  # TODO: NEED TORCH MPS FIX for 'aten::_standard_gamma'
-        else Gamma(concentration=concentration, rate=rate)
-    )
-    return gamma_d
+    return Gamma(concentration=concentration, rate=rate)
 
 
 class Poisson(PoissonTorch):
@@ -378,9 +404,8 @@ class NegativeBinomial(Distribution):
         scale: torch.Tensor | None = None,
         validate_args: bool = False,
     ):
-        self.on_mps = (
-            mu.device.type == "mps" if total_count is None else total_count.device.type == "mps"
-        )  # TODO: This is used until torch will solve the MPS issues
+        self._device = mu.device if total_count is None else total_count.device
+        self.on_mps = self._device.type == "mps"
         self._eps = 1e-8
         if (mu is None) == (total_count is None):
             raise ValueError(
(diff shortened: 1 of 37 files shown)


src/scvi/distributions/_negative_binomial.py
```diff
@@ -33,6 +38,30 @@ def torch_lgamma_mps(x: torch.Tensor) -> torch.Tensor:
     return torch.lgamma(x.contiguous())
 
 
+@cache
```
ori-kron-wis: Please move the 2 new functions to:
scvi/distributions/utils.py and use it from there in the current function

ori-kron-wis (commented):
@BenjaminDEMAILLE Thank you so much for supporting the MPS/torch integration in scvi-tools and for showing this amazing speed improvement in MPS for torch 2.14
I left a small comment here. Please also add the warning updates from the issue https://github.com/scverse/scvi-tools/issues/3980

I will also test and merge it when Torch releases 2.14

src/scvi/distributions/_negative_binomial.py
BenjaminDEMAILLE: Done

ori-kron-wis: I dont see it, is it pushed?

BenjaminDEMAILLE: Pushed now in 751af908 - my earlier reply was premature, the commit had not left my machine.

Both functions are now in `src/scvi/distributions/_utils.py` and imported from there. I used `_utils.py` rather than `utils.py` because that module already exists in this package (alongside `_constraints.py`) and holds `subset_distribution` and `DistributionConcatenator`. Happy to rename it to a public `utils.py` if you would rather the distributions helpers be part of the public surface.

The warning rewording from #3980 is in the same commit, in `parse_device_args`, with a CHANGELOG entry under Changed.

ori-kron-wis: Thank you, _utils.py is fine.

BenjaminDEMAILLE:
One question on timing: which release are you aiming this at? I put the CHANGELOG entries under 1.5.1, so let me know if they should move.

I ask because the two halves have different constraints. The warning rewording from #3980 stands on its own and does not depend on any torch version. The sampling change is gated by your plan to test and merge once torch 2.14 is out - though the probe means it is already correct on older builds, where it simply keeps today's CPU detour.

If 1.5.1 goes out before torch 2.14 lands, I am happy to split the warning change into its own PR so it is not held up.

ori-kron-wis:
> One question on timing: which release are you aiming this at? I put the CHANGELOG entries under 1.5.1, so let me know if they should move.
> 
> I ask because the two halves have different constraints. The warning rewording from #3980 stands on its own and does not depend on any torch version. The sampling change is gated by your plan to test and merge once torch 2.14 is out - though the probe means it is already correct on older builds, where it simply keeps today's CPU detour.
> 
> If 1.5.1 goes out before torch 2.14 lands, I am happy to split the warning change into its own PR so it is not held up.

Its fine. 1.5.1 is planned for next week, and torch 2.14 is already out, so Im looking into this now.

ori-kron-wis:
@BenjaminDEMAILLE I extended your work to treat all other models/functions to work in MPS with the updated torch version (lots of files and tests, but minor fixes for each)

Can you verify your tests again with this branch?

ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/3980

BenjaminDEMAILLE:
Verified on this branch. M4 Max, macOS 26.6.2, python 3.13.15, torch 2.14.0, run the way `test_mps.yaml` runs it:

```
uvx hatch run test-metal:run-cov -v --accelerator mps --devices auto
  (SCVI_ALLOW_MPS_AUTO=1, PYTORCH_ENABLE_MPS_FALLBACK=1, PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0)

889 passed, 142 skipped
```

Green on the branch as it stood at 4c73605e. Your extension looks right to me on 2.14.

One thing it does not cover, pushed as 6eddffcc. `_mps_supports_lgamma_on_noncontiguous` only checks that `torch.lgamma` does not raise, but the bug `torch_lgamma_mps` exists for is not an exception. On torch 2.10, lgamma on a stride-0 expanded mps tensor returns the right numbers for the first row and `inf` for the replicated ones:

```
x = (torch.rand(1, 4) * 3 + 0.1).to("mps")
torch.lgamma(x.expand(4, 4))
tensor([[-0.1140, 0.2199, 0.8898, 0.5801],
        [    inf,    inf,    inf,    inf],
        [    inf,    inf,    inf,    inf],
        [    inf,    inf,    inf,    inf]])
```

So the probe reports support that is not there and `_lgamma_fn` drops the detour while the bug is still present. `NegativeBinomial` is fine because `__init__` already copies its parameters contiguously on mps, but `NegativeBinomialMixture` does not, so `broadcast_all` leaves a per-gene `theta1` as an expanded view that reaches `lgamma` directly. That is the shape totalVI uses:

```
NegativeBinomialMixture(mu1=(4,6), mu2=(4,6), theta1=(6,), mixture_logits=(4,6)).log_prob(x)

                                torch 2.10    torch 2.14
cpu                             -131.59239    -131.59239
mps, branch as it stood         -inf          -131.59230
mps, with the detour forced     -131.59232    -131.59230
```

2.14 has the kernel fixed, which is why the suite above is green. It bites only the older builds the probe is there to support in the first place.

The commit makes the probe compare the expanded result against the contiguous one, and adds two regression tests: one on the probe, one on the mixture log_prob. Both fail on torch 2.10 before the change and pass after. `tests/distributions` is 32 passed on 2.14 and 31 passed 1 skipped on 2.10.

I have not re-run the full suite since the rebase onto 458c05ca, only `tests/distributions`; the merge did not touch the distributions.

ori-kron-wis:
@BenjaminDEMAILLE  Thank you for this additional fix.

I used this branch and ran the introduction tutorial of scvi-tools (https://docs.scvi-tools.org/en/latest/tutorials/notebooks/quick_start/api_overview.html) on a MAC M3 PRO machine, compared CPU vs MPS acceleration timings for the training section of it.

While everything works, as do all tests passes here, the actual performance enhancement is not that great (not more than 10%), way far than running CPU vs CUDA GPU on Linux machine (even a medium GPU is 100% faster) It probably means that there are still CPU fallback. 

What do you think of it? Do you see similar results on M4 MAX? Before I merge this branch, I want a proof of the gain of it.

BenjaminDEMAILLE:
You are right about the training section, and I think the benchmark is pointing at the wrong thing. Numbers below are M4 Max, 16 cores, torch 2.14.0, the tutorial dataset (18641 x 1200), 20 epochs, machine otherwise idle.

**Training: your 10% is real, and it is not a CPU fallback**

```
                      s/epoch    final ELBO
cpu                    1.239       309.49
mps                    0.976       309.13
```

1.27x, same order as what you saw on the M3 Pro. But it is not a fallback: running the whole training with `PYTORCH_ENABLE_MPS_FALLBACK=0` completes without error, so no op in the training graph is silently going through the CPU. Every remaining `_needs_cpu_detour` call site is in a sampling path (`_vae.py`, `_rnamixin.py`, `_totalvi.py`, `_autozivae.py`, velovi, drvi, `_preprocessing.py`), never in `train()`.

The reason is size. Splitting an epoch at the default `batch_size=128`:

```
dataloader only                 0.037 s/epoch   (146 batches, negligible)
fwd+bwd on a resident batch     cpu 7.49 ms   mps 4.73 ms   ->  1.58x
```

146 x 4.73 ms = 0.69 s of compute inside a 0.976 s epoch; the rest is Python and Lightning per-step overhead, which is identical on both devices. The GPU is idle most of the step. It scales the way you would expect once the batch is big enough to be worth a dispatch:

```
fwd+bwd     batch=128     cpu  7.49 ms   mps  4.73 ms    1.58x
fwd+bwd     batch=1024    cpu 18.14 ms   mps  5.62 ms    3.23x
fwd+bwd     batch=4096    cpu 59.19 ms   mps 12.58 ms    4.70x
```

**The training section cannot measure this PR either way**

```
                              s/epoch
mps, main before this PR        0.979
mps, this branch                0.976
```

Identical, and that is expected: training only ever calls `log_prob`, never `sample`. Nothing this PR changes is on that path apart from the lgamma copy.

**Where the gain actually is**

```
                                          main       branch     
NegativeBinomial.sample() 4000x2000      431.6 ms     4.6 ms    94x
posterior_predictive_sample 2000x1200    162.1 ms    72.8 ms     2.2x

cpu reference, same sample()             426.5 ms
```

The last line is the point of the whole PR: on main the mps figure (431.6) equals the cpu figure (426.5), because the work was going to the CPU and back. `posterior_predictive_sample` is what `scvi.criticism` is built on.

So I would not gate the merge on training throughput. If you want the tutorial to show a GPU-shaped speedup, that is a separate change (a larger default `batch_size` on accelerators), and I am happy to open it. Tell me if you want the 400-epoch tutorial default measured rather than my 20, or the same table on another model.

BenjaminDEMAILLE:
Follow-up on the throughput question, since it turned out to have a fix rather than just an explanation: #3998.

The default Adam path accounts for about 423 of the 916 aten ops a training step issues, and for 17 of its 22 host-device synchronisations, because the single-tensor path reads its step counter back to the host once per parameter. Using the fused implementation where the backend supports it:

```
        default        fused
mps     0.989 s/ep     0.697 s/ep     1.42x
cpu     1.138 s/ep     1.050 s/ep     1.08x
```

ELBO unchanged (309.13 against 309.14). It helps mps far more than cpu, so the mps advantage on the tutorial goes from 1.15x to 1.51x. That is most of the gap you were pointing at, and it is independent of this PR.

That one also fixes `compile=True` leaving `torch._dynamo.config.suppress_errors` on for the rest of the process, and warns when compilation silently falls back to eager. I nearly reported a wrong number to you because of that before counting the compiled frames.

Nothing there blocks this PR; keeping them separate on purpose.

BenjaminDEMAILLE:
Closing the loop on the fallback question, since "no op falls back" deserves more than the training loop as evidence. Full suite on this branch with the fallback switch turned off:

```
PYTORCH_ENABLE_MPS_FALLBACK=0
uvx hatch run test-metal:pytest --accelerator mps --devices auto

891 passed, 142 skipped, 0 failed
```

Not one `NotImplementedError`, so on torch 2.14 every op the suite exercises has an MPS kernel and nothing is quietly taking the CPU route. Worth keeping in mind that CI runs with `PYTORCH_ENABLE_MPS_FALLBACK=1`, which would hide exactly this, so a green mps job on its own would not have told you.

Two things that confirm the run is sound rather than accidentally lenient: 891 against the 889 I reported earlier is the regression tests added since, and both `test_datasplitter_load_sparse` pass here while they fail on `main` with `Could not run 'new_compressed_tensor' from the 'mps:0' device` - your 4c73605e is what fixes them.

M4 Max, python 3.13.15, torch 2.14.0.

BenjaminDEMAILLE:
Thanks for taking the time to benchmark it rather than just running the tests.

Two things, and I think only the second one is about this branch.

**On the 10%.** I don't think that number points at a CPU fallback. With the `SCVI` defaults the tutorial uses (`n_hidden=128`, `n_layers=1`, `batch_size=128`), a training step is a pair of 128x2000 by 2000x128 GEMMs plus elementwise work. At that size the Metal backend is bound by per-kernel dispatch latency, not by arithmetic throughput, so most of the wall clock is command submission that the CPU path simply doesn't pay. The CUDA comparison is not like-for-like here: launch overhead there is much lower, and a "medium GPU" on a Linux box is being fed the same tiny matrices. I would expect the MPS/CPU ratio to move a lot with `batch_size` and `n_hidden` -- if you have the notebook open, running the training section at `batch_size=1024` and `n_hidden=512` would tell us quickly whether we are latency-bound or genuinely falling back.

**What is genuinely still missing on MPS upstream**, from working through it:

- Sparse compressed (CSR/CSC/BSR/BSC) tensors did not exist on the MPS backend at all. That is what `_mps_supports_sparse_compressed_tensor()` probes, and on any released torch it returns `False`, so `transfer_batch_to_device` densifies on the host and `load_sparse_tensor=True` is effectively a no-op on Mac. I have this working locally and am preparing the upstream PRs.
- Even with construction enabled, the mathematical operations on a compressed tensor (`add`, `addmm`, `addmv`, `mm`, `bmm`, `sampled_addmm`, dimension reductions, `mm_reduce`, `triangular_solve`) still have no Metal kernels. Those are follow-up PRs.
- `aten::segment_reduce` has no MPS kernel, which is what `torch.masked` reductions over a CSR tensor need.
- The distribution samplers only gained Metal kernels recently: `_standard_gamma` in April 2026 (pytorch#179228), `_sample_dirichlet` in May (pytorch#185458), `poisson` in June (pytorch#173319), `binomial` in June (pytorch#187078). These affect `sample()`, so `posterior_predictive_sample` and `get_likelihood_parameters` rather than the training loop, but if you are on an older torch the probes in `scvi/distributions/_utils.py` will be routing them through the CPU.

Could you post `torch.__version__` from the run? If it predates those, part of what you measured is a torch that does not have the kernels yet, independently of this branch.

I also found and fixed three genuine MPS bugs upstream while validating this: a GPU hang in `scatter_reduce_`/`index_reduce_` whenever the source contains a NaN, `index_reduce_` silently dropping NaN for `amin`/`amax`, and `nn.BCELoss` tripping an internal assert on a zero-element tensor (the one `_multivae.py` already works around). None of those are performance, but the first one hangs the process rather than erroring, so it is worth having.

BenjaminDEMAILLE:
I opened an umbrella tracking issue upstream so this is followable from the torch side: pytorch/pytorch#196192. It lists what the first stack enables on MPS and, in the "Not covered" checklist, exactly which sparse operations still have no Metal kernel — that checklist is the honest answer to "is there still a CPU fallback", at least for the sparse path.

There is also pytorch/pytorch#196190 for one unresolved defect I hit while validating: `mul` on an integer CSR tensor intermittently produces an invalid compressed tensor. It does not affect the dataloader path this PR uses, but I would rather have it on record than discover it later.

ori-kron-wis:
@BenjaminDEMAILLE Thanks again for your contribution and detailed explanations!

I confirm that increasing the batch size does reflect the increase the speed in MPS vs CPU, as expected.
I can also confirm that including the fused Adam from the other branch and run compile, It is even more faster (I merged both branches together in order to test it). ELBO is the same for all cases. So a dramatic improvement indeed!

Do you think we should run the MPS CI workflow with PYTORCH_ENABLE_MPS_FALLBACK=0?  With "1" I think it is still faster, but perhaps were missing some checking (might be both options as well).

Any more commits you expect us to do? Do you think we can merge this and the other branch?

BenjaminDEMAILLE:
Glad the batch-size sweep reproduced on your side, that settles the latency-bound question.

**On `PYTORCH_ENABLE_MPS_FALLBACK=0` in CI: yes, and I would make it the only setting for that job.**

Worth being precise about what the variable does, because it is easy to read it backwards: `0` is torch's own default, and `1` is opting *into* silent CPU fallback. So the current CI is the permissive configuration, not the strict one.

With `1`, an op that has no Metal kernel quietly runs on the CPU. Results stay correct, the run is slower, and the job stays green. That is exactly the failure mode this whole PR exists to remove, and CI cannot currently see it.

It costs nothing to switch, because the strict setting already passes on this branch:

```
PYTORCH_ENABLE_MPS_FALLBACK=0
891 passed, 142 skipped, 0 failed        # zero NotImplementedError
```

On "with 1 I think it is still faster": the variable only has any effect when a kernel is missing, and none is on torch 2.14, so it cannot change the runtime here. Whatever difference you saw between the two runs is machine noise rather than the flag. Running both settings would give you two jobs exercising identical code paths at identical speed, which is why I would keep just the strict one.

The real cost is elsewhere, and you should weigh it: with `0`, any op added later without a Metal kernel turns a silent slowdown into a hard CI failure, including ops that arrive through a dependency rather than through scvi-tools. That is the intended behaviour, but it does mean the mps job will sometimes break for reasons unrelated to the PR that trips it.

**On merging.**

Nothing further from me on this branch; my last commit is 6eddffcc.

One coordination point though: bfff3b5a merged `training-plan-fused-adam-and-compile` into this branch, so #3981 now contains #3998's commits. If you merge this one, that work lands with it and #3998 should be closed rather than merged, otherwise you get the same commits twice on main. Either order is fine by me, it is just worth picking one deliberately:

- merge #3981 as it stands, close #3998 as superseded; or
- revert bfff3b5a here and merge them separately, which keeps the changelog and `git bisect` story cleaner since the optimizer change is device-independent and affects CUDA users too.

I lean slightly towards the second, but not enough to argue for it.

Either way there is an upside to the merge you already did: the `cuda` job passed on this branch with the fused optimizer included, which is coverage #3998 could not get on its own since its `cuda` and `mps` jobs are label-gated. If you do go the separate route, adding the `mps` label to #3998 before merging would be worth it, since that is the device the change targets and I could only test it locally.

PR #3995, opened 2026-09-06
fix(model): resolve BaseModelClass.view_registry  AttributeError and align _registry/registry access

ori-kron-wis:


ori-kron-wis:
close https://github.com/scverse/scvi-tools/pull/3991
close https://github.com/scverse/scvi-tools/pull/3992

PR #3998, opened 2026-09-06
Use fused Adam where available, and stop torch.compile leaking dynamo state

BenjaminDEMAILLE:
Two independent changes to `TrainingPlan`. They are in one PR because they touch the
same file; happy to split if you would rather review them separately.

## Where this came from

While answering @ori-kron-wis's question on #3981 about `mps` being only ~10% faster
than `cpu` on the introduction tutorial, I measured the training step rather than
guessing. It is dispatch-bound, not compute-bound:

- one step on a 1200-gene, 128-hidden SCVI model issues **916 aten ops**
- a single elementwise op costs **7.35 us on cpu against 11.14 us on mps**

The cpu figure is the Python and dispatcher floor, which is device-independent, so the
overhead specific to the Metal backend is only about a third of the per-op cost. That is
why the accelerator barely helps at the default `batch_size=128`: the GPU is idle waiting
for work to be dispatched.

Nothing is falling back to the CPU, incidentally. Running the whole training with
`PYTORCH_ENABLE_MPS_FALLBACK=0` completes without error.

## 1. Fused optimizer

The default Adam path accounts for about **423 of those 916 ops**, and for **17 of the 22
host-device synchronisations** per step, because the single-tensor path reads its step
counter back to the host once per parameter. Selecting the fused implementation brings the
step to 493 ops and 5 synchronisations.

Measured on the tutorial dataset (18641 x 1200), 20 epochs, machine otherwise idle,
alternating the two configurations twice to absorb thermal drift:

```
        default        fused
mps     0.989 s/ep     0.697 s/ep     1.42x
cpu     1.138 s/ep     1.050 s/ep     1.08x
```

ELBO is unchanged: 309.13 against 309.14 on mps. The `mps` advantage over `cpu` goes from
1.15x to 1.51x, which is most of what the original question was about.

Whether fused is available depends on the device, the parameter dtypes and the torch
version, and torch reports that by raising at construction time, so the fallback is driven
by the exception rather than by a device allowlist maintained here. `fused_optimizer=False`
restores the previous behaviour.

I did **not** pick `foreach=True` (slower here, 911 ops and *more* syncs) and
`capturable=True` is rejected outright on `mps`.

## 2. torch.compile no longer leaks dynamo state

`compile=True` set `torch._dynamo.config.suppress_errors = True` and never put it back, so
one compiled model disabled dynamo's error reporting for the rest of the process. Since
compilation is lazy, the flag is restored in `on_train_end` rather than around the
`torch.compile` call.

More importantly, a silent fallback to eager is now warned about. Previously a user could
ask for a compiled model, get an eager one, and have nothing say so; I nearly published a
wrong measurement because of this before counting the compiled frames.

`compile` had no test coverage at all before this.

## Testing

```
tests/train/                          49 passed, 2 skipped     (torch 2.14.0, mps)
full suite --accelerator mps          851 passed, 2 failed
```

The two failures are `test_datasplitter_load_sparse`, which fail identically on `main`
without this commit (`NotImplementedError: Could not run 'new_compressed_tensor' from the
'mps:0' device`) and are already fixed by 4c73605e on #3981.

New tests cover fused selection, opting out, the fallback when the backend rejects fused,
the dynamo flag being restored, the eager-fallback warning, and an mps-gated check that
frames actually compile.

Measurements on an M4 Max, 16 cores, torch 2.14.0, python 3.13.15.

### DIFF
--- a/.github/workflows/test_mps.yaml
+++ b/.github/workflows/test_mps.yaml
@@ -31,7 +31,7 @@ jobs:
       )
 
     name: mps
-    runs-on: [self-hosted, macOS, X64, MPS]
+    runs-on: macos-latest
 
     defaults:
       run:
@@ -60,7 +60,8 @@ jobs:
           DISPLAY: :42
           COLUMNS: 120
           PYTORCH_MPS_HIGH_WATERMARK_RATIO: 0.0
-          PYTORCH_ENABLE_MPS_FALLBACK: 1
+          PYTORCH_ENABLE_MPS_FALLBACK: 0
+          SCVI_ALLOW_MPS_AUTO: "1"
         run: uvx hatch run $HATCH_ENV:run-cov -v --color=yes --accelerator mps --devices auto
 
       - name: Generate coverage report
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,24 @@ to [Semantic Versioning]. The full commit history is available in the [commit lo
 
 #### Added
 
+- Extend native `mps` support across the rest of scvi-tools' gamma/Poisson/Dirichlet/Binomial
+    sampling and `lgamma` call sites, on top of
+    {class}`scvi.distributions.NegativeBinomial`, {class}`scvi.distributions.ZeroInflatedNegativeBinomial`
+    and {class}`scvi.distributions.NegativeBinomialMixture`. The CPU detours these relied on for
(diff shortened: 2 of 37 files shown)


ori-kron-wis:
This will be part of https://github.com/scverse/scvi-tools/pull/3981

PR #4007, opened 2026-09-09
fix(cytovi): avoid nan loss with beta likelihood on merged overlapping panels

ori-kron-wis:


ori-kron-wis:
close https://github.com/scverse/scvi-tools/issues/4006

